PR #71, opened 2022-02-18
a message that will be public

### LINKED ISSUE #66: Provide facility to manage multiple forks of the same upstream repository
It is expected that in a single organization, we would have more than one individual contributors to the same upstream repository.  

Example: Linda in Lisbon, Dave in Dublin, and Taro in Tokyo have all forked Node.JS and independently have worked on patches to contribute. GitProxy should recognize that these 3 upstreams have a common ancestor and allow to manage the approvals and audit of these together. 

The most straight-forward implementation approach is to be able to manually link multiple proxy-repos as a group. 

As a separate service, we can have repo-similarity detection service, looking for common branches, containing identical commit chains greater than a size-threshold. It is an open problem how to distinguish legit forks (i.e. OpenSSL vs. BoringSSL)

grovesy:
maoo, this is a test 

closes #66

### DIFF
--- a/test/1.js
+++ b/test/1.js
@@ -3,6 +3,7 @@ const chai = require('chai');
 const chaiHttp = require('chai-http');
 const service = require('../src/service');
 
+
 chai.use(chaiHttp);
 chai.should();
 


grovesy:
closing test

PR #100, opened 2023-03-09
fixes workflows

abacaphiliac:
fix: updates workflows to run on `main` (the default branch) instead of `master`, and installs `mocha` as a `devDependency` inside the project instead of requiring a global installation so that `npm test` works "out of the box" on developer workstations

### DIFF
--- a/.github/workflows/license-validation.yml
+++ b/.github/workflows/license-validation.yml
@@ -5,9 +5,9 @@ name: License Validation
 
 on:
   push:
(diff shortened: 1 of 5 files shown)


JamieSlome (approved):
LGTM! 👍

@grovesy?

grovesy:
> LGTM! 👍
> 
> @grovesy?

Looks good to me!, approved

PR #182, opened 2023-07-01
fix: commit lint bug

temideewan:
Introduced a preinstall hook that ensures that [husky](https://typicode.github.io/husky/) is installed after a npm install.
Checks for proper commit messages should pass properly after this.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,6 +7,7 @@
     "": {
       "name": "@finos/git-proxy",
       "version": "1.0.0",
+      "hasInstallScript": true,
       "license": "Apache-2.0",
       "dependencies": {
         "@material-ui/core": "^4.11.0",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,8 @@
     "server-test-ci": "mocha --exit",
     "server-test": "mocha --exit",
     "test": "mocha --exit",
-    "eject": "react-scripts eject"
+    "eject": "react-scripts eject",
+    "preinstall": "npx husky install"
   },
   "author": "Paul Groves",
   "license": "Apache-2.0",


JamieSlome:
@temideewan - I'm just thinking, does it make sense to run this before every `npm install`. Although, that said, is there any other point in time that we can run this?

temideewan:
I see what you're saying.
All the options I've considered so far have to do with making sure that `husky install` actually runs before any interaction with the project.
The reasonable thing anyone would run before interacting with the project (especially when they pull an update) is `npm install` that's why I'm trying to attach it to one of the hooks.
One thing I don't know yet is how to check if the current user has already ran `husky install`.

According to [this blog](https://www.gitkraken.com/blog/husky-git-hooks#:~:text=Husky%20is%20a%20tool%20that,Husky%20works%20within%20the%20package.) we could switch `preinstall: "npx husky install"` for `prepare: "husky install"` and it'd still achieve the same end results.
But prepare also[ runs with](https://docs.npmjs.com/cli/v9/using-npm/scripts#life-cycle-scripts) `npm install` so it's a bit dicy to figure out a good balance

JamieSlome:
@temideewan - thank you for doing the deep dive! Happy to go with the hook you think is most appropriate.

Happy to proceed? I'll merge once you give me the green flag!

JamieSlome (approved):
LGTM! 🍰

temideewan:
Yes there's an improvement I actually found to ensure we don't run it every single time, but I don't have access to my system right now so I can't test it out right now.
If you still feel a bit uncomfortable with the husky install running every single time [this blog](https://sonspring.com/journal/husky-v5-and-npm-prepare/) gives a pretty good description of the solution I wanted to try out.
Maybe you could have a look at it.

JamieSlome:
> Yes there's an improvement I actually found to ensure we don't run it every single time, but I don't have access to my system right now so I can't test it out right now. If you still feel a bit uncomfortable with the husky install running every single time [this blog](https://sonspring.com/journal/husky-v5-and-npm-prepare/) gives a pretty good description of the solution I wanted to try out. Maybe you could have a look at it.

Sorry, I didn't get back to you! I didn't have the time to get around to this over the weekend. Do you think we could give this another look? I have seen the usage of `prepare` in different projects and it seems to work. Shall we create an issue?

temideewan:
> > Yes there's an improvement I actually found to ensure we don't run it every single time, but I don't have access to my system right now so I can't test it out right now. If you still feel a bit uncomfortable with the husky install running every single time [this blog](https://sonspring.com/journal/husky-v5-and-npm-prepare/) gives a pretty good description of the solution I wanted to try out. Maybe you could have a look at it.
> 
> Sorry, I didn't get back to you! I didn't have the time to get around to this over the weekend. Do you think we could give this another look? I have seen the usage of `prepare` in different projects and it seems to work. Shall we create an issue?
Yes I think we should

JamieSlome:
@temideewan - created https://github.com/finos/git-proxy/issues/188 👍

Issue #183, opened 2023-07-03
Make installation and usage of the library easier


JamieSlome:
Now that we have distributed the library to [npm](https://www.npmjs.com), the easy installation and usage of the project are critical.

@grovesy - I am having trouble creating a simple set of steps to launch the service as a dependency. Can you support me here? I intend on adding this to #177.

- [ ] Describe a set of steps for installing and launching Git Proxy with base configuration

maoo:
@JamieSlome - I quickly looked at `package.json` and it seems that you could add `"bin" : "./index.js"` to be able to run it using `npx @finos/git-proxy` . Found https://deepgram.com/learn/npx-script quite useful, it also explains how to deal with arguments.

Hope this helps!

JamieSlome:
@maoo - nice suggestion. It looks like we still have an issue with absolute references to `resources/config.json` which isn't available when a developer is consuming the library as a dependency.

If you try to install the library with `npm install @finos/git-proxy` and then run `index.js`, you can replicate the issue.

I wonder if there is a nice way to populate the root with a **default** configuration which we can refer to instead? Is this what the `user-settings.json` looked to achieve?

JamieSlome:
https://github.com/finos/git-proxy/blob/7fb860c2cb200cad7e54864c4e616fa7159a5718/src/config/index.js#L2

This is the issue. On installation, the developer can't access this path. I think we are best generating a default config file (`proxy.config.json` and sticking it in the root. @maoo @TheJuanAndOnly99 - thoughts?

JamieSlome:
@maoo - I think it would be good to get a configuration file available at the point of installation, similar to how `nuxt.config.js` is created when you create a fresh Nuxt web application. This way the developer can see exactly what comes out of the box, and how they can begin configuring.

Moreover, we could have a configuration walkthrough at the start of the `npx` or `npm install` journey?

![image](https://github.com/finos/git-proxy/assets/55323451/1d4e6c2f-9564-4eca-a7d9-676fa1e12b6b)

This can be a future improvement, but for now, we need to fix the default configuration as currently neither the default nor custom configuration works. Unless I am missing something with regard to installation and setup?

JamieSlome:
@maoo - will give this a go now and open a PR with some of the changes discussed above, including `npx` usage 👍

JamieSlome:
Closing #184 post discussion with @coopernetes in favor of vanilla NPM installation and usage.

PR #184, opened 2023-07-04
Support `npx` usage and fix default configuration loading

JamieSlome:
In support of #183

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -1,3 +1,5 @@
+#!/usr/bin/env node
+
 const proxy = require('./src/proxy');
 const service = require('./src/service');
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "@finos/git-proxy",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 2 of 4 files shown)


JamieSlome:
@TheJuanAndOnly99 - looks like we can remove the `push` trigger for the CVE scanning. Thoughts?

TheJuanAndOnly99:
@JamieSlome Yes that would be fine. I think if we do that we should also set a branch protection rule on the `main` branch to require a pull request before merging. What do you think?

JamieSlome:
@TheJuanAndOnly99 - yes, perfect! I don't have full access to the repository's administrative settings. Is this something we can also arrange?

TheJuanAndOnly99:
@JamieSlome I've just sent a PR https://github.com/finos/git-proxy/pull/189 with the update to the action. And I have set branch protection on the `main` branch to require a PR before merging. 

As far as admin rights go. FINOS policy is to not grant permanent admin rights to anyone (only the finos-admin user has admin privileges). But if you need temporary admin access to set something up inside the repo we can arrange that if you email help@finos.org.

JamieSlome:
@TheJuanAndOnly99 - merged, thank you 🎉 

No worries with regards to the permissions!

JamieSlome:
Closing as discussed with @coopernetes in the fortnightly catch-up yesterday.

Issue #185, opened 2023-07-04
Fix breakages caused by migration to latest `react-router-dom@6.14.1`


JamieSlome:
**Describe the bug**

The web application is no longer able to generate due to major changes in the `react-router-dom` library. Examples include the deprecation of the [`Redirect`](https://stackoverflow.com/questions/63690695/react-redirect-is-not-exported-from-react-router-dom).

**To Reproduce**

1. Run `npm run start`
2. Visit `localhost:3000` to see all issues caused by misuse of the latest version of `react-router-dom`

PR #199, opened 2023-07-11
fix: rename files and use vite config in start script

vaibssingh:
Hey @JamieSlome , this should fix the start script and hopefully the build script too but I can't confirm either till we fix the issue with ```react-router-dom```.

On starting the server and navigating to ```localhost:3000``` I can see the API calls are being made and it tries to load the page till it encounters invalid imports from react-router-dom. So at least that's an improvement from yesterday's behaviour where it just gave 404 on opening.

Let me know how you want me to proceed with this. I am happy to help you with the updates regarding react-router-dom.

Thanks for your help.


JamieSlome:
@vaibssingh - awesome work 👍 

I have started fixing up the `react-router-dom` problem in #185. I have pushed my changes to `185-fix-breakages-caused-by-migration-to-latest-react-router-dom6141` which I believe addresses all of the remaining issues. I will merge into `main` and you can merge in the latest changes into your branch then we can go from there.

JamieSlome:
For your reference: https://github.com/finos/git-proxy/pull/200

JamieSlome:
@vaibssingh, a few conflicts for you there, sorry! The source of truth for what to keep comes from #200 😄

vaibssingh:
Alright @JamieSlome! I will merge this and test this. Hopefully this should resolve the issues 🤞🏽

vaibssingh:
Hey @JamieSlome , updated the PR with the latest code. Checked and it seems to be working. ```npm run build``` & ```npm start``` both are working.

On UI, login screen opens up asking for credentials which I do not have so not able to test further than that.

src/ui/variables/charts.js
```diff
@@ -185,8 +185,7 @@ const completedTasksChart = {
   },
 };
 
-module.exports = {
-  dailySalesChart,
-  emailsSubscriptionChart,
-  completedTasksChart,
-};
+// module.exports = {
```
JamieSlome, line 188: If these are covered by the above `export const` can we remove the comments? No need to keep the code if it isn't being used.

JamieSlome (changes requested):
Please address the one comment, and then we are ready for lift-off! 🚀

vaibssingh:
@JamieSlome Done! 👍🏽

JamieSlome (approved):
LGTM! 🍰

PR #258, opened 2023-08-23
fix: add pre-commit lint hook to husky

### LINKED ISSUE #241: Add githook to run linting on commits to the repo
Currently there is no way to insure that all files are linted as per the config before they are committed. We only get alert after pushing it to repo and running the lint workflow.

A githook that runs on each commit and runs the ```npm run lint``` script. In case of any error, it should fail the commit and ask the user to fix the issue.

Currently we already have a GitHub Workflow that runs after each commit and on creating a PR.

**Additional context**
Originally discussed in #238

vaibssingh:
Closes #241 

@JamieSlome Tested and verified on local by creating linting issues and then trying to commit. This fixes any auto fixable issues and doesn't allow the commit to go through if it can't fix them all. User can see the errors in the ```Command Output``` as well in the ```Problems``` console where it would be highlighted already.

As for the separate ```.prettierrc``` file that we had discussed, we don't really need this one here as we are already using ```eslint-plugin-prettier``` and ```eslint-config-prettier``` which essentially combine eslint with prettier to work together and work to avoid any conflicts. In case any conflicts still come up or specific customisation is needed, they are defined in the rules object in the ```eslintrc.cjs``` file. For more details, you can refer [here](https://prettier.io/docs/en/integrating-with-linters).

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-npx --no -- commitlint --edit ${1}
+npx --no -- commitlint --edit ${1} && npm run lint


JamieSlome:
@vaibssingh - very clear and as usual, thank you for your contribution 👍 

One thing I was thinking; on the PR checks, we get a duplication of linting review on push and pull request - do we need both?

vaibssingh:
Yeah I saw that too. Let's keep it only for pull requests since we already have one is place for push. Let me know if you are okay with this. :)

PR #263, opened 2023-08-24
fix: run lint check only on pull requests

vaibssingh:
Runs lint workflow only on pull requests instead of the previous behaviour where it was running on push as well as pull requests.

Discussed in #258


JamieSlome:
Thanks, @vaibssingh 👍

PR #219, opened 2023-08-04
Merging upstream changes into mvp

maoo:
There are 264 changes on `main` branch that are not ported into the `mvp` branch, which is the most feature-complete one.

PR #266, opened 2023-08-25
fix: cleanup deprecated behaviour from react-router

coopernetes:
preface: not well versed in front-end or React. Tidying up a series of errors seen when pulling up the ui. A follow-up from #185 

- fix redirect from `/login` to `/admin` on submit. It would dump you back to the main login page and wouldnt redirect.
- fix below error when loading `/admin` by replacing `<Navigate />` with a `<Route ...element={<Navigate />} />` component:
    ```
    Uncaught Error: [Navigate] is not a <Route> component. All component children of <Routes> must be a <Route> or <React.Fragment>
    ```

- make parent route /admin/* so children are routed correctly:
    ```
    You rendered descendant <Routes> (or called `useRoutes()`) at "/admin" (under <Route path="/admin">) but the parent route path has no trailing "*". This means if you navigate deeper, the parent won't match anymore and therefore the child routes will never render.
    ```
- convert raw props to useParams() in components with URL path params
    ```
    props.match is undefined
    ```
- in /admin/* Route setup, pass components as elements via JSX to resolve error:
    ```
    Functions are not valid as a React child. This may happen if you return a Component instead of <Component /> from render.
    ```

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -20,7 +20,7 @@ const hist = createBrowserHistory();
 ReactDOM.render(
   <Router history={hist}>
     <Routes>
-      <Route exact path="/admin" element={<Admin />} />
+      <Route exact path="/admin/*" element={<Admin />} />
       <Route exact path="/login" element={<Login />} />
       <Route exact path="/" element={<Navigate from="/" to="/login" />} />
     </Routes>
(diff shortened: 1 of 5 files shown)


coopernetes:
Realized this may conflict with a lot of changes in #219 - if it's not worth fixing at the moment, this can be closed until after that is merged.

JamieSlome (approved):
LGTM! 🍰

JamieSlome:
@coopernetes - not a problem to merge, as I am taking care of a large bulk of the reconciliation between `mvp` and `main` internally, and then will shortly publish the merge. Your changes are nice and discrete so easy to merge in 👍

Issue #275, opened 2023-09-04
feat: Add Dockerfile for container-based deploys


coopernetes:
**Is your feature request related to a problem? Please describe.**
It's very common in cloud-based deployments to ship a Docker/OCI container for web applications & servers. It would be useful if we provided a default Dockerfile for users building & deploying git-proxy from source to be able to conveniently build the project and create a Docker image. 

As part of a release of git-proxy, the project could also publish a Docker image to [Docker Hub](https://hub.docker.com) under a FINOS namespace for users to simply reuse directly.

**Describe the solution you'd like**
- add a Dockerfile to the repository to build git-proxy as a container image (see https://github.com/finos/git-proxy/pull/269/commits/095cc3a1b48db5ab13fce152585045567282e0e8 for an example)
- build & push an image as part of the release process

**Describe alternatives you've considered**
- we already publish a runnable package to NPM so this may be extraneous. A user Dockerfile could be something as simple as:

```Dockerfile
FROM node:18
ARG GITPROXY_VERSION="1.0.0"

RUN npm install -g @finos/git-proxy@$GITPROXY_VERSION

ENTRYPOINT ["git-proxy"]
```

Rajdeep1311:
I would like to see to this issue.

JamieSlome:
@Rajdeep1311 - awesome, thanks for getting involved! 🤝

I'll assign the issue to you!

Rajdeep1311:
Thanks sir @JamieSlome

JamieSlome:
@Rajdeep1311 - of course! Feel free to post any questions you have here.

I am happy to help and I am sure @coopernetes will be able to provide some guidance as well.

How can we help?

Rajdeep1311:
I created a dockerfile, built a docker image out of it, and ran it but having some issues with the port, it should run on 3000 but firewall problems I guess

coopernetes:
@Rajdeep1311 git-proxy runs on 8000 for the proxy server and 8080 for the management UI. You will need to expose those ports and not 3000.

Rajdeep1311:
Okay sir

Rajdeep1311:
Sir actually I am new to DevOps, Can you help which file is to be used to create the docker file @coopernetes

Rajdeep1311:
Also, I have tried something, if you can see to it, I would like to demonstrate @coopernetes

coopernetes:
[...]

PR #282, opened 2023-09-13
feat: Add Dockerfile for container-based deploys #275 (solved)

Rajdeep1311:
Solved issue #275

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Use an official Node.js runtime as a parent image
-FROM node:14
+FROM node:18
 
 # Set the working directory to /app
 WORKDIR /app
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+# Use an official Node.js runtime as a parent image
+FROM node:14
+
+# Set the working directory to /app
+WORKDIR /app
+
+# Copy package.json and package-lock.json to the container
+COPY package*.json ./
+
+# Install project dependencies
+RUN npm install
+
+# Copy the rest of the application code to the container
+COPY . .
+
+# Expose the port that your application listens on
+EXPOSE 8000
+
+EXPOSE 8080
+
+# Define the command to run your application
+CMD ["npm", "start"]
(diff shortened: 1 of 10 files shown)


Rajdeep1311:
Okay sir

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# Copy the rest of the application code to the container
+COPY . .
+
+# Expose the port that your application listens on
+EXPOSE 8000
+
+EXPOSE 8080
+
+# Define the command to run your application
+CMD ["npm", "start"]
```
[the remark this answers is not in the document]
Rajdeep1311: Should I use yarn?

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+FROM node:18
+
+# Set the working directory to /app
+WORKDIR /app
+
+# Copy package.json and package-lock.json to the container
+COPY package*.json ./
+
+# Install project dependencies
+RUN npm install
```
coopernetes, line 11:
```suggestion
RUN npm ci --omit=dev --ignore-scripts
```

Rajdeep1311: Okay sir

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+# Copy the rest of the application code to the container
+COPY . .
```
coopernetes, line 14: Combine this with the multi-stage change above.
```suggestion
FROM node:18-slim

COPY --from=build /app/node_modules/ node_modules/
COPY . .
```

Dockerfile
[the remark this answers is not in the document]
coopernetes: > Should I use yarn?
This project does not use yarn at the moment. 

1. There is an [existing `ENTRYPOINT` script in the node image which you should use directly](https://github.com/nodejs/docker-node/blob/main/docker-entrypoint.sh). 
2. npm should never be used to start a Node application because [npm swallows the exit signals that container runtimes send](https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#cmd), which can and will interfere with those runtimes trying to stop & gracefully shutdown containers running git-proxy.

```suggestion
CMD ["index.js"]
```

coopernetes (changes requested):

PR #313, opened 2023-10-18
Add note re. ssh support

maoo:


JamieSlome:
@maoo @TheJuanAndOnly99 are we able to only run documentation previews only when we change the contents of the `website` folder?

TheJuanAndOnly99:
@JamieSlome to trigger previews on all changes you can add a netlify.toml file to the root of the project just like this one https://github.com/finos/software-project-blueprint/blob/main/netlify.toml.

PR #325, opened 2023-10-25
Add `netlify.toml` with build base directory specified

### LINKED ISSUE #324: Only run doc build process with changes are made exclusively to the docs
On every pull request, the documentation preview, build, and deploy process happens. We only want this to occur on pull requests and commits that directly change the `website` folder, which holds all of the documentation website content.

JamieSlome:
Closes #324

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,2 +1,2 @@
 [build]
-   base = "/website"
+  ignore = "git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF . ../{docs,website}/"
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,2 +1,2 @@
 [build]
-   base = "./website"
+   base = "/website"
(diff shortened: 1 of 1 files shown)


maoo:
Thanks for this @JamieSlome !

I'm not sure this is the right solution (in fact, this PR should not have triggered the build/preview in theory); please keep in mind that we also need to rebuild against changes to the `docs` folder.

I'd suggest the following content on netlify.toml:
```
[build]
  ignore = "git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF . ../{docs,website}/"
```

This is currently working for FDC3, see https://github.com/finos/FDC3/blob/master/netlify.toml

JamieSlome:
How can we re-test this @mao?

maoo:
@JamieSlome , just realised that `docs` is a subfolder of `website`, so we can probably update this to :

```
[build]
  ignore = "git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF . ../website/"
```

Re. testing, it's probably ok to merge and run some PRs to see what's going on. I know it's not ideal.

JamieSlome:
Updated, but still seems to be running even on this PR (@maoo)

maoo:
> Updated, but still seems to be running even on this PR (@maoo)

Could be that netlify.toml changes always triggers a build? Found this discussion to be relevant - https://answers.netlify.com/t/builds-cancelled-for-a-new-branch-due-to-no-content-change/17169

I'd try to merge some of these changes to main and see if there's any change in behaviour.

Issue #48, opened 2021-01-27
Repository entitlements workflow 


grovesy:
When a repository is added to the git-proxy an admin used can through API and UI

1. Add a registered user as an authorizer (can authorize or reject)
2. Add a registered user as allowed to work with the repository
3. can pull / can push 

Add a constraint so that the authorizer cannot be the same as the author.

JamieSlome:
Addressed in #335. Closing this issue 👍

Issue #42, opened 2020-11-27
LDAP / Active Directory Authorisation


grovesy:
AD authentication plugin should have the following configuration options

1. LDAP/AD information
2. Admin group (can do anything group)
3. authorizer-group (can authorize pushes)
4. write-group (can-push)
5. read-group (can pull)

The login logic should be 

1. Check if the AD user is in the correct-groups 
2. If is in one or more  of the groups and does not exist as a local user - create a new local user
3. If the local user does exist in a group AND locally, update the local user's profile with the latest AD information

4. AD information to propagate
- email address, 
- access permisions based on group membership

JamieSlome:
Addressed in #335. Closing this issue 👍

Issue #38, opened 2020-11-23
Implement Authentication tokens for the git client


grovesy:
The developers git client uses the target GIT servers authentication/authorization mechanism - and this is pass-through/transparent to the proxy. 

- We need to tie the developer making a 'push' or 'pull' with a GitProxy user (which ties to the corporate account)
- We cannot interfere with the authentication with the target git-server (i.e. GitHub)
- We do not want to hold Git credentials for the user 

Therefore a Proxy user will be issued a token - 

**Getting a Token before working with the proxy through a git client** 
1. The developer logs into the GitProxy UI, navigates to their account page
2. The user enacts 'get token' - this reveals a token to the user. e.g. abc12345
3. Through the users git-client the user will run (replace finos/git-proxy.git with the repo they want to work with)

``` bash
git remote add origin https://abc12345@gitproxy.corp.net/finos/git-proxy.git

```

When Clone, Pull, Push requests are made through the proxy, the proxy will take the token and tie that to the git-proxy user. The proxy will then check that the token is valid and the user is allowed to work with the repository.

grovesy:
I am thinking of not doing this in the short term - instead for the MVP tie the git-proxy user with the git-user

Issue #28, opened 2020-10-07
Soft 'warning' on pushes


grovesy:
# Soft warnings

When making a push, a 'processor' could feel that it doesn't want to 'hard block' the push request but instead alert the 'approver' of a potential issue 

1. The contributor should be alerted of the 'warn' through their git client
2. The approver in the UI should be aware of the problem, it should be easy for them to review the warning and reject/approve

JamieSlome:
Achieved and available in latest version 👍

PR #335, opened 2023-10-31
Reconcile Git Proxy with most up-to-date improvements

### LINKED ISSUE #201: Reconcile changes on `mvp` branch with `main`
A significant number of changes were made to the library in the `mvp` branch. Are we able to reconcile the latest changes to `main` with the changes made on the `mvp` branch?

@grovesy @maoo - any help from your end would be appreciated as a lot of this code was created before my time 😆 

```[tasklist]
### Tasks
- [x] Open a pull request from `mvp` to `main`
- [ ] Merge in `main` latest changes to `mvp`
- [ ] Resolve merge conflicts (no doubt there will be loads)
```

JamieSlome:
Closes #201 #48 #42 #38 #17 #9 #423 #28 #454


JamieSlome:
> This is great, thanks @JamieSlome !
> 
> I looked at the diff, it seems that all changes apply to dashboard and testing, is that correct?
> 
> On top of the [quickstart instructions](https://git-proxy.finos.org/docs/category/quickstart/), what are the steps to run the UI? I'd be happy to help testing/reviewing here.

Added you as a reviewer slightly too early @maoo! Still lots of changes to come 👍

JamieSlome:
@maoo @coopernetes - getting extremely close with this now. It might be worth having a mull over the code. The service and proxy are broken in the current form, but I believe this is a result of too tightly coupling the setup to a MongoDB configuration rather than file-based storage.

Here if you have any questions 👍 ❤️

coopernetes:
This change set is fantastic - much appreciate all the work that's going into this PR @JamieSlome ❤️ 

I started a cursory review of the proxy server & API changes and will provide feedback if needed. Lots of exciting stuff landing in this PR 🙌

coopernetes:
Patch for `Error: Cannot find module 'express-mongodb-session'`

```diff
diff --git a/package.json b/package.json
index fb9e139..abb0686 100644
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "email-validator": "^2.0.4",
     "express": "^4.18.2",
     "express-http-proxy": "^2.0.0",
+    "express-mongodb-session": "^3.2.1",
     "express-session": "^1.17.1",
     "generate-password": "^1.5.1",
     "history": "5.3.0",
```

src/proxy/processors/push-action/parsePush.js
```diff
@@ -24,41 +25,96 @@ const exec = async (req, action) => {
...
+      const message = formattedContent
+        .slice(indexOfMessages + 1, formattedContent.length - 1)
+        .join(' ');
+      console.log({ message });
+
+      const commitTimestamp = committer.split(' ')[2];
+      console.log({ commitTimestamp });
+
+      const authorEmail = author.split(' ')[1].slice(1, -1);
+      console.log({ authorEmail });
```
coopernetes, line 98: Some of the parsing logic seems bugged. From the logs:

```
[0] { author: 'Thomas Cooper <coopernetes@proton.me> 1705253171 -0500' }
[0] { committer: 'Thomas Cooper <coopernetes@proton.me> 1705253171 -0500' }
[0] { indexOfMessages: 11 }
[0] { message: 'test' }
[0] { commitTimestamp: '<coopernetes@proton.me>' }
[0] { authorEmail: 'oope' }
```

```suggestion
      const authorEmail = author.substring(author.indexOf('<')+1, author.indexOf('>'));
      console.log({ authorEmail });
```

src/proxy/processors/push-action/parsePush.js
```diff
@@ -24,41 +25,96 @@ const exec = async (req, action) => {
...
+      const indexOfMessages = formattedContent.indexOf('');
+      console.log({ indexOfMessages });
+
+      const message = formattedContent
+        .slice(indexOfMessages + 1, formattedContent.length - 1)
+        .join(' ');
+      console.log({ message });
+
+      const commitTimestamp = committer.split(' ')[2];
+      console.log({ commitTimestamp });
```
coopernetes, line 95: Wrong index?

Logs:
```
[0] { committer: 'Thomas Cooper <coopernetes@proton.me> 1705253171 -0500' }
...
[0] { commitTimestamp: '<coopernetes@proton.me>' }
```

```suggestion
      const commitTimestamp = committer.split(' ')[3];
      console.log({ commitTimestamp });
```

coopernetes (commented):
First few comments from running this branch locally and poking at it. A few suggested fixes and comments.

src/proxy/processors/push-action/parsePush.js
```diff
@@ -24,41 +25,96 @@ const exec = async (req, action) => {
...
+
+      const authorEmail = author.split(' ')[1].slice(1, -1);
+      console.log({ authorEmail });
+
+      console.log({
+        tree,
+        parent,
+        author: author.split(' ')[0],
+        committer: committer.split(' ')[0],
```
coopernetes, line 104:
```suggestion
        author: author.split('<')[0].trim(),
        committer: committer.split('<')[0].trim(),
```

src/proxy/processors/push-action/parsePush.js
```diff
@@ -24,41 +25,96 @@ const exec = async (req, action) => {
...
-        tree: tree.replace('parent', '').trim(),
-        parent: parent.replace('parent', '').trim(),
-        author: author.split(' ')[1],
-        committer: committer.split(' ')[1],
-        commitTs: author.split(' ')[3],
-        message: message,
+        tree,
+        parent,
+        author: author.split(' ')[0],
+        committer: committer.split(' ')[0],
```
coopernetes, line 114:
```suggestion
        author: author.split('<')[0].trim(),
        committer: committer.split('<')[0].trim(),
```

src/proxy/processors/push-action/pullRemote.js
```diff
@@ -20,22 +20,43 @@ const exec = async (req, action) => {
...
+    const cloneUrl = authorizationValid
+      ? `https://${Buffer.from(authorizationHeader.split(' ')[1], 'base64')}@${action.url.replace(
+          /https*:\/\//,
+          '',
+        )}`
+      : action.url;
+
+    step.log(`Exectuting ${cmd}${authorizationValid ? ' with credentials' : ''}`);
+
+    const response = spawnSync('git', ['clone', cloneUrl, '--bare', '--progress'], {
```
coopernetes, line 43: I was recently bit by an issue using `spawnSync` on another Express application. If this process doesn't return in a timely manner, it can block the whole event loop including Express's HTTP routing/handlers. For a large repo, it's fair to assume this could take dozens of seconds or minutes to complete. I was also wrapping a call to `git` CLI and used the below pattern instead. Not sure if this will work for our usage of it here since we still have to "wait" for the clone to finish as part of the pull.

```javascript
const promisify = require('util').promisify;
const execFile = require('child_process').execFile;
const execFileAsync = promisify(execFile);

const exec = async (req, action) => {
  ...
  const { stdout } = await execFileAsync('git', ['clone', cloneUrl, '--bare', '--progress'], {
    cwd: action.proxyGitPath,
  });
  console.log(stdout);
}
```

src/proxy/processors/push-action/scanDiff.js
```diff
@@ -0,0 +1,99 @@
+const Step = require('../../actions').Step;
+const config = require('../../../config');
+
+const commitConfig = config.getCommitConfig();
+const privateOrganizations = config.getPrivateOrganizations();
```
coopernetes, line 5: These are new config sections. Can we add some example inputs or defaults into the [proxy.config.json](https://github.com/finos/git-proxy/blob/b326e26adf03475d8611c1b6d8f2a3c04d56ad51/proxy.config.json) at the repo root? Below patch is what I applied while testing.

```diff
diff --git a/proxy.config.json b/proxy.config.json
index 683fdf7..df8d6df 100644
--- a/proxy.config.json
+++ b/proxy.config.json
@@ -6,18 +6,36 @@
   },
   "authorisedList": [
     {
-      "project": "finos",
-      "name": "git-proxy",
-      "url": "https://github.com/finos/git-proxy.git"
+      "project": "coopernetes",
+      "name": "backstage",
+      "url": "https://github.com/coopernetes/backstage.git"
     }
   ],
+  "commitConfig": {
+    "message": {
+      "block": {
+        "literals": ["foo"],
+        "patterns": ["bar"]
+      }
+    },
+    "author": {
+      "email": {
+        "domain": {
+          "allow": ".*"
+        },
+        "local": {
+          "block": "foo"
+        }
+      }
+    }
+  },
   "sink": [
     {
       "type": "fs",
       "params": {
         "filepath": "./."
       },
-      "enabled": true
+      "enabled": false
     },
     {
       "type": "mongo",
@@ -25,7 +43,7 @@
       "options": {        
         "useUnifiedTopology": true
       },      
-      "enabled": false
+      "enabled": true
     }
   ],
   "authentication": [
diff --git a/src/config/index.js b/src/config/index.js
index e94f38c..c2b2db1 100644
--- a/src/config/index.js
+++ b/src/config/index.js
@@ -71,7 +71,37 @@ const logConfiguration = () => {
 };
 
 // logConfiguration();
-
+module.exports = {
+  getCommitConfig : () => {
+    return defaultSettings.commitConfig;
+  },
+  getPrivateOrganizations : () => {
+    return defaultSettings.privateOrganizations;
+  },
+  getAuthorisedList,
+  getDatabase,
tom@duncan:~/repos/git-proxy$ git diff src/config
diff --git a/src/config/index.js b/src/config/index.js
index e94f38c..c2b2db1 100644
--- a/src/config/index.js
+++ b/src/config/index.js
@@ -71,7 +71,37 @@ const logConfiguration = () => {
 };
 
 // logConfiguration();
-
+module.exports = {
+  getCommitConfig : () => {
+    return defaultSettings.commitConfig;
+  },
+  getPrivateOrganizations : () => {
+    return defaultSettings.privateOrganizations;
+  },
+  getAuthorisedList,
+  getDatabase,
+  logConfiguration,
+  getAuthentication,
+  getTempPasswordConfig,
+  getAllowSelfSignedCert : () => {
+    return true;
+  },
+  getProxyUrl : () => {
+    if (defaultSettings.proxyUrl) {
+      return defaultSettings.proxyUrl;
+    }
+    return 'https://github.com'
+  },
+  getCookieSecret : () => {
+    if (defaultSettings.cookieSecret) {
+      return defaultSettings.cookieSecret;
+    }
+    return 'secret';
+  },
+  getSessionMaxAgeHours : () => {
+    return 24;
+  },
+}
 exports.getAuthorisedList = getAuthorisedList;
 exports.getDatabase = getDatabase;
 exports.logConfiguration = logConfiguration;
```

src/proxy/processors/push-action/writePack.js
```diff
@@ -7,19 +7,21 @@ const exec = async (req, action) => {
     const cmd = `git receive-pack ${action.repoName}`;
     step.log(`executing ${cmd}`);
 
-    const content = execSync(`git receive-pack ${action.repoName}`, {
+    const content = spawnSync('git', ['receive-pack', action.repoName], {
```
coopernetes, line 10: See my previous comment on `spawnSync` use and whether we're OK blocking here or not. Probably needs the same `promisify(execFile)` pattern.

src/proxy/routes/index.js
```diff
@@ -1,17 +1,21 @@
...
 const express = require('express');
 const proxy = require('express-http-proxy');
-// eslint-disable-next-line new-cap
-const router = express.Router();
+const router = new express.Router();
 const chain = require('../chain');
+const config = require('../../config');
+
+if (config.getAllowSelfSignedCert()) {
+  process.env['NODE_TLS_REJECT_UNAUTHORIZED'] = 0;
```
coopernetes, line 8: I almost don't even want to support this option. For local development, testing or POC deployments, this can be set via the environment variable directly by those willing to compromise and bypass TLS validation. The preferred approach is using Node's existing TLS validation and having deployers use their own [`NODE_EXTRA_CA_CERTS`](https://nodejs.org/api/cli.html#node_extra_ca_certsfile) to trust either self-signed server certificates or their own custom roots/intermediate CAs. Wdyt?

src/proxy/routes/index.js
```diff
@@ -1,17 +1,21 @@
...
+}
 
 router.use(
   '/',
-  proxy('https://github.com', {
+  proxy(config.getProxyUrl(), {
```
coopernetes, line 13: I appreciate this being configurable, works well if you had to support flows to GitLab and other Git providers :+1: 

Maybe a separate feature but would this configuration ever be an array of URLs that all can be proxied by one git-proxy server simultaneously? The Express routes can be created as an array of URLs to pass to the proxy middleware.

The alternative approach can be to deploy separate servers with their own configured URLs, which could be a fair trade-off.

src/proxy/chain.js
```diff
@@ -3,11 +3,16 @@ const plugin = require('../plugin');
 
 const pushActionChain = [
   proc.push.parsePush,
+  proc.push.checkTraining,
```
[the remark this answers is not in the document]
JamieSlome: Sorted 👍

src/proxy/index.js
```diff
@@ -17,25 +14,9 @@ proxyApp.use(bodyParser.raw(options));
 proxyApp.use('/', routes);
 
 const start = async () => {
-  // Check to see if the default repos are in the repo list
-  const defaultAuthorisedRepoList = config.getAuthorisedList();
```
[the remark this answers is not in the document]
JamieSlome: Keeping ❤️

src/proxy/routes/index.js
JamieSlome: Agreed! 👍

src/proxy/routes/index.js
JamieSlome: @coopernetes - yes. I'd suggest we open this up as another issue 👍

src/proxy/processors/push-action/pullRemote.js
JamieSlome: I am not sure. If the function is asynchronous, we might be implementing a race condition where it is unclear whether the function call will end prior to the completion of the pull? I'd suggest that we leave it for now, but definitely open an issue to quickly swap out these `git` calls with a tried and tested library that serves as a wrapper around the `git` protocol.

Thoughts?

src/proxy/processors/push-action/writePack.js
JamieSlome: Same as my comments above, thoughts?

src/proxy/processors/push-action/parsePush.js
JamieSlome: @coopernetes - could you suggest a solution which encapsulates both a space-separated name and GitHub username at the zero-index? I think we agreed last time this is likely what is causing the issue of indexing for the rest of the properties.

src/proxy/processors/push-action/scanDiff.js
JamieSlome: @coopernetes - resolved 👍

JamieSlome:
@coopernetes @maoo - basically done with the full push now 👍 😌 Let's get this branch stable ❤️ 

Can you help figure out why the proxy and REST API aren't running properly on `8000` and `8080`, respectively?

src/proxy/processors/push-action/pullRemote.js
coopernetes: You should be able to validate this behaviour with trying to pull down a large repository. We saw this on [`backstage/backstage`](https://github.com/backstage/backstage) which is a couple GiB's. I suspect the current implementation will cause the server to stop responding to other requests such as health checks or API requests as it blocks the whole event loop waiting for `spawnSync` to return.

The [`spawn`](https://nodejs.org/api/child_process.html#child_processspawncommand-args-options) function can also be used but it relies on callbacks and is a bit more tricky to implement correctly since the `<cmd>.stdout.on(...)` callbacks happen asynchronously 😫

coopernetes:
> Can you help figure out why the proxy and REST API aren't running properly on `8000` and `8080`, respectively?

@JamieSlome I'm not seeing this behaviour. I was getting an error with the file-based db.

```
0] Proxy Listening on 8000
[0] /Users/544820491/code/opensource/git-proxy/src/service/index.js:29
[0]       store: db.getSessionStore(session),
[0]                 ^
[0] 
[0] TypeError: db.getSessionStore is not a function
[0]     at Object.start (/Users/544820491/code/opensource/git-proxy/src/service/index.js:29:17)
[0]     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
[0] 
[0] Node.js v20.10.0
```

After applying the below patch, both the proxy server & API is listening on 8000 & 8080 respectively.
diff
```
diff --git a/src/db/file/index.js b/src/db/file/index.js
index 583be50..1402fda 100644
--- a/src/db/file/index.js
+++ b/src/db/file/index.js
@@ -23,3 +23,6 @@ module.exports.removeUserCanPush = repo.removeUserCanPush;
 module.exports.removeUserCanAuthorise = repo.removeUserCanAuthorise;
 
 module.exports.deleteRepo = repo.deleteRepo;
+module.exports.getSessionStore = (session) => {
+  console.log(`File does not store sessions, no-op. ${session}`);
+};
```

```
[0] Proxy Listening on 8000
[0] File does not store sessions, no-op. function session(options) {
[0]   var opts = options || {}
...
[0] Service Listening on 8080
```

JamieSlome:
@coopernetes @maoo - when I try and logout, I get a 404. Could this be because of the HTTP filter or the new port configuration via ENV variables? Shouldn't port `8000` be porting to GitHub.com?

maoo:
> @coopernetes @maoo - when I try and logout, I get a 404. Could this be because of the HTTP filter or the new port configuration via ENV variables? Shouldn't port `8000` be porting to GitHub.com?

If I run `curl -X POST http://localhost:8080/api/auth/logout` , the server logs the following error and dies.

```
/Users/m/w/git-proxy/node_modules/passport/lib/sessionmanager.js:83
    req.session.regenerate(function(err) {
                ^

TypeError: Cannot read properties of undefined (reading 'regenerate')
    at Immediate.<anonymous> (/Users/m/w/git-proxy/node_modules/passport/lib/sessionmanager.js:83:17)
    at process.processImmediate (node:internal/timers:478:21)
```

I'll check what's going on.

maoo:
@JamieSlome - this worked for me - https://github.com/finos/git-proxy/compare/reconcile...fix-logout?expand=1

Can you please test on your end? It's 1-line fix, you don't need to switch branch.

maoo:
@JamieSlome - kicked Netlify - here's the preview - https://deploy-preview-335--endearing-brigadeiros-63f9d0.netlify.app/

JamieSlome:
> @JamieSlome - this worked for me - https://github.com/finos/git-proxy/compare/reconcile...fix-logout?expand=1
> 
> Can you please test on your end? It's 1-line fix, you don't need to switch branch.

Hmm, doesn't seem to work for me. I am trying to invoke logout from the UI?

JamieSlome:
```js

  const logout = () => {
    axios.post('/api/auth/logout').then((res) => {
      if (res.status === 204) {
        navigate('/admin/profile', { replace: true });
      }
    });
  };
```

@maoo - see the above code in `AdminNavbarLinks.jsx` 👍

JamieSlome:
@coopernetes - sorted 👍 Works a charm.

@maoo - I am now just working on fixing up logout and from a functionality perspective, we are done 🎉

JamieSlome:
> @JamieSlome - this worked for me - https://github.com/finos/git-proxy/compare/reconcile...fix-logout?expand=1
> 
> Can you please test on your end? It's 1-line fix, you don't need to switch branch.

Tried this fix, but doesn't actually log the user out, just prevents the error from bubbling 😢

JamieSlome:
@maoo @coopernetes - login and logout fixed 👍

JamieSlome:
@maoo @coopernetes - now only 2 unit tests away from passing all tests. Bit confused why these last two are failing. Any ideas?

```
  1) user creation
       should be able to create a new user:

      AssertionError: expected { Object (_events, _eventsCount, ...) } to have status code 200 but got 404
      + expected - actual

      -404

  2) user creation
       should access the profile:

      AssertionError: expected { Object (_events, _eventsCount, ...) } to have status code 200 but got 401
      + expected - actual

      -401
      +200
```

src/ui/services/user.js
```diff
@@ -24,46 +18,69 @@ const getUser = async (
...
-    .then(() => {})
-    .catch((error) => {
-      console.log(error.response.data.message);
-      throw error;
-    });
+  const url = new URL(`${baseUrl}/api/auth/gitAccount`);
+  await axios.post(url, data, { withCredentials: true }).catch((error) => {
+    console.log(error.response.data.message);
+    throw error;
+  });
```
coopernetes, line 73: @JamieSlome I believe these changes are causing the test failures.

JamieSlome: @coopernetes - thanks for spotting 👍 I'll `.skip()` the failing test for now and we can re-enable once we introduce the manual creation of new user records via the API.

src/proxy/processors/push-action/pullRemote.js
JamieSlome: @coopernetes - note that we should definitely open an issue for this and address post merge 👍 Are you able to create an issue and reference?

JamieSlome:
> > @coopernetes - any ideas how we can turn on / off the UI via the `npx` command?
> 
> Toggling the UI will likely require a runtime configuration flag. Alternatively, we can supply multiple `"bin"` values to support executing the git-proxy package with either the UI or without it. I don't think `npx` has any interaction with the running application unless we were to expose an API endpoint which shuts down the UI. That seems dubious.
> 
> ```diff
> diff --git a/package.json b/package.json
> index 28966b2..a7f56b1 100644
> --- a/package.json
> +++ b/package.json
> @@ -18,7 +18,10 @@
>      "lint": "eslint --fix . --ext .js,.jsx",
>      "gen-schema-doc": "node ./scripts/doc-schema.js"
>    },
> -  "bin": "./index.js",
> +  "bin": {
> +    "git-proxy": "./index.js",
> +    "fullstack": "concurrently 'node ./index.js' 'vite --config vite.config.js'"
> +  },
>    "author": "Paul Groves",
>    "license": "Apache-2.0",
>    "dependencies": {
> ```
> 
> The above would support this use case:
> 
> ```shell
> # No arguments, start just the proxy & API server
> $ npx @finos/git-proxy
> # with the "fullstack" command, runs all three
> $ npx @finos/git-proxy fullstack
> ```
> 
> Is that closer to what you're looking for?

Yes, definitely close to what I am looking for. Agree that having an endpoint to shutdown the UI is not particularly ideal. Multiple `bin` commands looks good 👍  Do you want to add to this branch or shall I update?

Also can we rename `fullstack` to `with-ui`? Something a little more descriptive?

JamieSlome:
@maoo - can you give your green tick review here as well? 👍 ✅

coopernetes (approved):
Minus the CSRF alert from CodeQL, LGTM 🎉

coopernetes:
This also closes #454 since the changes from related PR was integrated via https://github.com/finos/git-proxy/commit/6ff82d3a760d4f33ebc3673dcbd601ffa1cce9ed

package.json
```diff
@@ -15,31 +15,36 @@
...
-    "lint": "eslint --fix . --ext .js,.jsx",
+    "lint": "eslint \"src/**/*.{js,jsx,ts,tsx,json}\" \"test/**/*.{js,jsx,ts,tsx,json}\"",
+    "lint:fix": "eslint --fix \"src/**/*.{js,jsx,ts,tsx,json}\" \"test/**/*.{js,jsx,ts,tsx,json}\"",
+    "format": "prettier --write src/**/*.{js,jsx,ts,tsx,css,md,json,scss} test/**/*.{js,jsx,ts,tsx,json} --config ./.prettierrc",
     "gen-schema-doc": "node ./scripts/doc-schema.js"
   },
-  "bin": "./index.js",
+  "bin": {
+    "git-proxy": "./index.js",
+    "fullstack": "concurrently 'npm run server' 'npm run client'"
```
coopernetes, line 25: A better name here is needed. Testing this, it creates an executable with the literal name `fullstack`. It's not clear that this is from `@finos/git-proxy` and may cause more confusion. 

```bash
$ npm link
$ which fullstack
/home/tom/.asdf/shims/fullstack
```

Also, the executable cannot resolve `concurrently` for some reason and I wonder if this is just better ran from source until the whole application's entrypoint can be rewritten to allow these services to be selectively turned off. I do not believe `npm run start` is designed to be ran outside of the repo source since it relies on npm for setting the PATH to `./node_modules/.bin`.

This isn't really feasible for 1.x IMO.

I'm suggesting a better name at least but I suspect this executable will be challenging to get right with the current startup commands (running via `vite` and the current index.js entrypoint script).
```suggestion
     "git-proxy-all": "concurrently 'npm run server' 'npm run client'"
```

maoo: I'd like to propose using `gp` to shorten `git-proxy` , and therefore having `gp-all` instead of `git-proxy-all`, which coulld be longer to type. WDYT?

JamieSlome: What about `gp-shallow` and `gp-all`?

JamieSlome:
@maoo @coopernetes - I've discovered the correct pattern for applying CSRF and this now works 👍 The issue is that with CSRF protections, we'd need to make sure that any invocations against privileged endpoints are submitted with a CSRF token.

Are we comfortable with this?

coopernetes:
> I'd recommend that we make CSRF protections configurable. It is desirable to have CSRF protections with the UI. From my understanding CSRF is very much a browser-based vulnerability. I can add it as a flag to `proxy.config.json`.
> 
> Thoughts?

Makes sense. I suppose we can just [exclude the new middleware based on a config flag](https://github.com/finos/git-proxy/pull/335/files#diff-d56540c0251ece3369bdbb2f24e9e8ca1dc6ae74011fea5eaec5bf24b4b27464R48-R57)?

JamieSlome:
@coopernetes - the only thing is that I wouldn't want to support the usage of the UI without CSRF protections in place. Perhaps we couple the UI with CSRF protections and only allow disabling CSRF and UI together?

JamieSlome:
@maoo @coopernetes @msagi - this PR is now ready to go 🚀 

I have set CSRF protections as enabled by default in `proxy.config.json`. This can be turned off using the `csrfProtection` flag for anyone not interested in using the user interface (as CSRF is specifically a browser exploitable vulnerability).

I am happy to merge now and work on bug fixes as patch releases after the fact. A bit stumped by testing the two new `bin` commands with `npx` prior to merge. Any thoughts?

PR #379, opened 2024-01-01
feat: allow ports to be customised via env vars

### LINKED ISSUE #378: Make service ports configurable
Hard coded ports for the services cause a problem.

They can conflict with other services used locally, rendering the service useless in a broader setup.

Make all ports used by the service configurable.

Suggestion:

There could be two levels of configurations.
- configure by environment variable
- configure by the file config

the only alternative i can think of is running this in a Docker container but may be overkill

maoo:
Closes #378 

To test this PR:

```
npm install -g
export GIT_PROXY_UI_PORT="5000"
export GIT_PROXY_SERVER_PORT="9090"
git-proxy
```

You should see:
```
Service Listening on 5000
Listening on 9090
```

### DIFF
--- a/src/config/env.js
+++ b/src/config/env.js
@@ -1,6 +1,3 @@
-const {
-  GIT_PROXY_SERVER_PORT = 8000,
-  GIT_PROXY_UI_PORT = 8080
-} = process.env;
+const { GIT_PROXY_SERVER_PORT = 8000, GIT_PROXY_UI_PORT = 8080 } = process.env;
 
-exports.Vars = { GIT_PROXY_SERVER_PORT, GIT_PROXY_UI_PORT};
+exports.Vars = { GIT_PROXY_SERVER_PORT, GIT_PROXY_UI_PORT };
--- a/src/config/env.js
+++ b/src/config/env.js
@@ -1,0 +1,6 @@
+const {
+  GIT_PROXY_SERVER_PORT = 8000,
+  GIT_PROXY_UI_PORT = 8080
+} = process.env;
+
+exports.Vars = { GIT_PROXY_SERVER_PORT, GIT_PROXY_UI_PORT};
--- a/src/proxy/index.js
+++ b/src/proxy/index.js
@@ -4,7 +4,7 @@ const bodyParser = require('body-parser');
 const routes = require('./routes');
 const config = require('../config');
 const db = require('../db');
-const proxyHttpPort = process.env.GIT_PROXY_SERVER_PORT || 8000;
+const { GIT_PROXY_SERVER_PORT: proxyHttpPort } = require('../config/env').Vars;
 
 const options = {
   inflate: true,
@@ -33,7 +33,7 @@ const start = async () => {
   });
 
   proxyApp.listen(proxyHttpPort, () => {
-    console.log(`Listening on ${proxyHttpPort}`);
+    console.log(`!Listening on ${proxyHttpPort}`);
   });
 
   return proxyApp;
--- a/src/proxy/index.js
+++ b/src/proxy/index.js
@@ -4,7 +4,7 @@ const bodyParser = require('body-parser');
 const routes = require('./routes');
 const config = require('../config');
 const db = require('../db');
-const proxyHttpPort = 8000;
+const proxyHttpPort = process.env.GIT_PROXY_SERVER_PORT || 8000;
 
 const options = {
   inflate: true,
--- a/src/proxy/index.js
+++ b/src/proxy/index.js
@@ -33,7 +33,7 @@ const start = async () => {
   });
 
   proxyApp.listen(proxyHttpPort, () => {
-    console.log(`!Listening on ${proxyHttpPort}`);
+    console.log(`Listening on ${proxyHttpPort}`);
   });
 
   return proxyApp;
--- a/test/testConfig.js
+++ b/test/testConfig.js
@@ -10,7 +10,6 @@ const expect = chai.expect;
 describe('default configuration', function () {
   it('should use default values if no user-settings.json file exists', function () {
     const config = require('../src/config');
-    const env = require('../src/config/env');
 
     expect(config.getAuthentication()).to.be.eql(
       defaultSettings.authentication[0],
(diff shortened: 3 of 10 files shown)


src/proxy/index.js
```diff
@@ -4,7 +4,7 @@ const bodyParser = require('body-parser');
 const routes = require('./routes');
 const config = require('../config');
 const db = require('../db');
-const proxyHttpPort = 8000;
+const proxyHttpPort = process.env.GIT_PROXY_SERVER_PORT || 8000;
```
lwhiteley, line 7: Given there are more than 2 files affected here, it may be better to create a file to setup the environment variables and their defaults to avoid future mistakes

for eg.

```js
// file path: src/common/env.js

const {
  GIT_PROXY_SERVER_PORT = 8000,
  GIT_PROXY_UI_PORT = 8080
} = process.env

export { GIT_PROXY_SERVER_PORT, GIT_PROXY_UI_PORT}
```

then here:

```suggestion
const { GIT_PROXY_SERVER_PORT: proxyHttpPort } = require('../common/env')
```

similar for other files.

this way the defaults are only defined in one placeand easier to maintain.

website/docs/configuration/overview.mdx
```diff
@@ -43,6 +43,14 @@ Or with npx:
 npx -- @finos/git-proxy --config ./config.json
 ```
 
+### Change ports
+By default, Git Proxy uses port 8000 to expose the Git Server and 8080 for the UI frontend; 
+you can change ports by setting the `GIT_PROXY_SERVER_PORT` and `GIT_PROXY_UI_PORT` 
+environment variables.
+
+Note that `GIT_PROXY_UI_PORT` is needed for both server and UI Node processes, 
+whereas `GIT_PROXY_SERVER_PORT` is only needed by the server process.
```
lwhiteley, line 52: maybe also add an example of how to set the env vars via command line

lwhiteley (commented):
from the changes it looks good, 
i dont see any reason why it wouldnt work.

I wont be able to test it until next week either way.  [reviewing from phone]

maybe someone else can test to not hold it up

maoo:
@lwhiteley - all good suggestions, thanks! I've pushed changes following your suggestions.

@JamieSlome @coopernetes - please review/text at your earliest convenience. TY!

website/docs/configuration/overview.mdx
```diff
@@ -43,6 +43,19 @@ Or with npx:
 npx -- @finos/git-proxy --config ./config.json
 ```
 
+### Change ports
+By default, Git Proxy uses port 8000 to expose the Git Server and 8080 for the UI frontend;
```
lwhiteley, line 47: [optional]

```suggestion
By default, Git Proxy uses port 8000 to expose the Git Server and 8080 for the client application.
```

client application may sound better here

website/docs/configuration/overview.mdx
```diff
@@ -43,6 +43,19 @@ Or with npx:
 npx -- @finos/git-proxy --config ./config.json
 ```
 
+### Change ports
+By default, Git Proxy uses port 8000 to expose the Git Server and 8080 for the UI frontend; 
+you can change ports by setting the `GIT_PROXY_SERVER_PORT` and `GIT_PROXY_UI_PORT`
```
lwhiteley, line 48: [optional]

```suggestion
The ports can be changed by setting the `GIT_PROXY_SERVER_PORT` and `GIT_PROXY_UI_PORT` 
```

IMO: For documentation it is best to keep the language impersonal and objective..

so no `I`, `me` or `you`

src/proxy/index.js
```diff
@@ -33,7 +33,7 @@ const start = async () => {
   });
 
   proxyApp.listen(proxyHttpPort, () => {
-    console.log(`Listening on ${proxyHttpPort}`);
+    console.log(`!Listening on ${proxyHttpPort}`);
```
coopernetes, line 36: Possible typo?
```suggestion
    console.log(`Listening on ${proxyHttpPort}`);
```

coopernetes (approved):
LGTM - as a future enhancement, we should expose the server ports as both file-based config (via [proxy.config.json](https://github.com/finos/git-proxy/blob/main/proxy.config.json)) as well as environment variables. We should have this available for all our configurable values, merge or reconcile between the two sources and allow users to choose whichever method best suits their deployment to configure ports & other values.

JamieSlome:
Can we look to include some test functionality to demonstrate the usage and successful loading of ports via ENV?

website/docs/configuration/overview.mdx
```diff
@@ -43,6 +43,19 @@ Or with npx:
 npx -- @finos/git-proxy --config ./config.json
 ```
 
+### Change ports
```
JamieSlome, line 46:
```suggestion
### Set ports with ENV variables
```

src/config/env.js
```diff
@@ -0,0 +1,3 @@
+const { GIT_PROXY_SERVER_PORT = 8000, GIT_PROXY_UI_PORT = 8080 } = process.env;
```
JamieSlome, line 1: I assume this shorthand means that if object destructuring results in `undefined` or `null` values, then the `const` will be assigned the specific values written?

JamieSlome (commented):
Nice PR! 👍

I've made a couple of comments and mentioned in a previous commit that a test case would be great prior to merge.

Thank you @maoo @lwhiteley 🍰

src/config/env.js
lwhiteley: in general,

object detructuring default value is only set when the value is `undefined`. If a value is null in javascript, that means it was most likely intentional.

When an environment variable is not set, it will be `undefined`. Hence why this shorthand works.

JamieSlome: Makes sense, nice and clear clarification.

Thank you @lwhiteley 👍

maoo:
Added tests, but when testing custom ports, the test fails; I think that the problem is how I'm stubbing environment variables in `testConfig.js`

test/testConfig.js
```diff
@@ -21,6 +22,40 @@ describe('default configuration', function () {
...
+  const originalEnv = process.env;
+  before(function () {
+    process.env = {
+      ...originalEnv,
+      GIT_PROXY_SERVER_PORT: 9000,
+      GIT_PROXY_UI_PORT: 9090
+    };
+  });
+
+  it('should use custom ports defined by environment variables', function () {
```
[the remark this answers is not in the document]
lwhiteley: Also this is my opinion i guess, 

But this test may not be very useful.

if you would test if the service works as intended with setting env vars then maybe an integration/e2e test is better and ping the actual running services to see if they are started on the correct ports

else you are just trying to test how nodejs itself works via a unit test.

maoo:
Thanks @lwhiteley - that's the same code I did, and yes, the tutorial I followed used Jest; I didn't really think that mocha would not work.

Re. e2e testing, I agree with you, but the issue with env var setup would still be there.

I've tried several approaches, including https://medium.com/nodejsmadeeasy/elegant-ways-to-pass-env-variables-to-mocha-test-cases-4486cb238bb1 and https://glebbahmutov.com/blog/mocking-process-env/

I'll sleep on it and see if next week I can find a working solution. @JamieSlome - did you face this issue in the past with mocha? And is there any e2e testing where we can add the custom port testing?

lwhiteley:
Yup but I don't mean in a spec runner like jest or mocha where you have to manipulate the require cache.

It may be possible to setup jobs that simulate how the end user would test it in CI and just provide the env vars then confirm the ping.

test/testConfig.js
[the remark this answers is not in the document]
JamieSlome: Agree with the above @lwhiteley. Looking at the test you're right that it isn't doing too much more than evaluating whether the native .env functionality is working as expected.

I wonder if we can just verify local running processes and verify that the active ports are the ones specified by default or exposed via the test?

maoo:
I see that @coopernetes raised https://github.com/finos/git-proxy/issues/394 - so maybe we could simply skip the e2e part on this PR and tackle this challenge in a separate issue/PR ? WDYT?

JamieSlome:
@maoo - sounds good 👍 Feel free to remove the test(s) and we can get this merged 🍰

JamieSlome (approved):
Ready to merge 👍

maoo:
Thanks for stepping in @JamieSlome !

PR #409, opened 2024-01-22
fix(server): improved packfile encoding to support short messages and…

sammoorhouse:
… unicode in errors

This change allows you to say

```
step.setError(`🥺 Rejecting repo ${action.repo} not in the authorisedList`);
```

in a step.

Also fixes a padding bug that meant fewer than 5chars of errors broke the proxy response.

PR #415, opened 2024-01-24
feat: http filtering

### LINKED ISSUE #410: Filter HTTP requests that are not specific to git operations
Currently, git-proxy via its express http proxy middleware is acting as a transparent proxy to `https://github.com`. There is business logic in place that restricts certain actions a git client make take (push a commit to a repo, pull a repo, etc). However, when a user visits a Git Proxy deployed server via the browser, there is no restrictions in place due to some fallback logic. As such, Git Proxy's middleware will permit the request to the upstream web server (GitHub) via the browser.

Given that this is meant to enforce _git_ protections via pushes & pulls, Git Proxy should _not_ permit web access through a browser by default. In most scenarios, git-proxy is deployed behind a corporate firewall and this presents path out of an organization that is not captured, filtered or inspected by the current proxying. This can be used to bypass a network perimeter, exfiltrate data with other GitHub services like gists, etc.

1. When a user is pushing or pulling code through Git Proxy using any compatible git client, the proxy operates as it certain does and uses the list of Actions to apply filtering and protections on git interactions.
2. When a user visits a Git Proxy server in the browser to attempt to browse GitHub via https://git-proxy.example.com, they should receive an error.

Git Proxy can filter out the following properties of the inbound HTTP request:
* The server should perform content negotiation[^1] and reject requests that _do not_ contain an `Accept` header matching the content types of  `application/x-git-*` (multiple values are used[^2][^3]). Clients making requests from the browser with an Accept header of `text/html` should be rejected.
* The server should only permit requests matching known "safe URL paths" used by git clients such as `/info/refs?service=git-receive-pack` & `/git-receive/pack` for pushes and `/info/refs?service=git-upload-pack` & `/git-upload-pack` for pulls.
* The error response should be a 400 Bad Request and return some minimal information such as "invalid request". The response should be minimal so as to not [introduce potential security issues](https://owasp.org/www-community/Improper_Error_Handling). At most, the server can respond with a 415 Unsupported Media Type.

You could deploy a reverse proxy in front of Git Proxy to handle this filter (via nginx). However, given that git-proxy is "git aware" and applies application-level filtering on git operations, it should apply this as a default.

Other networking solutions likely are available to apply filtering at the infrastructure layer but those appliances generally cannot implement git application logic to restrict access. Git Proxy is typically deployed as a trusted system and is permitted through firewalls, etc.

**Additional context**
N/A

[^1]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation
[^2]: https://www.git-scm.com/docs/http-protocol
[^3]: https://github.com/git/git/blob/master/http-backend.c

coopernetes:
Fixes #410 

(Hello from Git Proxy 👋)

### DIFF
--- a/src/proxy/routes/index.js
+++ b/src/proxy/routes/index.js
@@ -6,43 +6,38 @@ const router = express.Router();
 const chain = require('../chain');
 
 /**
- * Check whether an HTTP request matches the list of permitted User-Agent strings.
- * @param {string} agent the User-Agent string from the http request
- * @return {boolean} whether the User-Agent is allowed
+ * Check whether an HTTP request is a Git request. It is a Git request if it
+ * matches one of the following:
+ * - the URL path is a well-known Git endpoint
+ * - the User-Agent starts with "git/"
+ * - the Accept header starts with "application/x-git-" (for specific paths)
+ * @param {express.Request} req Raw HTTP request
+ * @return {boolean} Whether the request is a Git request
  */
-const allowedUserAgent = (agent) => {
-  return agent.startsWith('git/');
-};
-
-/**
- * As part of content negotiation, the client sends an Accept header to indicate
- * the media types it understands. This function checks whether the media type
- * is allowed.
- * @param {string} contentType Received media type from the request Accept header
- * @return {boolean} Whether the media type is allowed
- */
-const allowedContentType = (contentType) => {
-  return contentType.startsWith('application/x-git-');
-};
-
-/**
- * Test whether the request URL is allowed for proxying.
- * @param {string} url the request URL (path)
- * @return {boolean} whether the URL is allowed
- */
-const allowedUrl = (url) => {
-  const safePaths = [
-    '/info/refs?service=git-upload-pack',
-    '/info/refs?service=git-receive-pack',
-    '/git-upload-pack',
-    '/git-receive-pack',
-  ];
+const validGitRequest = (req) => {
+  const { 'user-agent': agent, accept } = req.headers;
+  const { url } = req;
   const parts = url.split('/');
+  // only valid GitHub URLs are supported.
+  // url = '/{owner}/{repo}.git/{git-path}'
+  // url.split('/') = ['', '{owner}', '{repo}.git', '{git-path}']
+  if (parts.length !== 4 && parts.length !== 5) {
+    return false;
+  }
+  parts.splice(1, 2); // remove the {owner} and {repo} from the array
+  const gitPath = parts.join('/');
   if (
-    (parts.length === 4 || parts.length === 5) &&
-    Boolean(safePaths.find((path) => url.endsWith(path)))
+    gitPath === '/info/refs?service=git-upload-pack' ||
+    gitPath === '/info/refs?service=git-receive-pack'
   ) {
-    return true;
+    // https://www.git-scm.com/docs/http-protocol#_discovering_references
+    // We can only filter based on User-Agent since the Accept header is not
+    // sent in this request
+    return agent.startsWith('git/');
+  }
+  if (gitPath === '/git-upload-pack' || gitPath === '/git-receive-pack') {
+    // https://www.git-scm.com/docs/http-protocol#_uploading_data
+    return agent.startsWith('git/') && accept.startsWith('application/x-git-');
   }
   return false;
 };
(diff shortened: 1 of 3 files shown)


src/proxy/routes/index.js
```diff
@@ -5,13 +5,55 @@ const proxy = require('express-http-proxy');
...
+  // url = '/{owner}/{repo}.git/{git-path}'
+  // url.split('/') = ['', '{owner}', '{repo}.git', '{git-path}']
+  if (parts.length !== 4 && parts.length !== 5) {
+    return false;
+  }
+  parts.splice(1, 2); // remove the {owner} and {repo} from the array
+  const gitPath = parts.join('/');
+  if (
+    gitPath === '/info/refs?service=git-upload-pack' ||
+    gitPath === '/info/refs?service=git-receive-pack'
```
JamieSlome, line 31:
```
['/info/refs?service=git-upload-pack', '/info/refs?service=git-receive-pack'].includes(gitPath)
```
Shorthand?

src/proxy/routes/index.js
```diff
@@ -5,13 +5,55 @@ const proxy = require('express-http-proxy');
...
+  ) {
+    // https://www.git-scm.com/docs/http-protocol#_discovering_references
+    // We can only filter based on User-Agent since the Accept header is not
+    // sent in this request
+    return agent.startsWith('git/');
+  }
+  if (gitPath === '/git-upload-pack' || gitPath === '/git-receive-pack') {
```
JamieSlome, line 38: Same shorthand as above? What do you prefer in terms of readability?

src/proxy/routes/index.js
```diff
@@ -5,13 +5,55 @@ const proxy = require('express-http-proxy');
...
+/**
+ * Check whether an HTTP request is a Git request. It is a Git request if it
+ * matches one of the following:
+ * - the URL path is a well-known Git endpoint
+ * - the User-Agent starts with "git/"
+ * - the Accept header starts with "application/x-git-" (for specific paths)
+ * @param {express.Request} req Raw HTTP request
+ * @return {boolean} Whether the request is a Git request
+ */
+const validGitRequest = (req) => {
```
JamieSlome, line 17: If specific to GitHub, should we name it accordingly?

JamieSlome (commented):
I've left a couple of comments; otherwise very straight forward and solid implementation 👍 ❤️ 🍰 

I love my emojis...

src/proxy/routes/index.js
coopernetes: Good point. I think having a different "URL path sanitizer" function which is specific to GitHub is probably warranted. Other than parsing the path, this is generic to git.

test/testRouteFilter.js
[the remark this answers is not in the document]
coopernetes: I think we have decent coverage but I'm gonna add a few additional test cases for other combinations of headers and paths.

JamieSlome:
@coopernetes - can we resolve merge conflicts? I merged #409 which also had changes to the router in `src/proxy/index.js`?

JamieSlome (approved):
LGTM! 🍰 

Just need the merge conflict resolved and good to go!

JamieSlome:
Oh, CodeCov how we love you so. @coopernetes - happy to bump up the coverage before we merge?

We should probably make the project level Action informational and the PR level breaking whilst we slowly bring up the coverage for the rest of the repository.

coopernetes:
> Oh, Codecov how we love you so. @coopernetes - happy to bump up the coverage before we merge?
> 

Most of the required refactoring for getting additional coverage is not something I want to do in this PR. I would essentially have to rewrite the Express middleware & routing layer to abstract it into a set of testable functions. That's probably effort better spent on changes to move us to 2.x. 

> We should probably make the project level Action informational and the PR level breaking whilst we slowly bring up the coverage for the rest of the repository.

Let me see what codecov makes available in their configuration. I would like to preserve the PR comment for the report even if the status check isn't blocking builds. This gives needed feedback in the PR while we improve overall coverage. 

https://docs.codecov.com/docs/codecovyml-reference

coopernetes:
Looking at the details, I think we can retest this after #429 is merged. The "baseline" report is misreported against a different branch. 

https://app.codecov.io/gh/finos/git-proxy/tree/bugfix%2Fci-permissions

JamieSlome:
@coopernetes - agree with your comment before last. #429 is now merged by the way 👌

JamieSlome:
@coopernetes - as long as we can keep a view of the coverage of suggested changes in the PR that is good enough for now.

JamieSlome:
[Looking at Codecov](https://app.codecov.io/gh/finos/git-proxy/pull/415), it looks like it is already achieving this. We just don't want to return a failing state to CI.

coopernetes:
Now that #442 is merged, let me rebase again and retest. @JamieSlome are you good with the latest changes? Hoping to get this merged soon.

JamieSlome (approved):
LGTM @coopernetes ❤️ 

Would like to get an idea of the coverage percentage but happy to merge in any case. Will keep my eyes open for your re-base.

coopernetes:
> LGTM @coopernetes ❤️
> 
> Would like to get an idea of the coverage percentage but happy to merge in any case. Will keep my eyes open for your re-base.

`codecov/project — 57.15% (+0.30%) compared to e5bb34a` 😅

Issue #203, opened 2023-07-12
Implement code coverage reports on pull requests 🔍


JamieSlome:
To encourage contribution best practices, we should run coverage checks on critical logic in the library against our current test suite.

```[tasklist]
### Tasks
- [ ] Use `nyc` to generate test coverage reports
- [ ] Assess coverage reports on new code only (if possible)
- [ ] Implement a GitHub Action to present and block PRs if coverage on new logic is below 80%
- [ ] Rename test files to `*.test.js`
```

JamieSlome:
This Action could be a nice display function for the result of a code coverage report:
https://github.com/marketplace/actions/code-coverage-report

CyberCitizen01:
Hi @JamieSlome 👋🏻

I'm thrilled to explore the opportunity to contribute to the project as a first-time contributor. I noticed that this issue has been labelled as a "Good First Issue," and I'm eager to take on the challenge.

As a newcomer, I'm committed to learning and adhering to the project's conventions and guidelines. While I've already reviewed the contributing guidelines in the repository, I was hoping to gain further insight into the specific requirements and expectations for this particular issue.

Could you kindly provide any additional information or context that would assist me in tackling this "Good First Issue"? I'm excited to demonstrate my skills and make a positive impact on the project.

In the meantime, I'll be actively preparing and researching to ensure I approach the task with the best possible understanding.

Thank you for your time and consideration. I'm looking forward to collaborating with the team and contributing meaningfully to the project.

JamieSlome:
@CyberCitizen01, thanks for getting involved!

I'd recommend starting with reading [this article](https://www.atlassian.com/continuous-delivery/software-testing/code-coverage) as it should give you a good overview of what code coverage is, and why it is important.

Next, I'd take a look through [Istanbul's code coverage documentation](https://istanbul.js.org/). This will give you a good idea of how to run code coverage against our codebase, including how to install it.

Let me know if there is anything else I can help you with 👍

CyberCitizen01:
Thank you @JamieSlome! I'll take a close look at these and get started right away.

Also if you can assign the issue to me, I would be really delighted. Thanks!

JamieSlome:
@CyberCitizen01 - thanks for your PR! Does your contribution address:

"Assess coverage reports on new code only (if possible)"

It would be nice to show contributors the coverage they are missing rather than the coverage for the entire repository every time. Let me know your feedback and thoughts! 👍

[...]

PR #429, opened 2024-01-30
fix: use codecov for test coverage reports

coopernetes:
Fixes: #430

Related: #203

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -56,19 +56,12 @@ jobs:
           echo "isNotEmpty=false" >> $GITHUB_OUTPUT
         fi
 
-    - name: Install LCOV
+    - name: Upload test coverage report
       if: ${{ steps.check-lcov-non-empty.outputs.isNotEmpty == 'true'}}
-      run: |
-        sudo apt-get update && sudo apt-get install --assume-yes lcov
-
-    - name: Report code coverage
-      if: ${{ steps.check-lcov-non-empty.outputs.isNotEmpty == 'true' }}
-      uses: zgosalvez/github-actions-report-lcov@v4
+      uses: codecov/codecov-action@v3.1.5
       with:
-        coverage-files: ./coverage/lcov.info
-        github-token: ${{ secrets.GITHUB_TOKEN }}
-        update-comment: true
-
+        files: ./coverage/lcov.info
+        flags: unittests
     # - name: Exit if coverage condition not met
     #   if: ${{ steps.test.outputs.exit_code }} != 0
     #   run: exit ${{ steps.test.outputs.exit_code }}
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -9,6 +9,9 @@ on:
   pull_request:
     branches: [ main ]
 
+permissions:
+  pull-requests: write
+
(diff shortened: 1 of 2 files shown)


src/db/index.js
```diff
@@ -2,6 +2,7 @@ const nodemailer = require('nodemailer');
 const generator = require('generate-password');
 const passwordHash = require('password-hash');
 const validator = require('email-validator');
+
```
JamieSlome, line 5: Linting?

.github/workflows/nodejs.yml
```diff
@@ -56,19 +56,12 @@ jobs:
...
-      if: ${{ steps.check-lcov-non-empty.outputs.isNotEmpty == 'true' }}
-      uses: zgosalvez/github-actions-report-lcov@v4
+      uses: codecov/codecov-action@v3.1.5
       with:
-        coverage-files: ./coverage/lcov.info
-        github-token: ${{ secrets.GITHUB_TOKEN }}
-        update-comment: true
-
+        files: ./coverage/lcov.info
+        flags: unittests
```
JamieSlome, line 64: What is the fix here?

coopernetes: Converting this action over to CodeCov. We have codecov in use on other FINOS repos so in theory, we're bypassing the GitHub permission restrictions. By default, any PR from a forked repository cannot execute write actions against a PR (without contorting the workflow, separating out checkout+test from the coverage reporting which is pretty kludgy). 

I can't really tell from our angle if this is possible though. I need @maoo or @finos-admin to confirm whether the codecov app is available for our repo and if not, do the necessary setup to make it so.

src/db/index.js
coopernetes: Just to trigger a coverage report to be generated. It can be reverted before we merge.

.github/workflows/nodejs.yml
JamieSlome: @maoo - can we please take a look at the permissions for CodeCov against our repository?

maoo: > @maoo - can we please take a look at the permissions for CodeCov against our repository?

There was no permission; I've setup codecov and added a `CODECOV_TOKEN` secret on this repo.

JamieSlome: @coopernetes - awesome, are we good to merge?

src/db/index.js
JamieSlome: @coopernetes - can revert?

.github/workflows/nodejs.yml
```diff
@@ -9,6 +9,9 @@ on:
   pull_request:
     branches: [ main ]
 
+permissions:
+  pull-requests: write
+
```
coopernetes, line 14: I don't believe this is needed any longer.
```suggestion
```

src/db/index.js
coopernetes, line 5: Revert this before merging.
```suggestion
```

JamieSlome:
@coopernetes - once we've got the branch updated, I think we are good to merge?

coopernetes:
@JamieSlome good to go for final review & merge.

JamieSlome (approved):
LGTM! 🍰

PR #451, opened 2024-02-26
Replace usage of `activeClassName` in `react-router-dom` v6 ⬆️ 

maoo:


src/ui/views/User/User.jsx
```diff
@@ -35,9 +35,11 @@ export default function Dashboard(props) {
   const navigate = useNavigate();
 
   useEffect(() => {
-    const id = props.match.params.id;
-    if (id == null) {
+    let id = null;
+    if (!props.match) {
       setIsProfile(true);
+    } else {
+      id = props.match.params.id;
```
JamieSlome, line 42: `props` doesn't exist here anymore. The `useParams()` function is required from `react-router-dom` 👍

JamieSlome:
@maoo - can you take care of the merge conflicts?

src/ui/views/User/User.jsx
maoo: Is this what you mean?

```
  const params = useParams();

  useEffect(() => {
    let id = null;
    if (!params.match) {
      setIsProfile(true);
    } else {
      id = params.match.params.id;
    }
```

maoo: saw your fix on the reconcile branch. Thank you!

maoo:
> @maoo - can you take care of the merge conflicts?

done!

src/ui/components/Sidebar/Sidebar.jsx
```diff
@@ -39,7 +39,7 @@ export default function Sidebar(props) {
           <NavLink
             to={prop.layout + prop.path}
             className={activePro + classes.item}
-            activeClassName='active'
+            activeclassname='active'
```
JamieSlome, line 42: I think the fix here is actually to attach the `exact` property and keep the usage of `activeClassName` the same?

Ref: https://stackoverflow.com/questions/40218050/react-router-link-activeclassname-not-working

maoo: > I think the fix here is actually to attach the `exact` property and keep the usage of `activeClassName` the same?
> 
> Ref: https://stackoverflow.com/questions/40218050/react-router-link-activeclassname-not-working

This post is quite old; my understanding is that `activeClassName` was deprecated, see https://dev.to/gabrlcj/react-router-v6-some-of-the-new-changes-181m#navlink-activeclassname-prop-does-not-exists-anymore

JamieSlome: @maoo - from the article, it looks like we want the following:

```js
className={(navData) => navData.isActive ? "active" : "" }
```

src/ui/views/Login/Login.jsx
```diff
@@ -79,13 +79,13 @@ export default function UserProfile() {
   return (
     <form onSubmit={handleSubmit}>
       <Snackbar
-        open={message}
+        open={true}
```
JamieSlome, line 82: This doesn't seem correct? This means the `<Snackbar />` will always appear?

src/ui/components/Sidebar/Sidebar.jsx
JamieSlome: > @maoo - from the article, it looks like we want the following:
> 
> ```js
> className={(navData) => navData.isActive ? "active" : "" }
> ```

Just tried, doesn't work...

JamieSlome: @maoo - can we redact this change for now as this doesn't fix the issue and we can look to put into another issue and merge your other changes?

JamieSlome:
@maoo - I've introduced the changes directly to `reconcile` 👍 Thank you ❤️ 

We can now focus on the `activeClassName` issue here...

JamieSlome:
As referenced in the commit message above, `activeClassName` is no longer required and the `<Sidebar />` component is working as expected.

Closing 👍 🎉

PR #459, opened 2024-02-28
Create CVE ignore list in scanning workflow

JamieSlome:
Addresses failing PRs due to false positive CVE in `IP` package `v1.1.8`


JamieSlome:
Ignore list is no longer required. Fix available via `npm audit fix` 👍

Issue #492, opened 2024-03-22
Error: CSRF token missing 🐛 


JamieSlome:
**Describe the bug**

When logging in via the UI, the following error is received:

```bash
Error: CSRF token missing
 at checkCsrf (.../git-proxy/node_modules/lusca/lib/csrf.js:169:18)
```

**To Reproduce**
1. Checkout to the `reconcile` branch
2. Go to `/login` via the UI
3. Enter `admin` & `admin` for username and password credentials
4. Submit login form
5. View locally running logs on dev machine

**Additional context**

CSRF added in PR: https://github.com/finos/git-proxy/pull/462

JamieSlome:
@vaibssingh - heads up on the above ⤴️ 👍

JamieSlome:
@vaibssingh - reopening this. Are we able to take a look at this issue on the `reconcile` branch and open a pull request pointing there instead of `main`?

Thanks for your time today ❤️

JamieSlome:
Reference: https://github.com/nextauthjs/next-auth/issues/88

JamieSlome:
This blog has been the most useful in getting CSRF setup, I am getting ever closer:

https://medium.com/@sakshi_13861/in-depth-guide-to-counteracting-cross-site-request-forgery-csrf-in-expressjs-8fbe3c36691d

JamieSlome:
Closing as no longer an issue 👍 Thanks @vaibssingh for the heads up in our catch up.

PR #497, opened 2024-03-22
Re-introduce CSRF protection with Lusca ❤️ 

JamieSlome:
Closes #492 

@vaibssingh - the fix was to move the CSRF exposure lower into `app.use()` call chain 👍

PR #499, opened 2024-03-22
Re-introduce CSRF protection with Lusca ❤️ #497

JamieSlome:
Closes https://github.com/finos/git-proxy/issues/492


JamieSlome:
@maoo - it seems like the checks are still running on the repository? Any ideas?

PR #508, opened 2024-03-30
Replace console.log with structured logging 

### LINKED ISSUE #399: Replace console.log with structured logging
Git-proxy does a lot "under the hood". For operators, it would be beneficial for git-proxy to emit messages using structured logging using a library such as [`winston`](https://github.com/winstonjs/winston) to better indicate the state of requests flowing through the system.

LewisBroadhurst:
Closes #399 

Winston logger has been used to replace the console.log() node logging in the git-proxy application. Logs generated from git-proxy use now have a format of `[timestamp] [log level]: [log message]`, and can be found in the files `git-proxy.log` and `error.log` under the `src/logging` directory. 

Before merging, I wanted to confirm that the old console.log() lines should be removed also. Maybe they are useful in some cases?


JamieSlome:
@LewisBroadhurst - thanks for the pull request - you're a star ⭐ 

In terms of keeping the old `console.log` calls, is there anything this would provide us that the usage of `winston` does not?

scripts/prepare.js
```diff
@@ -27,7 +28,7 @@ const CLI_HUSKY = 'npx husky install';
 
 // this will create .husky/_/husky.sh if it does not yet exist.
 if (!existsSync(FILE_HUSKY)) {
-  global.console.log(CLI_HUSKY);
+  logger.info(CLI_HUSKY);
```
JamieSlome, line 31: Do we want to make this change? Surely we want the Winston logger to be used for Git Proxy related services only not necessarily development tooling?

scripts/prepare.js
```diff
@@ -37,6 +38,6 @@ if (!existsSync(FILE_HUSKY)) {
 
 // this will create .husky/commit-msg if it does not yet exist.
 if (!existsSync(FILE_COMMIT)) {
-  global.console.log(CLI_COMMIT);
+  logger.info(CLI_COMMIT);
```
JamieSlome, line 41: Do we want to make this change? Surely we want the Winston logger to be used for Git Proxy related services only not necessarily development tooling?

JamieSlome (commented):
Looking good! 🍰 

Can we think about whether we should use `winston` logging in `log()` in `Step.js`:

```
log(message) {
    const m = `${this.stepName} - ${message}`;
    this.logs.push(m);
    console.info(m);
}
```

LewisBroadhurst:
> Looking good! 🍰
> 
> Can we think about whether we should use `winston` logging in `log()` in `Step.js`:
> 
> ```
> log(message) {
>     const m = `${this.stepName} - ${message}`;
>     this.logs.push(m);
>     console.info(m);
> }
> ```

I think this is an example of keeping the console.log I mentioned initially, as having the values logged to console will be a much better UX? (and an npx installer wouldn't have log access?)

For this would the best solution be adding the winston logging in to the step, as well as keeping the console.log? Then we get the UX positives and logging.

Can remove the double logging from **checkRepoInAuthorisedList.js** with this too 😄

```diff
step.log(`repo ${action.repo} is not in the authorisedList, ending`);
- logger.info('setting error');

step.setError(`Rejecting repo ${action.repo} not in the authorisedList`);
- logger.error(`Rejecting repo ${action.repo} not in the authorisedList`);
```

```diff
+ log(message, isError = false) {
     const m = `${this.stepName} - ${message}`;
     this.logs.push(m);
+   if (!isError) {
+     console.info(m);
+     logger.info(m);
+   } else {
+     console.error(m);
+     logger.error(m);
+   }
```

coopernetes (commented):
Small suggestion on winston setup.

LewisBroadhurst:
Hey @coopernetes - I remember you mentioning about using winston logger in production and wanted your take on adding a filename/path to a log. The naming is off as I just wanted to try it out without ripping everything up. 

```javascript
const logger = createLogger({
  format: combine(timestamp(), format.errors({ stack: true }), cleanFormat),
  transports: [
    new transports.File({
      level: 'error',
      filename: 'error.log',
      dirname: './src/logging',
    }),
    new transports.File({
      level: 'info',
      filename: 'git-proxy.log',
      dirname: './src/logging',
    }),
    new transports.Console(),
  ],
});

/**
 * @param {string} logLevel
 * @param {string} filename
 * @param {string} message
 */
function winstonLogger(logLevel, filename, message) {
  if (logLevel === 'error') {
    logger.error(`${filename}: ${message}`);
  } else {
    logger.info(`${filename}: ${message}`);
  }
}

module.exports.logger = logger;
module.exports.winstonLogger = winstonLogger;

// in js now call...
winstonLogger('info', 'service/index.js', `Service Listening on ${uiPort}`);

// produces the log...
2024-04-11T20:15:55.575Z info: service/index.js: Service Listening on 8080
```

I've seen a different approach mentioned a few places: `const logger = require('./logger')(module);` too, e.g. https://stackoverflow.com/questions/13410754/i-want-to-display-the-file-name-in-the-log-statement.

Did you follow either of these patterns/find success with a different pattern?

coopernetes:
@LewisBroadhurst are you able to address the conflicts?

coopernetes:
This PR has been open for about 6 months now. I'm going to close it until the upstream changes can be merged back in. Feel free to re-open so these changes can be pulled in.

Issue #509, opened 2024-04-01
Current config file in repo fails git-proxy --verify test


msagi:
**Describe the bug**
The existing proxy.config.json file is not valid as per the config schema in the repository

**To Reproduce**
```
 % npx -- @finos/git-proxy --validate

/Users/dev/git/git-proxy/node_modules/jsonschema/lib/helpers.js:61
    throw err;
    ^
ValidationError {
  path: [],
  property: 'instance',
  message: 'is not allowed to have the additional property "tempPassword"',
  schema: 'https://git-proxy.finos.org/config.schema.json',
  instance: {
    tempPassword: { sendEmail: false, emailConfig: {} },
    authorisedList: [
      {
        project: 'finos',
        name: 'git-proxy',
        url: 'https://github.com/finos/git-proxy.git'
      }
    ],
    sink: [
      { type: 'fs', params: { filepath: './.' }, enabled: true },
      {
        type: 'mongo',
        connectionString: 'mongodb://localhost:27017/gitproxy',
        options: { useUnifiedTopology: true },
        enabled: false
      }
    ],
    authentication: [ { type: 'local', enabled: true } ]
  },
  name: 'additionalProperties',
  argument: 'tempPassword',
  stack: 'instance is not allowed to have the additional property "tempPassword"'
}

Node.js v21.7.1

```

**Expected behavior**
```
 % npx -- @finos/git-proxy --validate
proxy.config.json is valid
```

**Screenshots**
n/a

**Desktop (please complete the following information):**
 - OS: macOS 14.4.1 (23E224)
 - Browser n/a
 - Version n/a

**Smartphone (please complete the following information):**
 - Device: n/a
 - OS: n/a
 - Browser n/a
 - Version n/a

**Additional context**
n/a

msagi:
hi @JamieSlome , I can take this as a quick one

msagi:
fixed

PR #526, opened 2024-04-13
fix: config schema

msagi:
Fix for #509.

- Updated the schema which did not seem to be correct as per the schema spec ("oneOf" should go under "properties" not the other way around).
- Moved config file validation logic into the config file module for improved encapsulation
- Added a few positive and negative test scenarios for config files

@JamieSlome: should GitProxy always validate the config file when starting? It seems it only runs the validation if -v/--validate is present.


JamieSlome:
@msagi - can we create a `fixtures` folder in the `test` folder to store all of the mock data?

JamieSlome (approved):
LGTM! 🍰 

Would like to get @coopernetes eyes on this as he pulled together this initial process 👍

coopernetes (approved):

PR #545, opened 2024-05-04
fix: add gitlab support

### LINKED ISSUE #511: Git Proxy does not support GitLab repos
**Describe the bug**
Failed to do git push with a GitLab repo.

**To Reproduce**
1. unzip [gitlab-issue.tar.gz](https://github.com/finos/git-proxy/files/14829449/gitlab-issue.tar.gz)
2. start git-proxy with the config file (proxy.config.json) from the .tar.gz file
3. execute git-proxy-issue-repro.sh
4. compare the *.run.log files with your run logs
5. compare .data/db files with your data

**Expected behavior**
git push is successful (with GitLab repo)

**Screenshots**
n/a

- OS: macOS 14.4.1 (23E224)
- Git: git version 2.39.3 (Apple Git-146)
- Browser n/a
- Version n/a

**Additional context**
src/proxy/actions/Action.js: 56 
    this.url = `https://github.com/${repo}`;

this might be the cause of the issue: GitHub is hardcoded baseURL into the action

msagi:
Fix #511 .

This is my first stab at adding multi host support to GitProxy. The main idea behind it is: not to lose the original git host during repointing the repo to GitProxy via `git remote set-url` command but keep the it in the repointed url as the first path segment, e.g.

```
$git remote -v
origin	git@github.com:finos/git-proxy.git (fetch)
origin	git@github.com:finos/git-proxy.git (push)

$git remote set-url origin https://localhost:8000/github.com/finos/git-proxy.git

$git remote -v
origin	http://localhost:8000/github.com/finos/git-proxy.git (fetch)
origin	http://localhost:8000/github.com/finos/git-proxy.git (push)
```

Since the original git host is preserved, specific routes can be configured (in ```proxy.config.json```) connecting the first path segment to the original host, e.g.

```
  "proxyConfig": [
    {
      "path": "/github.com",
      "url": "https://github.com",
      "enabled": true
    },
    {
      "path": "/gitlab.com",
      "url": "https://gitlab.com",
      "enabled": true
    }
  ],
```

I've added GitHub and GitLab repos to the authorisedList and tested ```git push``` with both host, works fine.

GitHub.com is hardcoded in several places, and whilst the proxy now supports GitHub and GitLab too, perhaps we shall sit down and chat how we can evolve the UI component (e.g., GitHub account is hardcoded, etc.) forward.

Note: due to multi host support, the `project/name` is not unique ID any more. A repo can uniquely identified only by its URL, e.g. `https://gitlab.com/finos/git-proxy.git`

@abinash2512 @JamieSlome - let me know your thoughts

### DIFF
--- a/config.schema.json
+++ b/config.schema.json
@@ -5,7 +5,13 @@
   "description": "Configuration for customizing git-proxy",
   "type": "object",
   "properties": {
-    "proxyUrl": { "type": "string" },
+    "proxyConfig": { 
+      "description": "Proxy configuration",
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/proxy"
+      }
+    },
     "cookieSecret": { "type": "string" },
     "sessionMaxAgeHours": { "type": "number" },
     "api": {
@@ -70,6 +76,15 @@
     }
   },
   "definitions": {
+    "proxy": {
+      "type": "object",
+      "properties": {
+        "path": { "type": "string"},
+        "url": { "type": "string"},
+        "enabled": { "type": "boolean" }
+      },
+      "required": ["path", "url", "enabled"]
+    },
     "authorisedRepo": {
       "type": "object",
       "properties": {
--- a/proxy.config.json
+++ b/proxy.config.json
@@ -1,5 +1,16 @@
 {
-  "proxyUrl": "https://github.com",
+  "proxyConfig": [
+    {
+      "path": "/github.com",
+      "url": "https://github.com",
+      "enabled": true
+    },
+    {
+      "path": "/gitlab.com",
+      "url": "https://gitlab.com",
+      "enabled": true
+    }
+  ],
   "cookieSecret": "cookie secret",
   "sessionMaxAgeHours": 12,
   "tempPassword": {
--- a/src/db/file/pushes.js
+++ b/src/db/file/pushes.js
@@ -94,8 +94,8 @@ const cancel = async (id, logger) => {
 const canUserCancelPush = async (id, user) => {
   return new Promise(async (resolve) => {
     const pushDetail = await getPush(id);
-    const repoName = pushDetail.repoName.replace('.git', '');
-    const isAllowed = await repo.isUserPushAllowed(repoName, user);
+    const repoUrl = pushDetail.repo.url;
+    const isAllowed = await repo.isUserPushAllowed(repoUrl, user);
 
     if (isAllowed) {
       resolve(true);
--- a/src/db/file/repo.js
+++ b/src/db/file/repo.js
@@ -34,6 +34,22 @@ exports.getRepo = async (name) => {
   });
 };
 
+exports.getRepoByUrl = async (url) => {
+  return new Promise((resolve, reject) => {
+    db.findOne({ url: url }, (err, doc) => {
+      if (err) {
+        reject(err);
+      } else {
+        if (!doc) {
+          resolve(null);
+        } else {
+          resolve(doc);
+        }
+      }
+    });
+  });
+};
+
 exports.createRepo = async (repo) => {
   repo.users = {
     canPush: [],
@@ -140,10 +156,9 @@ exports.deleteRepo = async (name) => {
   });
 };
 
-exports.isUserPushAllowed = async (name, user) => {
-  name = name.toLowerCase();
+exports.isUserPushAllowed = async (url, user) => {
   return new Promise(async (resolve, reject) => {
-    const repo = await exports.getRepo(name);
+    const repo = await exports.getRepoByUrl(url);
     console.log(repo.users.canPush);
     console.log(repo.users.canAuthorise);
 
--- a/src/db/mongo/repo.js
+++ b/src/db/mongo/repo.js
@@ -15,6 +15,11 @@ exports.getRepo = async (name) => {
   return collection.findOne({ name: { $eq: name } });
 };
 
+exports.getRepoByUrl = async (url) => {
+  const collection = await connect(cnName);
+  return collection.findOne({ url: { $eq: url}});
+}
+
 exports.createRepo = async (repo) => {
   console.log(`creating new repo ${JSON.stringify(repo)}`);
 
@@ -67,10 +72,9 @@ exports.deleteRepo = async (name) => {
   await collection.deleteMany({ name: name });
 };
 
-exports.isUserPushAllowed = async (name, user) => {
-  name = name.toLowerCase();
+exports.isUserPushAllowed = async (url, user) => {
   return new Promise(async (resolve, reject) => {
-    const repo = await exports.getRepo(name);
+    const repo = await exports.getRepoByUrl(url);
     console.log(repo.users.canPush);
     console.log(repo.users.canAuthorise);
 
--- a/src/model/Repo.js
+++ b/src/model/Repo.js
@@ -1,6 +1,5 @@
 // Regex inspector - https://www.debuggex.com/
-// eslint-disable-next-line no-useless-escape
-const GIT_URL_REGEX = new RegExp("^(https)://(github\.com|gitlab\.com)/([a-zA-Z0-9\-\.]+)/([a-zA-Z0-9\-]+)(\.git)(/)?$");
+const GIT_URL_REGEX = new RegExp("^(https)://(github\\.com|gitlab\\.com)/([a-zA-Z0-9\\-\\.]+)/([a-zA-Z0-9\\-]+)(\\.git)(/)?$");
 
 /** Class representing a Repo. */
 class Repo {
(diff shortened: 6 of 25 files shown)


JamieSlome:
@msagi - since the wonderful @eddie-knight contributed automated releases, are you able to take a look at the latest status check failure? All PRs will not require a valid label prior to merge. This ensures that we can automate our semantic versioning based on the chosen and agreed titles of improvements, i.e. fix, feature, etc.

msagi:
> @msagi - since the wonderful @eddie-knight contributed automated releases, are you able to take a look at the latest status check failure? All PRs will not require a valid label prior to merge. This ensures that we can automate our semantic versioning based on the chosen and agreed titles of improvements, i.e. fix, feature, etc.

fixed @JamieSlome

msagi:
hi @JamieSlome , would you like me to make any further changes?

config.schema.json
```diff
@@ -70,6 +76,15 @@
     }
   },
   "definitions": {
+    "proxy": {
+      "type": "object",
+      "properties": {
+        "path": { "type": "string"},
+        "url": { "type": "string"},
```
JamieSlome, line 83: Do we have type `url`?

proxy.config.json
```diff
@@ -1,5 +1,16 @@
...
+  "proxyConfig": [
+    {
+      "path": "/github.com",
+      "url": "https://github.com",
+      "enabled": true
+    },
+    {
+      "path": "/gitlab.com",
+      "url": "https://gitlab.com",
+      "enabled": true
```
JamieSlome, line 11: I like your proposed configuration. Would it not be better to recommend `/github` and `/gitlab`? I realise that this is configurable so doesn't really matter and gives control and choice to the developer.

website/docs/quickstart/intercept.mdx
```diff
@@ -58,7 +58,7 @@ By default the clone of your repository will communicate with GitHub. To change
 
 ```bash
 git remote -v
-git remote set-url origin http://localhost:8000/<YOUR-GITHUB-USERNAME>/git-proxy.git
+git remote set-url origin http://localhost:8000/github.com/<YOUR-GITHUB-USERNAME>/git-proxy.git
```
JamieSlome, line 61:
```suggestion
git remote set-url origin http://localhost:8000/github/<YOUR-GITHUB-USERNAME>/git-proxy.git

git remote set-url origin http://localhost:8000/gitlab/<YOUR-GITLAB-USERNAME>/git-proxy.git
```

JamieSlome:
@msagi - getting some breaches in the UI (`admin/push`). We need to think about how this implementation will break existing databases already using previous models. How do we translate between the two and/or support both?

src/db/file/pushes.js
```diff
@@ -94,8 +94,8 @@ const cancel = async (id, logger) => {
 const canUserCancelPush = async (id, user) => {
   return new Promise(async (resolve) => {
     const pushDetail = await getPush(id);
-    const repoName = pushDetail.repoName.replace('.git', '');
-    const isAllowed = await repo.isUserPushAllowed(repoName, user);
+    const repoUrl = pushDetail.repo.url;
```
JamieSlome, line 97: What about pre-existing repositories using the `repoName` property in the database? We need to think about how the update propagates to the previous data model and its impact on already stored properties in DB.

src/db/file/repo.js
```diff
@@ -34,6 +34,22 @@ exports.getRepo = async (name) => {
   });
 };
 
+exports.getRepoByUrl = async (url) => {
```
JamieSlome, line 37: Keep me honest, but can we not enforce uniqueness with `organization`, `name` and `provider`, i.e. `zsh-org`, `zsh` and `gitlab`.

Can we take advantage of the fact that both GitHub and GitLab follow the same naming convention for repositories and enforce uniqueness this way?

GitHub: `finos`, `git-proxy`, `github`
GitLab: `zsh-org`, `zsh`, `gitlab`

src/db/file/repo.js
```diff
@@ -140,10 +156,9 @@ exports.deleteRepo = async (name) => {
   });
 };
 
-exports.isUserPushAllowed = async (name, user) => {
-  name = name.toLowerCase();
+exports.isUserPushAllowed = async (url, user) => {
   return new Promise(async (resolve, reject) => {
-    const repo = await exports.getRepo(name);
+    const repo = await exports.getRepoByUrl(url);
```
JamieSlome, line 161: Again, same point here - this is fine moving forward for updated data properties but what about previous stored data? Do we need a transformer?

src/db/mongo/repo.js
```diff
@@ -15,6 +15,11 @@ exports.getRepo = async (name) => {
   return collection.findOne({ name: { $eq: name } });
 };
 
+exports.getRepoByUrl = async (url) => {
+  const collection = await connect(cnName);
+  return collection.findOne({ url: { $eq: url}});
```
JamieSlome, line 20: `return collection.findOne({ org: ..., repo: ..., platform: ... })`

Thoughts? We should probably step back a little and define a data model for repository together that will serve us well moving forward and isn't too opinionated. I am not sure about URL as only as this is quite un-atomic in its nature and includes content that we don't need in DB, i.e. `https`, `://` - I feel like it can be broken down but happy to hear otherwise of course.

src/model/Repo.js
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * 
+     * @param {string} url The url for the repo.
+     */
+    constructor(url) {
+        const parsedUrl = url?.match(GIT_URL_REGEX);
+        if (parsedUrl) {
+            this.url = url;
+            this.protocol = parsedUrl[1];
```
JamieSlome, line 20: Why is this required?

src/proxy/actions/Action.js
```diff
@@ -35,9 +34,9 @@ class Action {
     this.type = type;
     this.method = method;
     this.timestamp = timestamp;
-    this.project = repo.split('/')[0];
-    this.repoName = repo.split('/')[1];
-    this.url = `${config.getProxyUrl()}/${repo}`;
+    this.project = repo.project;
+    this.repoName = repo.name;
+    this.url = repo.url;
```
JamieSlome, line 39: Nice 👍

config.schema.json
kriswest: nope - but there is a format assertion for URIs
```suggestion
        "url": { "type": "string", "format": "uri"},
```

src/db/file/repo.js
kriswest: While Gitlab and GitHub follow similar patterns, does everyone? What if there's no provider and just a URL (e.g. a project or firm running their own git server)?

The current model does need to change as it appears to just be the repo name and that's not workable alone (two different forks of the same project can't exist nor can two different projects with the same name from different orgs).

src/db/file/repo.js
kriswest: @JamieSlome I find myself looking at the similar changes to Miklos before me. Do you mean something specific by a transformer? I.e. is there something in place for updating the DB models or do we need to invent/adopt/implement something to do so?

src/db/mongo/repo.js
kriswest: As per previous comments, I'm making another attempt at this (and only just discovered this PR addressing the same thing). I had independently run into the same issue when testing with two forks of the same project and realized that changes applied in the UI where being stored on the wrong project as Git Proxy couldn't differentiate them by name alone.

Regarding breaking the URL up into the component date (platform/org/repo), I foresee (an arguably small) problem in going beyond GitHub/GitLab - in any ideal world Git proxy should be applicable to any Git repository. However, that will limit us to concepts in git protocol only and there you only really have the repository URL to work with AFAIK (orgs are contstructs that the platforms apply on top of the repos); you can't break out an org and platform which may not exist.

I'm torn as to whether you do need the protocol details in the DB (i.e. the https://) as there are multiple protocols (ssh + git://) although neither are supported (and are unlikely to be). You could also host the repo with out using SSL - but that would be daft.

Hence, that just leaves simplicity. It is simpler to store and use the whole valid URL as the identity of the project/repo as it avoids the need to process it (and to process it conditionally based on which host/type of host it is). This doesn't affect the end user, only the implementation, where simplicity is a benefit.

kriswest:
@msagi  @JamieSlome we're going to have another crack at this this week, but try to key on the URL (minus protocol). See https://github.com/finos/git-proxy/issues/950#issuecomment-2891244090 for some commentary explaining why.

coopernetes:
Since Miklos has rolled off this project and this PR has been outstanding for awhile, closing for now. Feel free to re-open or cherry pick this onto a new branch without conflicts.

Issue #413, opened 2024-01-24
Implement a CI step to detect unused dependencies 📦 


JamieSlome:
To ensure we don't unnecessarily hit "dependency bloat" nor merge pull requests with dependencies that are unused, we should introduce a GitHub Action or CI step to detect which dependencies are unused. This is a great first issue for someone looking to make their first contribution to the project 🎉 

```[tasklist]
### Tasks
- [x] Detect unused dependencies and fail CI if more than 0
- [x] Detect unused `dependencies` and `devDependencies`
- [x] Only allow PR merges if the Action or CI step returns safely
```

I have previously come across [depcheck](https://www.npmjs.com/package/depcheck) - a nice lightweight NPM library that discovers unused dependencies.

JamieSlome:
@marksmith388 - would you be interested in picking this one up? 🤔

JamieSlome:
@tt-gideonaryeetey and @divinetettey - asked discussed, are you interested in picking this issue up? Let me know below and then I can assign to you if you're happy to 👍

divinetettey:
Hi @JamieSlome   yes I would pick this issue

JamieSlome:
Achieved in #592. You're a rockstar @divinetettey 🚀

PR #592, opened 2024-06-05
chore: added a CI step to detect unused dependencies

divinetettey:
adding this PR to resolve this issue  [Implement a CI step to detect unused dependencies 📦](https://github.com/finos/git-proxy/issues/413)

The changes include: 
1. installing the [depcheck](https://www.npmjs.com/package/depcheck) package 
2. Running `depcheck --oneline --skip-missing`
3.  Fail if the the check fails


.github/workflows/dependency-review.yml
```diff
@@ -18,4 +18,20 @@ jobs:
...
         fail-on-scopes: development, runtime
-        allow-dependencies-licenses: 'pkg:npm/caniuse-lite'
\ No newline at end of file
+        allow-dependencies-licenses: 'pkg:npm/caniuse-lite'
+     - name: 'Setup Node.js'
+       uses: actions/setup-node@v4
+       with:
+        node-version: '18.x'
+     - name: 'Install dependencies'
+       run: npm install
```
JamieSlome, line 27: Do we need this or the `npm install` command if we are invoking `depcheck` via `npx`? 🤔

divinecodes: yeah, I don't think we need. Thanks for catching that. Pushing an update

divinetettey: yh, I don't think we need that. Thanks for catching it. Pushing an update

.github/workflows/dependency-review.yml
```diff
@@ -18,4 +18,18 @@ jobs:
...
         fail-on-scopes: development, runtime
-        allow-dependencies-licenses: 'pkg:npm/caniuse-lite'
\ No newline at end of file
+        allow-dependencies-licenses: 'pkg:npm/caniuse-lite'
+     - name: 'Setup Node.js'
+       uses: actions/setup-node@v4
+       with:
+        node-version: '18.x'
+     - name: 'Install depcheck globally'
+       run: npm install -g depcheck
```
JamieSlome, line 27:
```suggestion
```

JamieSlome:
@divinetettey - is there any chance we can move the check to its own workflow file? What do you think?

divinetettey:
> @divinetettey - is there any chance we can move the check to its own workflow file? What do you think?

I think that's fine, we can go with approach

JamieSlome:
@divinetettey - thanks for making the adjustments (perfect 🎉 ). It seems like the `depcheck` is reporting that `concurrently` and a few other `devDependencies` are unused which is not the case. Is there any way to add these to an allow list or as permissible exceptions?

divinetettey:
> @divinetettey - thanks for making the adjustments (perfect 🎉 ). It seems like the `depcheck` is reporting that `concurrently` and a few other `devDependencies` are unused which is not the case. Is there any way to add these to an allow list or as permissible exceptions?

I have added them to the ignore list

.github/workflows/unused-dependencies.yml
```diff
@@ -0,0 +1,26 @@
+name: 'Unused Dependencies Check'
```
JamieSlome, line 1:
```suggestion
name: 'Unused Dependencies'
```

.github/workflows/unused-dependencies.yml
```diff
@@ -0,0 +1,26 @@
+name: 'Unused Dependencies'
+on: [pull_request]
+
+permissions: 
+  contents: read 
+  pull-requests: write
```
JamieSlome, line 6:
```suggestion
```
Are `write` permissions actually required here?

JamieSlome:
@maoo - can we make unused dependencies required on status checks to `main` please?

JamieSlome (approved):
LGTM! 🍰

maoo:
> @maoo - can we make unused dependencies required on status checks to `main` please?

Done!

Issue #557, opened 2024-05-10
Automatically increment version in `package.json` on new release 📦 


JamieSlome:
Due to the fabulous @eddie-knight, GitProxy now has auto releases 💪  As part of #532, generally, all pull requests merged into `main` will result in a new release to GitHub Releases and NPM.

To ensure consistency across the codebase, GitHub Releases and NPM versions, we should ensure that the root `package.json` reflects the latest auto release version.

```[tasklist]
### Tasks
- [ ] Automatically update the `package.json` to reflect the latest auto generated semver number
- [ ] For bonus points ⚡ update the `docusaurus.config.js` `version` with the latest auto generated semver number
```

JamieSlome:
@eddie-knight - interested in taking this one and polishing off your Mona Lisa of contributions?

eddie-knight:
Hey @JamieSlome— I was looking into this and it may need a pivot from what I put up based on my previous experience with golang and python, where we didn't need to store the version number in the codebase. There are a few things that turned up in a cursory search that may be viable without big changes to what I've already got rolling, but I think it'll be better to take a look at it with an eye that's more familiar with node/npm than me.

eddie-knight:
@JamieSlome spitballing here... what about adding something like this?

```
      - run: npm ci
      - name: Update package.json version
        id: update_version
        run: |
          version="${{ steps.release-drafter.outputs.tag_name }}"
          jq --arg version "$version" '.version = $version' package.json > package_tmp.json
          mv package_tmp.json package.json
      - name: Commit version change
        run: |
          git config --global user.name 'github-actions'
          git config --global user.email 'github-actions@github.com'
          git add package.json
          git commit -m "chore: bump version to ${{ steps.release-drafter.outputs.tag_name }}"
          git push
```

eddie-knight:
@JamieSlome spitballing here... what about adding something like this?

```
      - run: npm ci
      - name: Update package.json version
        id: update_version
        run: |
          version="${{ steps.release-drafter.outputs.tag_name }}"
          jq --arg version "$version" '.version = $version' package.json > package_tmp.json
          mv package_tmp.json package.json
```

Note that the version number in the git repo will no longer be managed.

[...]

PR #596, opened 2024-06-07
chore: automatically increment version in package.json on new release

divinetettey:
This PR attempts to resolve this issue  [Automatically increment version in package.json on new release 📦](https://github.com/finos/git-proxy/issues/557) .  The PR aims at automating the  update off the version in pacakge.json 

1. A new step has been added to extract the version number from the draft release. this version number is then used in subsequent steps to update the package.json
2. The extracted version number is used to update the package.json,package-lock.json and the docusaurus.config.js
3. after update the update a commit is pushed back to the repository

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,60 +1,91 @@
-# ---
-# name: Release
+---
+name: Release
 
-# on:
-#   workflow_dispatch:
-#   push:
-#     branches:
-#       - main
-#     paths:
-#       - 'src/**'
-#       - 'test/**'
-#       - 'scripts/**'
-#       - 'public/**'
-#       - 'packages/**'
-#       - 'package.json'
-#       - 'package-lock.json'
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - 'src/**'
+      - 'test/**'
+      - 'scripts/**'
+      - 'public/**'
+      - 'packages/**'
+      - 'package.json'
+      - 'package-lock.json'
 
-# permissions:
-#   contents: read
+permissions:
+  contents: read
 
-# jobs:
-#   create_github_release:
-#     outputs:
-#       full-tag: ${{ steps.release-drafter.outputs.tag_name }}
-#       short-tag: ${{ steps.get_tag_name.outputs.SHORT_TAG }}
-#       body: ${{ steps.release-drafter.outputs.body }}
-#     runs-on: ubuntu-latest
-#     permissions:
-#       contents: write
-#       pull-requests: read
-#     steps:
-#       - uses: release-drafter/release-drafter@v6
-#         id: release-drafter
-#         env:
-#           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-#         with:
-#           publish: true
-#       - name: Get the short tag
-#         id: get_tag_name
-#         run: |
-#           short_tag=$(echo ${{ steps.release-drafter.outputs.tag_name }} | cut -d. -f1)
-#           echo "SHORT_TAG=$short_tag" >> $GITHUB_OUTPUT
-#   create_npm_release:
-#     needs: create_github_release
-#     runs-on: ubuntu-latest
-#     permissions:
-#       packages: write
-#     env:
-#       REGISTRY: ghcr.io
-#       IMAGE_NAME: ${{ github.repository }}
-#     steps:
-#       - uses: actions/checkout@8459bc0 # v4
-#       - uses: actions/setup-node@c2ac33f # v4, Setup .npmrc file to publish to npm
-#         with:
-#           node-version: '18.x'
-#           registry-url: 'https://registry.npmjs.org'
-#       - run: npm ci
-#       - run: npm publish --access=public
-#         env:
-#           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+jobs:
+  create_github_release:
+    outputs:
+      full-tag: ${{ steps.release-drafter.outputs.tag_name }}
+      short-tag: ${{ steps.get_tag_name.outputs.SHORT_TAG }}
+      body: ${{ steps.release-drafter.outputs.body }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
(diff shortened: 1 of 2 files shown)


JamieSlome:
Thanks for the PR, @divinetettey 👍 Is there anyway we can update the `package.json` and `package-lock.json` version before we run the release? i.e. include the latest `version` number in the actual release itself

divinetettey:
> Thanks for the PR, @divinetettey 👍 Is there anyway we can update the `package.json` and `package-lock.json` version before we run the release? i.e. include the latest `version` number in the actual release itself

sure, let me take a look at this

divinetettey:
> > Thanks for the PR, @divinetettey 👍 Is there anyway we can update the `package.json` and `package-lock.json` version before we run the release? i.e. include the latest `version` number in the actual release itself
> 
> sure, let me take a look at this

should this run before/after the release-drafter

JamieSlome:
@divinetettey - we want the file bumps to occur before we make the NPM and GitHub releases. This is to ensure that the releases themselves include the correct `version` number.

JamieSlome:
@divinetettey, the @divinecodes will also need to authorise with EasyCLA to merge this pull request (heads up).

divinetettey:
> @divinetettey, the @divinecodes will also need to authorise with EasyCLA to merge this pull request (heads up).

yeah, my gitconfigs got messed and the commit was pushed with my personal account. I am going to squash the commit and push it with @divinetettey

.github/workflows/release.yml
```diff
@@ -1,60 +1,110 @@
...
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git add package.json package-lock.json website/docusaurus.config.js
+          git commit -m "chore: update version to ${{ steps.extract_version.outputs.version }}"
+  
+      - name: Push Changes
+        uses: ad-m/github-push-action@master
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
JamieSlome, line 70: @divinetettey - we are super close now. These changes are perfect. Can we make sure that this commit and push won't re-trigger another release? Every commit to `main`, according to this workflow will result in a release. That is spot on, apart from we don't want this commit to trigger another re-deploy.

divinetettey: what about adding a phrase to the commit message of the version bump  eg. chore(skip ci) and skip the workflow if the commit message contains the phrase ?

JamieSlome: Yes, perfect! GitHub Actions detects the usage of [[skip ci]](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs#:~:text=a%20pull%20request%3A-,%5Bskip%20ci%5D,-%5Bci%20skip%5D).

.github/workflows/release.yml
```diff
@@ -1,60 +1,110 @@
...
+            echo "Version is empty. Exiting"
+            exit 1
+          fi
+          sed -i "s/version: '.*'/version: '$VERSION'/" website/docusaurus.config.js
+      - name: Commit version change
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git add package.json package-lock.json website/docusaurus.config.js
+          git commit -m "chore: update version to ${{ steps.extract_version.outputs.version }}"
```
JamieSlome, line 65:
```suggestion
          git commit -m "chore: update version to ${{ steps.extract_version.outputs.version }} [skip-ci]"
```

JamieSlome (commented):
Happy from my end. @divinetettey - have you been able to test the `update_version` step?

divinetettey:
> Happy from my end. @divinetettey - have you been able to test the `update_version` step?

Yes, getting some failures with release-drafter, taking a look

divinetettey:
> Happy from my end. @divinetettey - have you been able to test the `update_version` step?

all good on my end

.github/workflows/release.yml
```diff
@@ -1,60 +1,111 @@
...
+          echo "version=$version" >> $GITHUB_OUTPUT
+      - name: Update package.json and package-lock.json version
+        run: 
+          VERSION=${{ steps.extract_version.outputs.version }}
+          if [ -z "$version"]; then
+            echo "Version is empty. Exiting"
+            exit 1
+          fi
+          jq --arg version "$VERSION" '.version = $version' package.json > package_tmp.json && mv package_tmp.json package.json
+          jq --arg version "$VERSION" '.version = $version' package-lock.json > package_lock_tmp.json && mv package_lock_tmp.json package-lock.json
```
JamieSlome, line 52: Instead of this step, can we update the `package.json` only and then run `npm install`? This will automatically update the `package-lock.json` for us.

divinetettey: sure would make this change

JamieSlome: Awesome, thank you @divinetettey 👍

.github/workflows/release.yml
```diff
@@ -1,60 +1,112 @@
...
+          VERSION=${{ steps.extract_version.outputs.version }}
+          if [ -z "$version"]; then
+            echo "Version is empty. Exiting"
+            exit 1
+          fi
+          sed -i "s/version: '.*'/version: '$VERSION'/" website/docusaurus.config.js
+      - name: Commit version change
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
```
Parvezkhan0, line 65: The echo` "$version" || sed 's/^v//`' statement is incorrect because `echo `will always succeed, meaning the `sed `command after `|| `will never execute. Instead, you should be using `| `(pipe) instead of` ||`.

Parvezkhan0 (changes requested):

.github/workflows/release.yml
```diff
@@ -1,60 +1,112 @@
...
+          git add package.json package-lock.json website/docusaurus.config.js
+          git commit -m "chore: update version to ${{ steps.extract_version.outputs.version }} [skip ci]"
```
jescalada, line 67: Since this action triggers when pushing to main, committing and pushing in the action will probably cause an infinite loop.

Issue #425, opened 2024-01-26
Git Proxy architecture (2.x?)


coopernetes:
Tracking issue to revisit Git Proxy's internal architecture and figuring out how best to design the application going forward. I'm including my initial sketch and using this issue to bounce around ideas, flesh out the right interfaces & API boundaries as well as discuss overall what we want to change in Git Proxy internals.

Issues that are blocked on the current architecture or would greatly benefit from a new one:
- #27
- #61
- #62 
- #66 
- #10

Related issues:
- #276
- #99  
- #414

coopernetes:
Here's first stab at a new architecture. Some initial thoughts:

- the idea of a "chain" of actions that Git Proxy executes in some sequence is extremely valuable. I don't think we need to rework this that much as we currently have a series of Step/Action types that the application cycles through for every git push.
- @JamieSlome brought up some interesting points in #99 around policy management. What constitutes a "policy" vs a "filter" is not immediately clear to me but some separation between a decision to permit and the underlying git request seems like a useful abstraction. It would also allow us to implement an interface that is generic to other policy engines like OPA
- Each phase of the request lifecycle should be able to fire off events in an agnostic way. How we persist those events can be abstracted into a persistence interface and backed by a variety of sources (object storage, SQL, Mongo, etc).
- I included an idea of using git-proxy purely in a localized sense such as a library running on a development machine. This may be preferable to a hosted server so I've separated those out into their own components. At the very least, this allows a developer to directly use git-proxy without the new to run it as a server.

![git-proxy-Architecture drawio-2](https://github.com/finos/git-proxy/assets/57812123/185e263a-76f3-4f46-877e-5fb85e70e2ad)

Much of this will likely result in breaking changes so I consider this a version 2.x of git-proxy so we're free to iterate on the new design.

maoo:
Really like this new architecture, and the drawing.

For the latter, I think it would help to draw where users connect to, specifically:
- a Developer (connecting directly to git-proxy-server, or the local block close to it)
- a Compliance manager, which should connect to git-proxy-ui, though I don't see it mentioned ; I also assume that the UI connects to an API layer that is exposed by the core package, which would be helpful to draw.

WDYT?

JamieSlome:
I'd like to bring in #428 to `2.x` as well. Likely sits in the filter chain in the architecture diagram above.

@coopernetes - thoughts?

coopernetes:
[...]

PR #638, opened 2024-07-06
feat: use native node loading for plugins + docs

### LINKED ISSUE #539: Creating your first plugin documentation
I would really like some more detailed configuration. Especially as it comes to "plugins" (configuring them, creating them, etc.).

With the current documentation I'm not sure if the project is useful for my intended purpose: We want to automate some Git processes (e.g. tagging commits) but since those run in an automated fashion (on our build agents) we want to limit what these automated processes can do (e.g. only push tags).

Is that possible with Git Proxy? If so, could you provide some example configuration how?

coopernetes:
fixes #539 

- update PluginLoader to use load-plugin's native Node module resolution. this library already handles loading of files vs modules in node_modules/ so let's use that directly
- add configuration-based plugin loading and deprecate environment variables
- new documentation page for plugin development. loading, writing, etc.

TODO:
- [ ] Add test cases for PluginLoader
- [ ] Decide whether plugin development docs should be on the doc site or fine to leave it in the repo. @JamieSlome or @maoo lmk your thoughts on the best place for this (both?)


JamieSlome:
@coopernetes - I'd say in both the `README.md` and docs site if possible. In the `README.md`, if we can compress into a few lines of how to create your first plugin, that would be fabulous 👍 

In the docs site, we can go through more details for each step, the granular details etc. Thoughts?

JamieSlome:
Just had a fairly in-depth review of the pull request. Does the push plugins defined just add the plugins directly to the end of the array of the chain in order that they are configured in `pushPlugins`? Also, why `pushPlugins`? Should we not just expose `plugins` and include the config of `push` or `pull` at a lower level?

maoo:
Great work @coopernetes, somehow you managed to wrap all of this with very few changes!

Re. the docs, I'd suggest to add the plugins/README.md as part of the website, then link the URL into the main README.md , to avoid defining the same docs in 2 different places.

I also agree with @JamieSlome re. `pushPlugins` VS `plugins`; the operation that gets implemented (in this case `push`) could be defined by the function signature that gets implemented (or how they get exported?)

I'll try to allocate some time this week to give this a go.

maoo:
@coopernetes - I couldn't get it to work; I did some debugging, and it seems to me that the following block in `plugin.js` does not get executed:
```
    Promise.all(modulePromises).then((vals) => {
      const modules = vals;
      console.log(`Found ${modules.length} plugin modules`); // TODO: log.debug()
      const pluginObjPromises = [];
      for (const mod of modules) {
        pluginObjPromises.push(this._castToPluginObjects(mod));
      }
      Promise.all(pluginObjPromises).then((vals) => {
        for (const pluginObjs of vals) {
          this.plugins = this.plugins.concat(pluginObjs);
        }
        console.log(`Loaded ${this.plugins.length} plugins`); // TODO: log.debug()
      });
    });
```

even though the line `modulePromises.push(this._loadPlugin(target));` gets correctly executed.

This is what I did to reproduce the issue:

- Followed https://github.com/finos/git-proxy/pull/649/files#diff-9eca5a1e442f863695367206f19cbb2b6c0aab0da2211ba0f6b9194c2fcfdb12R33 for local checkout
- Updated proxy.config.json with a new entry in `pushPlugins` with `"./plugins/git-proxy-hello-world/index.js"` and a new entry in `authorizedList`
```
{
  "project": "maoo",
  "name": "maven-tiles",
  "url": "https://github.com/maoo/maven-tiles.git"
}
```
- Run `git-proxy --config ./proxy.config.json`
- Open a new terminal
- Checkout repo
```
cd /tmp
git clone http://localhost:8000/maoo/maven-tiles.git
cd maven-tiles
```
- Edit `README.md`
- `git add README.md ; git commit -m 'test: small change' ; git push origin master`
- commented out `proc.push.checkUserPushPermission`and `proc.push.blockForAuth` in `chain.js`, to make testing easier; I wonder if this list should be added as entries in `proxy.config.js` under `pushPlugins`.

Could it be something related with Promises?

coopernetes:
> Does the push plugins defined just add the plugins directly to the end of the array of the chain in order that they are configured in `pushPlugins`? 

Custom plugins are added after the parseAction but before any builtin actions that are part of the default chain. 

See https://github.com/coopernetes/git-proxy/blob/chore/plugin-docs/src/proxy/chain.js#L56-L59

> Also, why `pushPlugins`? Should we not just expose `plugins` and specify the type of plugin from inside the plugin itself?

We can have a generic set of "plugins" via configuration and attempt to do some funky casting....Note that this will become much cleaner with 2.0 + TypeScript (#425) since we can enforce stronger API contracts through TS. It's just not useful today with the JS code base. The thought was to be explicit about the plugin types so that when we load 3rd party modules, we will presume that plugin authors wrote the specific object (class) based on its functionality. In reality, a push and pull are virtually identical so we can update this in the new configuration & docs.

coopernetes:
@maoo I had this working when running from source. I'm not sure how the global install affects the plugin loading but I suspect it has to do with the "path" of git-proxy itself when it's sitting in your user's global npm cache in relation to the plugins. The `load-plugin` library relies on Node's builtin module resolution which looks at a few known locations. 

I'll do some debugging on my side with a similar setup.

coopernetes:
Closing this to re-open on a separate fork. Our org is iterating on this a bit and testing internally first before it's suitable to merge.

PR #648, opened 2024-07-17
docs: adding Zoom meeting and mailing list to README

maoo:
@JamieSlome - should we also add something on git-proxy.finos.org ?


JamieSlome (approved):
LGTM! 🍰

PR #720, opened 2024-09-30
chore: move from bcrypt to bcryptjs

06kellyjac:
Removes dependency on node-gyp built extensions. Pure JS.
There could be slight performance penalty but makes the project more portable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "@primer/octicons-react": "^19.8.0",
     "@seald-io/nedb": "^4.0.2",
     "axios": "^1.6.0",
-    "bcrypt": "^5.1.1",
+    "bcryptjs": "^2.4.3",
     "bit-mask": "^1.0.2",
     "body-parser": "^1.20.1",
     "chai-http": "^4.3.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "@primer/octicons-react": "^19.8.0",
         "@seald-io/nedb": "^4.0.2",
         "axios": "^1.6.0",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
     "@primer/octicons-react": "^19.8.0",
     "@seald-io/nedb": "^4.0.2",
     "axios": "^1.6.0",
-    "bcrypt": "^5.1.1",
+    "bcryptjs": "^2.4.3",
```
rgmz, line 37: According to npm, [`bcryptjs`](https://www.npmjs.com/package/bcryptjs?activeTab=versions) has a single collaborator and last publish a version 8 years ago, whereas [`bcrypt`](https://www.npmjs.com/package/bcrypt?activeTab=versions) has 5 collaborators and last published a version 1 year ago. Is maintenance a concern with this dependency?

JamieSlome: Thanks for the question @rgmz and for the clarification @06kellyjac. I'm happy to proceed with a merge, but will propose that we migrate our authentication services to a plugin-based model. For our usage of GitProxy, we do not require the `file` based password service (using `bcrypt`) however are trumped by installation issues associated with a child dependency and running of `node-gyp`.

JamieSlome:
@rgmz - left some comments. Let us know if you have any further thoughts, otherwise will proceed with a merge (for the time being) and open an issue to build an auth plugin.

rgmz:
I'm happy if you're happy. Just figured it was worth highlighting.

JamieSlome:
@rgmz - definitely worth highlighting and thank you as always for your contribution ❤️

PR #735, opened 2024-10-02
fix: #669-ui/docs: clone component is undefined when running as a server

### LINKED ISSUE #669: ui/docs: clone component is undefined when running as a server
**Describe the bug**
When Git Proxy is deployed on a server, the "Clone" component under /admin/repo on the UI has an `undefined` value for the host.
It seems to stem from an unset environment variable `VITE_SERVER_URI`.

Ideally, this component would use the Origin URL and not require additional config when deploying Git Proxy's UI.

**To Reproduce**
1. Deploy Git Proxy behind a load balancer and running on a server.
2. Navigate to http[s]://<host>:8080/admin/repo
3. Click on the "Code" component on any of the listed repos.

**Expected behavior**
The frontend uses the server origin as the hostname for the clone URL.

Screenshot:
image

RajvirS99:
closes #669 

Instead of getting the origin from the vite environment variables, the component will now get the origin from browsers 
JavaScript location API and It won't show undefined in the server

### DIFF
--- a/src/ui/views/RepoList/Components/RepoOverview.jsx
+++ b/src/ui/views/RepoList/Components/RepoOverview.jsx
@@ -585,7 +585,7 @@ export default function Repositories(props) {
   };
 
   const { project: org, name } = props?.data || {};
-  const cloneURL = `${import.meta.env.VITE_SERVER_URI}/${org}/${name}.git`;
+  const cloneURL = `${window.location.origin.toString()}/${org}/${name}.git`;
 
   return (
     <TableRow>


src/ui/views/RepoList/Components/RepoOverview.jsx
```diff
@@ -585,7 +585,7 @@ export default function Repositories(props) {
   };
 
   const { project: org, name } = props?.data || {};
-  const cloneURL = `${import.meta.env.VITE_SERVER_URI}/${org}/${name}.git`;
+  const cloneURL = `${window.location.origin.toString()}/${org}/${name}.git`;
```
JamieSlome, line 588: Is it worth doing import.meta.env.VITE_SERVER_URI `OR` window.location.origin.toString()?

JamieSlome: Ignore me 👍

PR #738, opened 2024-10-06
refactor: replace react-html-parser with html-react-parser

lwhiteley:
After an audit of the dependencies of this project, it was determined that react and a few libraries were outdated.

one of the dependencies preventing an upgrade is `react-html-parser` . it has a peer dependency of of react 16.
A successor of the package is `html-react-parser`

However, there is no storybook setup in the project or a way to test this easily.

How can i get diff data passed to this component to test in isolation or 
Can someone gives steps to arrive at this view to test it in the UI to verify it works?

I also see that there are cypress tests but i dont believe this is covered

old package: https://www.npmjs.com/package/react-html-parser
proposed package: https://www.npmjs.com/package/html-react-parser

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,6 +32,7 @@
         "express-rate-limit": "^7.1.5",
         "express-session": "^1.17.1",
         "history": "5.3.0",
+        "html-react-parser": "^5.1.17",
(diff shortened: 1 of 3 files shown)


lwhiteley:
close in favour of #740

PR #742, opened 2024-10-07
fix: allow all default config to be replaced by user config

06kellyjac:
Currently most config options can be replaced, this change adjusts the rest
to also allow for this.


JamieSlome (approved):
LGTM! 🍰

PR #755, opened 2024-10-23
chore: add cookie secret env var

### LINKED ISSUE #690: making COOKIE_SECRET an env var
make `COOKIE_SECRET` an env var instead of checking in `proxy.config.json`

https://github.com/finos/git-proxy/blob/fd638483c70bc84f3f9cedb6e10f8d0a18bafacf/proxy.config.json#L3

Aditya-PS-05:
closes #690


src/config/index.js
```diff
@@ -100,8 +100,8 @@ const getAPIs = () => {
 };
 
 const getCookieSecret = () => {
-  if (_userSettings && _userSettings.cookieSecret) {
-    _cookieSecret = _userSettings.cookieSecret;
+  if (!_cookieSecret) {
+    throw new Error('COOKIE_SECRET is not defined in the environment variables.');
```
rgmz, line 104: Wouldn't this never trigger because of the fallback to `'cookie secret'`?

src/plugin.js
rgmz, line 1: Are there any non-whitespace changes in this file?

src/config/index.js
Aditya-PS-05: Yeah sorry about that, I need to correct it.

Aditya-PS-05:
@rgmz 
Please review the pr as soon as possible please.

JamieSlome:
@Aditya-PS-05 - it looks like you have a prettier or linter that doesn't align with the standard of the project? Are you able to revert non-related changes so the PR includes changes only relevant to your suggested improvement? Massively appreciate your PR and effort ❤️

Issue #745, opened 2024-10-07
Support filter by file with file pattern and literal configuration


JamieSlome:
### Background

GitProxy currently supports commit message, commit diff and commit e-mail domain scanning. A common use case that is not supported is file name filtering and scanning.

### Vision 🪄 

```json
"message": {
  "block": {
    "literals": ["xlsx", "csv"],
    "patterns": [".*.txt"]
  }
},
```
```[tasklist]
### Tasks
```

laukik-target:
Hi @JamieSlome 
Ready to take this up!

JamieSlome:
@laukik-target - all yours ❤️

laukik-target:
Looks like [Psingle20](https://github.com/Psingle20) PR is ready before mine

@JamieSlome  You can assign it to [Psingle20](https://github.com/Psingle20) and team

PR #761, opened 2024-10-26
Feat: Sensitive Data Detection in files like (.csv , .xlsx , json) 

Psingle20:
This PR introduces the checkSensitiveData feature, which enhances the security by scanning files like .csv for vulnerabilities and sensitive information.
 The implementation includes:

Functionality:
- Created a push-action CheckSensitiveData which take this diff and scan the changed files for Sensitive Information.
- Integrated it push_action chain .
- Implemented a Test file for the push-action and modified the chain test file to make sure it works with feature added.

I think this Functionality Solves the issue #745
you can run the custom test implemented using command `npx mocha test/SensitiveData.test.js`
@JamieSlome  Please review this PR and suggest any changes necessary

Citi Hackathon
Team Members
Prachit Ingle [Psingle20](https://github.com/Psingle20)
Shabbir Kaderi  [shabbirflow](https://github.com/shabbirflow)
Chaitanya Deshmukh [ChaitanyaD48](https://github.com/ChaitanyaD48)


.husky/commit-msg
laukik-target, line 1: Any specific reason, why this has been done in this PR?

test/CheckSensitive.test.js
laukik-target, line 1: The current test suite looks solid in terms of validating the core functionality, but adding additional tests (like edge cases: Empty File Test and multi-file handling) will increase the coverage and robustness of the tests.

laukik-target (commented):
The code is well-structured, handles different file types effectively.
few improvements are recommended:

- Test Coverage: Add tests for no sensitive data, empty files, and file-not-found scenarios.
- Optimization: Consider streaming large files for better memory management.

.husky/commit-msg
Psingle20: In the latest versions of Husky, the` _husky.sh` helper script has been removed because it’s no longer needed. Previously, this helper script was used in each Git hook file to set up the environment, allowing Husky to run specified commands.

With the updated Husky versions, the Git hook files now run the commands directly without relying on `_husky.sh` or other setup scripts.

test/CheckSensitive.test.js
Psingle20: Yeah actually I had that in back of my mind . I just wanted to make sure I was on that right track and the core functionality itself was working. 
I will get to it as soon as possible.

src/proxy/processors/push-action/checkSensitiveData.js
```diff
@@ -0,0 +1,136 @@
...
+// Async exec function to handle actions
+const exec = async (req, action) => {
+    // getDiffExec(req, action); // Call to getDiffExec if necessary
+    const diffStep = action.steps.find((s) => s.stepName === 'diff');
+    if (diffStep && diffStep.content) {
+        console.log('Diff content:', diffStep.content);
+        const filePaths = diffStep.content.filePaths || [];
+        if (filePaths.length > 0) {
+            // Check for sensitive data in all files
+            const sensitiveDataFound = await Promise.all(filePaths.map(parseFile));
```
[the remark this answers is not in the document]
Psingle20: shouldn't we prioritize concurrent execution over resource allocation as if the no of files increases it would slow down the speed if we go with for .. of with await approach although I agree in case large datasets it would consume a lot of memory.

src/proxy/processors/push-action/checkSensitiveData.js
```diff
@@ -0,0 +1,161 @@
...
+const Step = require('../../actions').Step;
+// const { exec: getDiffExec } = require('./getDiff');
+// Function to check for sensitive data patterns
+const checkForSensitiveData = (cell) => {
+    const sensitivePatterns = [
+        /\d{3}-\d{2}-\d{4}/, // Social Security Number (SSN)
+        /\b\d{16}\b/, // Credit card numbers
+        /\b\d{5}-\d{4}\b/, // ZIP+4 codes
+        // Add more patterns as needed
+    ];
```
rgmz, line 14: The intent behind this change is good, though it must be noted these will produce a large number of false positives.

Ideally this wouldn't block, or would have an easy way to exclude false positives.

src/proxy/processors/push-action/getDiff.js
```diff
@@ -2,6 +2,7 @@ const Step = require('../../actions').Step;
 const simpleGit = require('simple-git')
 
 
+
```
coopernetes, line 5: Mistaken change?
```suggestion
```

src/proxy/processors/push-action/checkSensitiveData.js
```diff
@@ -0,0 +1,171 @@
...
+        console.log('No diff content available.');
+    }
+    action.addStep(step);
+    return action; // Returning action for testing purposes
+};
+
+
+
+exec.displayName = 'logFileChanges.exec';
+exports.exec = exec;
```
coopernetes, line 171:
```suggestion
exports.exec = exec;

```

src/proxy/processors/push-action/index.js
```diff
@@ -11,3 +11,4 @@ exports.checkCommitMessages = require('./checkCommitMessages').exec;
 exports.checkAuthorEmails = require('./checkAuthorEmails').exec;
 exports.checkUserPushPermission = require('./checkUserPushPermission').exec;
 exports.clearBareClone = require('./clearBareClone').exec;
+exports.checkSensitiveData = require('./checkSensitiveData').exec;
```
coopernetes, line 14: Add missing newline at the end of the file.
```suggestion
exports.checkSensitiveData = require('./checkSensitiveData').exec;

```

src/proxy/processors/push-action/checkSensitiveData.js
```diff
@@ -0,0 +1,171 @@
...
+            }
+        } else {
+            console.log('No file paths provided in the diff step.');
+        }
+    } else {
+        console.log('No diff content available.');
+    }
+    action.addStep(step);
+    return action; // Returning action for testing purposes
+};
```
coopernetes, line 166: @Psingle20 since #793 has been released in 1.6.0, can we restructure this functionality into its own plugin? It'll require moving some files around and creating an npm package using `npm init`.

The other change will be this:
```
const Step = require('@finos/git-proxy/src/proxy/actions').Step;
const config = require('@finos/git-proxy/src/config');
```

Use [plugins/git-proxy-sample-plugins](https://git-proxy.finos.org/docs/development/plugins) and refer to the docs (to be improved via #811) for details.

test/CreateExcel.js
```diff
@@ -0,0 +1,19 @@
...
+const workbook = XLSX.utils.book_new();
+XLSX.utils.book_append_sheet(workbook, worksheet, "SensitiveData");
+// Create the path to the test_data directory
+const testDataPath = path.join(__dirname, 'test_data'); // Ensure this points to the correct directory
+// Create the test_data directory if it doesn't exist
+if (!fs.existsSync(testDataPath)){
+    fs.mkdirSync(testDataPath, { recursive: true }); // Using recursive to ensure all directories are created
+}
+// Write the Excel file to the test_data directory
+XLSX.writeFile(workbook, path.join(testDataPath, 'sensitive_data2.xlsx'));
```
coopernetes, line 19: Add missing newline at the end of the file.
```suggestion
XLSX.writeFile(workbook, path.join(testDataPath, 'sensitive_data2.xlsx'));

```

test/chain.test.js
```diff
@@ -25,6 +26,7 @@ const mockPushProcessors = {
   pullRemote: sinon.stub(),
   writePack: sinon.stub(),
   getDiff: sinon.stub(),
+  checkSensitiveData : sinon.stub(),
```
coopernetes, line 29: By moving this new functionality in its own plugin, it will keep the proxy chain easier to test. Having to constantly add new functions here will not be maintainable long term. This test itself is also not exactly well structured or easy to maintain so we want to not add to it as much as possible. Plugins are preferred.

proxy.config.json
```diff
@@ -77,7 +77,8 @@
       "block": {
         "literals": [],
         "patterns": [],
-        "providers": {}
+        "providers": {},
+        "proxyFileTypes":[]
```
coopernetes, line 81: Suggested name change to keep the configuration property consistent with the rest of the functions (checkSensitiveData, etc).

```suggestion
        "sensitiveFileTypes":[]  
```

.husky/commit-msg
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
+
```
coopernetes, line 1: Kindly revert this please. Not sure why it was removed.

```suggestion
#!/usr/bin/env sh
. "$(dirname -- "$0")/_/husky.sh"
```

.gitignore
```diff
@@ -11,6 +11,7 @@ yarn-debug.log*
 yarn-error.log*
 lerna-debug.log*
 package-lock.json
+/git-proxy-test
```
coopernetes, line 14: This seems unnecessary.
```suggestion
```

src/proxy/processors/push-action/checkSensitiveData.js
coopernetes: I agree. Not to mention, this does not cover all [geographies](https://www.canada.ca/en/revenue-agency/services/tax/individuals/topics/about-your-tax-return/tax-return/completing-a-tax-return/personal-address-information/social-insurance-number.html).

I'm inclined to merge it as it is not configured by default. A more holistic approach with better heuristics is worth investing in for the GitProxy project granted but this is a good enough start.

coopernetes (changes requested):

src/proxy/processors/push-action/checkSensitiveData.js
Psingle20: I will do that. Btw we have added some refactor in #810  which also contains some more feature should I do a PR individually for all the plugins ?
ALong with this we have added gitleaks support , EXIF metadata check and AIML usage check.

jescalada (commented):
@Psingle20 Thanks for the contribution!

These changes seem more appropriate for a plugin rather than a push action, as many organizations don't need to scan those specific file types for those specific patterns. The `commitConfig.diff` config property along with `scanDiff` already allows scanning `.csv` and `.json` files which are stored in plaintext.

It'd be great to adapt the Excel checker into its own plugin, so it can be used by users that really need it instead of slowing down the default push chain.

I'll close this PR to keep our backlog clean. Feel free to open a new PR with the requested changes 🙂

Issue #673, opened 2024-08-02
pullRemote clone dir permissions are extraneous 


coopernetes:
When Git Proxy is running in a secure environment, the node process is likely running under some sort of user. This user should have local file permissions to create directories for the file-based DB (if used) as well as cloning repositories that are in-flight via Git Proxy. The pullRemote step expects to create a directory under `.remote/{timestamp}`. Currently, that directory is expected to have full rwx permissions across user, group and everyone/world. 

This is generally not best practice. In certain deployment environments such as Openshift, it's explicitly disallowed.

A simple fix is to create a directory with only user permissions which is typically rwxr-xr-x (0755) for directories.

See: https://betterstack.com/community/questions/what-permissions-should-my-website-directory-have-on-linux/

Affected line: https://github.com/finos/git-proxy/blob/1f39f10e3ec4515a3684c3bf881c5edfa435554e/src/proxy/processors/push-action/pullRemote.js#L19

JamieSlome:
@coopernetes - can we close this with the merge of https://github.com/finos/git-proxy/pull/782?

PR #782, opened 2024-11-06
fix: Update pullRemote.js to set appropriate directory permissions

yadnyeshkolte:
This PR modifies the directory creation permissions in the pullRemote function of pullRemote.js. 

Previously, the function was setting directory permissions to 0777 (full read, write, and execute permissions for user, group, and others). This approach is not aligned with best practices for security, particularly in secure environments such as OpenShift, where overly permissive settings can lead to vulnerabilities.

The updated code now sets the permissions to 0755 (read, write, and execute for the user; read and execute for group and others). This change enhances security by restricting write access to the owner only while still allowing necessary read and execute permissions.

### DIFF
--- a/src/proxy/processors/push-action/pullRemote.js
+++ b/src/proxy/processors/push-action/pullRemote.js
@@ -16,7 +16,7 @@ const exec = async (req, action) => {
     }
 
     if (!fs.existsSync(action.proxyGitPath)) {
-      fs.mkdirSync(action.proxyGitPath, '0777', true);
+      fs.mkdirSync(action.proxyGitPath, '0755', true);
     }
 
     const cmd = `git clone ${action.url} --bare`;


yadnyeshkolte:
@JamieSlome  could you please review this PR.

06kellyjac (commented):
Firstly I'm always anti `777` so I'm happy to see this change. :tada: 

If we're already changing this then `700` might be more appropriate. IDK if we need any other groups or public looking at these files.

---

Some extra nuance:
Regarding deployment in a container on a platform such as OpenShift this shouldn't have been a particularly dangerous permission level as your container is typically single purpose, single user.
If someone is going to break into the container they'd probably do it via the service itself making them the `7` in `755` anyway.

If someone was able to get into the container not as root or the user running git-proxy, but as some other user, then this would help.

This is also a much larger problem if git-proxy is running alongside many other processes/users so good to fix either way.

JamieSlome (approved):
LGTM! 🍰 

Thanks for the comments @06kellyjac 👍

coopernetes:
Thank you @yadnyeshkolte for the contribution. In the future, please reference any related issue(s) that your pull request is intended to close. In this case, I've edited the description to reference #673

See [GitHub docs](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls#issues-and-pull-requests) for details.

yadnyeshkolte:
> Thank you @yadnyeshkolte for the contribution. In the future, please reference any related issue(s) that your pull request is intended to close. In this case, I've edited the description to reference #673
> 
> See [GitHub docs](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls#issues-and-pull-requests) for details.

If there are any other best practices or tips you think would be helpful for a new contributor like me, I would love to hear them! Thank you @coopernetes for correcting me.

PR #784, opened 2024-11-07
fix: update server start script for cross-platform compatibility

laukik-target:
**This PR updates the start script in package.json to ensure cross-platform compatibility by modifying the command syntax. The change addresses potential issues on different operating systems when running npm run server and npm run client concurrently.**

### Details of the Change:

The _**start**_ script was modified from single quotes to double quotes to make it compatible across Windows, macOS, and Linux environments:

`"start": "concurrently 'npm run server' 'npm run client'",`
_was updated to:_
`"start": "concurrently \"npm run server\" \"npm run client\"",`

### Reason for the Change:
- Single quotes in scripts can cause issues on Windows systems, as they are interpreted differently. Using double quotes ensures that the command runs consistently on all platforms.

### Impact:
- This change improves the reliability of the development workflow by preventing errors related to command parsing on Windows.

### Testing:
- Verified that the start script now runs without issues on both Windows and Unix-based systems.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "client": "vite --config vite.config.js",
     "clientinstall": "npm install --prefix client",
     "server": "node index.js",
-    "start": "concurrently 'npm run server' 'npm run client'",
+    "start": "concurrently \"npm run server\" \"npm run client\"",
     "build": "vite build",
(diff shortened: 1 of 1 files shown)


laukik-target:
@JamieSlome  one line change :) 
Please review and approve.

PR #785, opened 2024-11-07
Format Violations output on a blocked push

saibhotla:


JamieSlome:
@saibhotla - are you able to rebase and set all of your commits to your public Citi e-mail address and then force push? There is an e-mail in [this commit](https://api.github.com/repos/finos/git-proxy/commits/1f93410f632f14b018cadbaba0eb603a73151c61) that doesn't exist on your GitHub account, so needs replacing.

PR #786, opened 2024-11-08
fix: adjust auth processor to use custom service url

06kellyjac:
Now we can override the domains used by the service and proxy
the auth processor needs to use those values


JamieSlome (approved):
LGTM! 🍰

PR #789, opened 2024-11-12
feat: format violation text output on a blocked push 

saibhotla:
feat: format violations on block


test/scanDiff.test.js
```diff
@@ -0,0 +1,280 @@
+const chai = require('chai');
+const crypto = require('crypto');
```
JamieSlome, line 2: @saibhotla - it looks like we don't have this dependency installed?

JamieSlome: Ignore me! 👍 It is Node native.

JamieSlome (approved):
LGTM! 🍰

PR #793, opened 2024-11-13
refactor: remove exports from package.json and adjust imports

06kellyjac:
Exports makes it hard to access other paths in the project
Until we arrive on stable interfaces to access other parts of git-proxy
removing exports would be helpful.
Import paths have been adjusted slightly for plugins and the docs for
plugins due to the exports aliases being a bit different.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,11 +23,6 @@
     "git-proxy": "./index.js",
     "git-proxy-all": "concurrently 'npm run server' 'npm run client'"
   },
-  "exports": {
-    "./plugin": "./src/plugin.js",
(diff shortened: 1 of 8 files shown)


JamieSlome:
Seeing as we'd be breaking import statements, we'll make a minor bump to `v1.6.0` if we proceed 👍

JamieSlome (approved):
LGTM! 🍰

coopernetes (approved):
Agreed with the points raised here. `exports` can be re-introduced in a later release.

Need update the plugin docs as well once this is released.

PR #799, opened 2024-11-15
feat: dependency-check-plugin

gonuguntlakavya:
A plugin that uses dependency checker to check for the usage of vulnerable dependencies in git pushes. 
This PR is being submitted as part of citi finos hackathon


package.json
```diff
@@ -33,6 +33,8 @@
     "url": "https://github.com/finos/git-proxy"
   },
   "dependencies": {
+    "@finos/check-dependency-vulnerabilities": "file:plugins/check-dependency-vulnerabilities/finos-check-dependency-vulnerabilities-0.1.0-alpha.0.tgz",
+    "@finos/git-proxy-plugin-samples": "file:plugins/git-proxy-plugin-samples/finos-git-proxy-plugin-samples-0.1.0.tgz",
```
jescalada, line 37: I don't think we should use references to the filesystem in package.json. As shown in the [plugin docs](https://git-proxy.finos.org/docs/development/plugins/), we can add the plugin filename to the `plugins` config instead.

config.schema.json
```diff
@@ -8,6 +8,7 @@
     "proxyUrl": { "type": "string" },
     "cookieSecret": { "type": "string" },
     "sessionMaxAgeHours": { "type": "number" },
+    "dependencyVulnThreshold": { "type": "string" },
```
jescalada, line 11: This should be an entry within `plugins`, not a top-level config entry.

plugins/check-dependency-vulnerabilities/checkDependencyVuln.js
jescalada, line 1: This should probably be moved to `plugins/git-proxy-sample-plugins` since we don't necessarily want to include it in the base GitProxy setup.

plugins/check-dependency-vulnerabilities/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "Kavya",
+  "license": "Apache-2.0",
+  "type": "module",
+  "exports": {
+    ".": "./checkDependencyVuln.js"
+  },
+  "dependencies": {
```
jescalada, line 14: `dependency-check` should be added to the plugin dependencies, rather than the parent repo since the library won't be used outside the plugin.

plugins/check-dependency-vulnerabilities/README.md
jescalada, line 1: Would be great to run this through an LLM for fixing typos 👍🏼

jescalada (changes requested):
@gonuguntlakavya Thanks for the contribution!

Looks good so far, just a few comments on the implementation.

RichardUri:
@jescalada Created a new PR with changes regarding the comments and clean up. Please review
#1529

coopernetes:
Closed in favour of #1529

PR #806, opened 2024-11-20
Support secure interactions with private repositories 🔒

### LINKED ISSUE #552: Support secure interactions with private repositories 🔒
GitProxy is able to generate server-side diffs for public repositories without having to authorise with the upstream repository. This is because the nature of open source is to allow the cloning of public repositories directly without assignment of permissions.

To support private repositories that require a layer of authorisation, to generate a server side diff on behalf of the developer we have to consider how to safely and securely connect to the upstream and protected codebase without compromising the security of the developer's credentials or personal access token.

JamieSlome:
Closes #552

### DIFF
--- a/src/proxy/processors/push-action/pullRemote.js
+++ b/src/proxy/processors/push-action/pullRemote.js
@@ -1,7 +1,8 @@
-const spawnSync = require('child_process').spawnSync;
 const Step = require('../../actions').Step;
 const fs = require('fs');
 const dir = './.remote';
+const git = require('isomorphic-git');
+const http = require('isomorphic-git/http/node');
 
 const exec = async (req, action) => {
   const step = new Step('pullRemote');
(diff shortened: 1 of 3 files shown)


src/proxy/processors/push-action/pullRemote.js
```diff
@@ -19,20 +20,33 @@ const exec = async (req, action) => {
...
+      .clone({
+        fs,
+        http,
+        url: action.url,
+        onAuth: () => {
+          return {
+            username: username,
+            password: password,
+          };
+        },
```
06kellyjac, line 41: For a really tiny arrow function returning an object you can use this syntax.
Matching names in the object can also be inlined

```suggestion
        onAuth: () => ({
          username,
          password,
        }),
```

src/proxy/processors/push-action/pullRemote.js
```diff
@@ -19,20 +20,33 @@ const exec = async (req, action) => {
...
+        dir: `${action.proxyGitPath}/${action.repoName}`,
+      })
+      .then(() => {
+        console.log('Clone Success: ', action.url);
+      });
```
06kellyjac, line 46:
```suggestion
      });

      console.log('Clone Success: ', action.url);
```

Since there's an `await` above just putting the log on the line after is the same as putting it inside a `then(() => {})`

src/proxy/processors/push-action/pullRemote.js
```diff
@@ -19,20 +20,33 @@ const exec = async (req, action) => {
...
 
     step.log(`Completed ${cmd}`);
-    step.setContent(cloneOutput);
+    step.setContent(`Completed ${cmd}`);
```
06kellyjac, line 49: I'm not sure I'd refer to this as a command anymore necessarily but I guess what we're doing maps to a `git clone URL` anyway so should be fine

src/proxy/processors/push-action/pullRemote.js
```diff
@@ -1,7 +1,8 @@
-const spawnSync = require('child_process').spawnSync;
 const Step = require('../../actions').Step;
 const fs = require('fs');
 const dir = './.remote';
+const git = require('isomorphic-git');
+const http = require('isomorphic-git/http/node');
```
06kellyjac, line 5:
```suggestion
const gitHttpClient = require('isomorphic-git/http/node');
```

I'm hesitant to just call this `http` as it's probably not useful outside of isomorphic-git itself so this name might be a bit more descriptive.
But this action is also just for cloning with this library so we could leave this as `http` too :)

src/proxy/processors/push-action/pullRemote.js
```diff
@@ -19,20 +20,33 @@ const exec = async (req, action) => {
...
-    step.log(cloneOutput);
+    const authHeader = req.headers?.authorization;
+    const [username, password] = Buffer.from(authHeader.split(' ')[1], 'base64')
+      .toString()
+      .split(':');
+
+    await git
+      .clone({
+        fs,
+        http,
```
06kellyjac, line 34:
```suggestion
        http: gitHttpClient,
```

06kellyjac (commented):
LGTM. It's great to get private repo contributions back without embedding the creds in the `.git/config` :rocket: 

Added some thoughts but nothing major holding this back

PR #854, opened 2024-12-13
refactor(experimental): abstract data layer

06kellyjac:
Defined a database layer
Added an implementation using mongoose
Refactored the express routes to take an implementation
Added a mockable implementation
Refactored the tests to use the mockable implementation
Added tests for the get by uuid route

---

I'm not huge on OOP so the classes are more similar to go services
This method allows for easier mocking of the DB interactions with correct typechecking.
We also get to ensure we're using the same DB connection everywhere rather than trusting global variables.


JamieSlome (approved):
LGTM! 🍰

PR #896, opened 2025-02-04
fix: Database query built from user-controlled sources

JamieSlome:
Potential fix for [https://github.com/finos/git-proxy/security/code-scanning/83](https://github.com/finos/git-proxy/security/code-scanning/83)

To fix the problem, we need to ensure that the user-provided `id` is treated as a literal value in the MongoDB query. This can be achieved by using the `$eq` operator, which ensures that the value is interpreted as a literal and not as a query object. Additionally, we should validate that the `id` is a string before using it in the query.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._


src/service/routes/push.js
```diff
@@ -91,6 +91,10 @@ router.post('/:id/authorise', async (req, res) => {
 
   if (req.user && attestationComplete) {
     const id = req.params.id;
+    if (typeof id !== "string") {
+      res.status(400).send({ message: "Invalid ID format" });
+      return;
+    }
```
JamieSlome, line 97:
```suggestion
```

Issue #905, opened 2025-02-14
OIDC Login Implementation


jescalada:
**Is your feature request related to a problem? Please describe.**
Our organization would like to add OIDC support for both the frontend (route access) and the API (resource access). We have developed a working prototype for the login OIDC support (https://github.com/G-Research/git-proxy/pull/7) and are currently revamping the auth system to allow multiple auth strategies to coexist. Extending the auth methods will make git-proxy much more flexible.

**Describe the solution you'd like**
For now, we are adding the OIDC provider and configuration to the list of authentication methods available in `proxy.config.json` and adding an OIDC login button to the login page. Ideally, we would have multiple login methods (local auth, AD, OIDC, etc.) and users can pick whatever their organization provides.

**Describe alternatives you've considered**
Our organization cannot use local auth or AD, so adding OIDC support was our only option.

**Additional context**
We have a few issues and PRs for improving git-proxy's auth to be more flexible and secure, but we want to take it one step at a time and make sure our enhancements are valuable to the community, too! Let us know if there are any issues with our requests or if we should focus on improving specific parts of the project.

PR #906, opened 2025-02-14
feat: Preliminary OIDC implementation

jescalada:
This is a preliminary PR for #905. It adds the OIDC login button to the login page (/login) and the internals for OIDC. I would appreciate some feedback before I fill in the missing MongoDB implementation and clean up the `package-lock.json`.

To get this to work, I modified the `proxy.config.json`. I filled in the authentication section with my own OIDC provider config. If you want to use my own config from below, let me know your email so I can register you with my Google client.

Also, let me know if there is a better way to set the config! (I noticed `user-settings.json` being mentioned in the unit tests)

```json
"authentication": [
    {
      "type": "local",
      "enabled": false
    },
    {
      "type": "ActiveDirectory",
      "enabled": false,
      "adminGroup": "",
      "userGroup": "",
      "domain": "",
      "adConfig": {
        "url": "",
        "baseDN": "",
        "searchBase": ""
      }
    },
    {
      "type": "openidconnect",
      "enabled": true,
      "oidcConfig": {
        "issuer": "https://accounts.google.com/",
        "clientID": "...", // Add your client ID
        "clientSecret": "...", // Add your client secret
        "authorizationURL": "https://accounts.google.com/o/oauth2/auth",
        "tokenURL": "https://oauth2.googleapis.com/token",
        "userInfoURL": "https://openidconnect.googleapis.com/v1/userinfo",
        "callbackURL": "http://localhost:8080/api/auth/oidc/callback",
        "scope": "openid email profile"
      }
    }
  ],
```

**Note:** Currently, the above config will make existing unit tests fail (because it disables the local auth for OIDC to work). To make them pass again, set the local authentication in the config to "enabled": true. We're working on patching this up by refactoring to allow multiple auth methods to coexist (https://github.com/G-Research/git-proxy/issues/6).

If you feel like this PR is too barebones to merge in, I can add more improvements, like the multiple auth system, so that everything works well and is backward-compatible! Let me know if you prefer small, incremental PRs or completed and well-tested flows.

### Changelog
- Add OIDC login backend (endpoints, filesystem database functions) and frontend
  - Add trivial E2E test for OIDC, as it testing the whole flow is challenging
- Add checks for easier development
  - Why? To speed up refresh time by running the frontend separately in dev. Let me know if there's a cleaner way to handle it
- Fix minor issues to complete OIDC auth flow


package.json
```diff
@@ -63,6 +63,7 @@
     "passport": "^0.7.0",
     "passport-activedirectory": "^1.0.4",
     "passport-local": "^1.0.0",
+    "passport-openidconnect": "^0.1.2",
```
coopernetes, line 66: I have concerns with pulling in this library. It doesn't seem to be maintained (doesn't seem to have any release in the past 10 years) and likely lacks support for common flows such as auth code+PKCE. I would suggest investigating an alternative. Would you mind checking if the package linked below can be integrated via this PR instead?

- https://www.passportjs.org/packages/openid-client
- https://github.com/panva/openid-client/blob/main/examples/passport.ts

src/service/index.js
```diff
@@ -64,10 +64,16 @@ const createApp = async () => {
...
+
+  // In production, serves the static files from the build directory
+  if (process.env.NODE_ENV === 'production') {
+    app.use('/', express.static(absBuildPath));
+    app.get('/*', (req, res) => {
+      res.sendFile(path.join(`${absBuildPath}/index.html`));
+    });
+  } else {
+    console.log('Not serving static files');
+  }
```
coopernetes, line 76: Is this section necessary for testing? I'm genuinely not sure what this is for. 

If it must remain in the code, this is probably best left to be read from Git Proxy's own configuration as a user setting as opposed to using `NODE_ENV`[^1].

[^1]: https://nodejs.org/en/learn/getting-started/nodejs-the-difference-between-development-and-production#why-is-node_env-considered-an-antipattern

src/service/routes/auth.js
```diff
@@ -3,6 +3,13 @@ const router = new express.Router();
 const passport = require('../passport').getPassport();
 const db = require('../../db');
 const passportType = passport.type;
+const { GIT_PROXY_UI_HOST: uiHost = 'http://localhost', NODE_ENV } = process.env;
+
+// TODO: Refactor this through proper .env loading. This handles redirects in dev.
+let uiPort = 3000;
+if (NODE_ENV === 'production') {
+  uiPort = process.env.GIT_PROXY_UI_PORT;
+}
```
coopernetes, line 12: Directly assign to `uiPort` from process.env with a default in that assignment line instead of wrapping it in a `NODE_ENV` based check.
```suggestion
const { GIT_PROXY_UI_HOST: uiHost = 'http://localhost', GIT_PROXY_UI_PORT: uiPort = 3000 } = process.env;
```

coopernetes:
Thank you for the contribution @jescalada ! Glad to see the addition of OIDC login support to enable the use of more identity providers :grin:

When you get a chance, please ensure to sign the CLA so that this PR can be eventually merged.

package.json
jescalada: Thank you for the advice! You're right, `openid-client` is a lot more modern. I don't think it should be hard to replace since they're implemented similarly. I'll get right to it 🙂

src/service/index.js
jescalada: This was due to a misunderstanding when I set up my dev environment. I think my UI port was not set correctly, so hitting some frontend routes would try to query the static files (and return 404).

The line you mentioned below fixed it for me:

```js
const { GIT_PROXY_UI_HOST: uiHost = 'http://localhost', GIT_PROXY_UI_PORT: uiPort = 3000 } = process.env;
```

src/service/routes/auth.js
jescalada: Thank you!

jescalada:
@coopernetes It was tough to integrate `openid-client` since the latest version (v6) completely changed how auth works, so the documentation was sparse. However, I did manage to get the whole flow working again.

Some additions were the manual `sub` check (the previous library perhaps did this automatically?) and the manual user info fetch to handle user serialization.

Let me know what you think!

Things will make more sense after integrating the multiple auth method system, as this PR will force you to use OIDC exclusively.

JamieSlome:
@06kellyjac (cc) 🍰

src/service/passport/index.js
```diff
@@ -14,10 +15,15 @@ const configure = async () => {
     case 'local':
       _passport = await local.configure();
       break;
+    case 'openidconnect':
+      _passport = await oidc.configure();
+      break;
     default:
       throw Error(`uknown authentication type ${type}`);
   }
-  _passport.type = authenticationConfig.type;
```
JamieSlome, line 20: @jescalada - why the change here?

src/service/passport/oidc.js
JamieSlome, line 1: Can we address the lint warnings pulled up here?

JamieSlome (commented):
Solid PR! 🍰 A couple of comments and requested changes 👍

src/service/passport/index.js
jescalada: @JamieSlome The type is being defined twice, once in each `configure` function and again in this file. It's just a temporary fix since I'm getting rid of it entirely in the multiple auth types PR!

jescalada:
Hi @JamieSlome and @coopernetes, our CLA got cleared, so I'd appreciate one final look! 😃

JamieSlome (approved):
LGTM! 🚀 🍰

JamieSlome:
@jescalada - can we look at the Codecov above and address as some technical debt? 🤝 

I'll package this feature up and release under `v1.9.0` 👍

jescalada:
@JamieSlome Absolutely! Once all the OIDC-related features are merged, I can do a final PR to increase test coverage. We have 4 more bite-sized OIDC/Auth/JWT PRs and no additional ones for the foreseeable future.

Thanks for taking a look at this one! 😃

JamieSlome:
Great, thanks! ❤️@jescalada

PR #908, opened 2025-02-17
feat: Add support for executing pre-receive hooks

fabiovincenzi:
This PR adds support for managing pre-receive hooks by executing _.sh_ scripts located in the _./hooks_ directory. These scripts allow administrators to define conditions for automatically approving or rejecting pushes.

**Changes Introduced**

- Implement execution of _.sh_ scripts in the _./hooks_ directory to validate pushes.
- Add basic tests to verify hook execution functionality.

You can also test this feature using the following [pre-receive samples](https://github.com/github/platform-samples/tree/master/pre-receive-hooks). Simply place a sample script in the _./hooks_ directory and attempt a push. The script should execute and determine whether the push is accepted or rejected.


JamieSlome:
@06kellyjac (cc) 🍰

package.json
```diff
@@ -9,7 +9,7 @@
     "server": "node index.js",
     "start": "concurrently \"npm run server\" \"npm run client\"",
     "build": "vite build",
-    "test": "NODE_ENV=test mocha --exit",
+    "test": "NODE_ENV=test mocha './test/**/*.js' --exit",
```
JamieSlome, line 12: Are we sure the previous command wasn't dipping into other test folders to run test specs? Happy otherwise 👍

package.json
```diff
@@ -91,7 +91,7 @@
     "eslint-plugin-react": "^7.21.5",
     "eslint-plugin-standard": "^5.0.0",
     "husky": "^9.0.0",
-    "mocha": "^10.2.0",
+    "mocha": "^10.8.2",
```
JamieSlome, line 94: Not a huge deal but can we remove the Mocha update?

src/proxy/processors/push-action/preReceive.js
```diff
@@ -0,0 +1,54 @@
...
+    });
+
+    const { stdout, stderr, status } = hookProcess;
+
+    if (status !== 0) {
+      step.error = true;
+      step.log(`Hook stderr: ${stderr.trim()}`);
+      step.setError(stdout.trim());
+      action.addStep(step);
+      return action;
```
JamieSlome, line 39: Can we take a look at the code coverage warning here? If not particularly important, give me the heads up 👍

JamieSlome (commented):
Another solid G-PR 😏 

Can you confirm my understanding? This PR supports the definition and execution of `.sh` scripts which can be exposed to the chain via the `hooks` folder?

JamieSlome:
Just a reminder to get the EasyCLA signed and we're good to proceed 🍰

package.json
fabiovincenzi: Yes, the previous command wasn't resolving subdirectories, I just checked again.

package.json
fabiovincenzi: Sure!

package.json
JamieSlome: Great, thanks @fabiovincenzi ❤️

src/proxy/processors/push-action/preReceive.js
fabiovincenzi: I think now it should be ok👍

fabiovincenzi:
> Another solid G-PR 😏
> 
> Can you confirm my understanding? This PR supports the definition and execution of `.sh` scripts which can be exposed to the chain via the `hooks` folder?

Thank you @JamieSlome for the review!
Exactly that is the flow🙌🏻

JamieSlome:
Thanks for the clarifications and updates ❤️ Once the EasyCLA is signed I'll merge the PR 🍰

fabiovincenzi:
@JamieSlome  I added a check to verify whether the pre-receive hook exists. If it doesn’t, we simply skip its execution without any errors. This way, the check is only enforced if both the hooks folder and the pre-receive file are present. What do you think?

Aside from that, we were considering implementing auto-approval of pushes using this hook. I'll create a separate PR for that, but perhaps adding a configuration option like `enable-auto-approval: true` could allow pushes to be automatically approved and pushed to the upstream.

We are working to get the EasyCLA signed, thanks again!😊

fabiovincenzi:
@JamieSlome, All set! EasyCLA is now signed.

JamieSlome (approved):
LGTM! 🍰

PR #917, opened 2025-02-21
fix: add authorization for UI routes

### LINKED ISSUE #918: Add UI route authorization and decouple admin routes
To improve security and make access more granular, we'd like to add authorization to UI routes. In other words, we want users to log in if they wish to access any route. Furthermore, only admins should be able to see certain resources (such as the list of users and their details, changing a user's admin status, and so on).

Currently, even unlogged users can see the list of users, access admin routes, etc. We're not sure if this is by design, but we feel it would help to make access more strict.

A basic frontend route guard could check a user's auth and admin status and redirect them accordingly (to the login page, "unauthorized" page, or actual page).

None

**Additional context**
Our biggest concern is how our route changes will affect current users. Are users currently using git-proxy without logging in or distinguishing between admins and non-admins?

Our changes may affect their current workflow, so ideally we want to make things as intuitive as possible and back them up with documentation so they can keep using the library as usual.

jescalada:
Fixes #918.

This PR fixes the UI route access issue (unlogged users and non-admins can no longer access any route). In other words, only admins should be able to access the list of users, and only logged-in users can see the rest. Unlogged users can only access the login page.

You'll see an Unauthorized page when accessing a route you don't have access to.

Another issue we'd like to fix is the layout and the admin routes, which are coupled. Thus, all routes are called `/admin/*` when we only want the actual admin routes to have this naming (and replace regular routes with `/dashboard/`). To keep things bite-sized, this will be pushed in a separate PR.

It builds on top of the OIDC implementation at #905. As mentioned in that PR, testing with OIDC will require editing the config.

Tests will be modified/added in the layout fix PR since the routes will change.

Thank you!

### Changelog
- Added `AuthProvider`
  - To provide context with the current user
- Added `PrivateRoute`
  - A wrapper that handles route authorization/redirection
- Added `NotAuthorized` and `NotFound` pages
  - To prevent showing blank pages or redirecting (which is confusing)


JamieSlome:
Thanks for the PR @jescalada and apologies for the delay! From a product perspective, I understand the position of closing down certain routes. My reason for opening them up to all users was to encourage "openness" when engaging with open source, and driving transparency on what projects are being contributed to and by whom. That said, I also appreciate the purpose of this change.

Perhaps we make it configurable?

jescalada:
@JamieSlome That makes sense! I'll make it optional. I'll modify #961, which contains the changes in this PR.

sam-holmes2:
@jescalada please could you confirm you're happy that these configurable changes are included in #961? If so guess this older PR can be closed?

jescalada:
@sam-holmes2 Sounds good!

PR #935, opened 2025-03-10
feat: implements config loader to enable remote or external configs

### LINKED ISSUE #934: Allow for the configuration to be sourced externally and updated at runtime
Instead of a static `proxy.config.json`, we would like to have a way to provide the configuration from an external source (like a gitops repo, or an http service, etc.) and allow it to be reloaded/applied at runtime.

The idea is to implement a robust remote configuration management solution for GitProxy that would support dynamic loading of configurations from multiple sources (local files, HTTP endpoints, and Git repositories). The system would maintain a default configuration in `proxy.config.json` which could be overridden by user-specific settings, and would feature a dynamic configuration loader that periodically checks for updates at configurable intervals. When changes are detected, the system would validate the new configuration, gracefully stop existing services, apply the updates, and restart services with the new configuration. The solution would include built-in error handling with automatic rollback to previous configurations if updates fail, and would provide an admin-only endpoint for manual configuration reloading. This architecture would ensure that configuration changes could be managed centrally while maintaining system stability and availability, with the ability to merge configurations from multiple sources when needed.

None

**Additional context**
The idea behind remote configuration management for GitProxy is to enable automated, remote configuration updates without requiring manual intervention or service restarts. This would allow administrators to centrally manage and update configurations across multiple GitProxy instances, ensuring consistent settings and reducing operational overhead. The system would automatically detect and apply configuration changes from remote sources while maintaining system stability through validation and rollback capabilities.

dcoric:
This pull request introduces a significant update to the configuration management system of the GitProxy application. The main changes include adding support for multiple configuration sources, implementing a new configuration loader class, and providing the ability to reload configurations without restarting the application.

### Configuration Management Enhancements:

* Added support for multiple configuration sources in `config.schema.json` and `proxy.config.json`, allowing the configuration to be sourced from files, HTTP endpoints, or Git repositories. [[1]](diffhunk://#diff-be1695b1e63a508d59982601f9e1fb7f58247deecb1e427adb77bcad758ae5e5R8-R19) [[2]](diffhunk://#diff-c465aafa0fe603e2d28b017938f55e5ce3253aac7aa303efeabfc06a4ad52d5fR5-R42)

* Introduced a new `ConfigLoader` class in `src/config/ConfigLoader.js` to manage the loading and reloading of configurations from the specified sources. This class handles periodic reloading, merging configurations, and emitting events on configuration changes.

### CLI and API Improvements:

* Added a new CLI command `reload-config` in `packages/git-proxy-cli/index.js` to trigger a configuration reload without restarting the process. This command checks for authentication and makes a POST request to the admin API endpoint to reload the configuration. [[1]](diffhunk://#diff-ee51eb1c2264303569f2a8fa9f5bb2de1b6b51eed24eed943d079f341f4139b0R309-R331) [[2]](diffhunk://#diff-ee51eb1c2264303569f2a8fa9f5bb2de1b6b51eed24eed943d079f341f4139b0R462-R466)

* Added a new admin-only API endpoint `/api/v1/admin/reload-config` in `src/service/index.js` to handle configuration reload requests. This endpoint validates the request, reloads the configuration, and restarts the services if necessary.

### Proxy Service Enhancements:

* Updated the proxy service in `src/proxy/index.js` to support stopping and restarting the HTTP and HTTPS servers, enabling the application to apply new configurations without a full restart.

### Configuration Access and Updates:

* Refactored `src/config/index.js` to use the new `ConfigLoader` and updated existing getter functions to retrieve values from the current configuration. Added functionality to handle configuration updates dynamically and restart services as needed. [[1]](diffhunk://#diff-1523e3000d64e09cd6eb8370a4568715090b6763fd9192f61ab6044ac6d86fa9R2-R3) [[2]](diffhunk://#diff-1523e3000d64e09cd6eb8370a4568715090b6763fd9192f61ab6044ac6d86fa9L10-R123) [[3]](diffhunk://#diff-1523e3000d64e09cd6eb8370a4568715090b6763fd9192f61ab6044ac6d86fa9R143)fix: config loader clone command issue

This PR fixes #934

### DIFF
--- a/proxy.config.json
+++ b/proxy.config.json
@@ -136,7 +136,7 @@
   "csrfProtection": true,
   "plugins": [],
   "tls": {
-    "enabled": true,
+    "enabled": false,
     "key": "certs/key.pem",
     "cert": "certs/cert.pem"
   }
--- a/proxy.config.json
+++ b/proxy.config.json
@@ -96,6 +96,39 @@
       }
     ]
   },
+  "configurationSources": {
+    "enabled": false,
+    "reloadIntervalSeconds": 60,
+    "merge": false,
+    "sources": [
+      {
+        "type": "file",
+        "enabled": false,
+        "path": "./external-config.json"
+      },
+      {
+        "type": "http",
+        "enabled": false,
+        "url": "http://config-service/git-proxy-config",
+        "headers": {},
+        "auth": {
+          "type": "bearer",
+          "token": ""
+        }
+      },
+      {
+        "type": "git",
+        "enabled": false,
+        "repository": "https://git-server.com/project/git-proxy-config",
+        "branch": "main",
+        "path": "git-proxy/config.json",
+        "auth": {
+          "type": "ssh",
+          "privateKeyPath": "/path/to/.ssh/id_rsa"
+        }
+      }
+    ]
+  },
   "domains": {},
   "privateOrganizations": [],
   "urlShortener": "",
--- a/website/docs/configuration/reference.mdx
+++ b/website/docs/configuration/reference.mdx
@@ -648,5 +648,105 @@ description: JSON schema reference documentation for GitProxy
 </blockquote>
 </details>
 
-----------------------------------------------------------------------------------------------------------------------------
+<details>
+<summary>
+<strong> <a name="configurationSources"></a>19. [Optional] Property GitProxy configuration file > configurationSources</strong>
+</summary>
+<blockquote>
+
+|                           |                                                         |
+| ------------------------- | ------------------------------------------------------- |
+| **Type**                  | `object`                                                |
+| **Required**              | No                                                      |
+| **Additional properties** | [[Not allowed]](# "Additional Properties not allowed.") |
+
+**Description:** Configuration for dynamic loading from external sources
+
+<details>
+<summary>
+<strong> <a name="configurationSources_enabled"></a>19.1. [Optional] Property configurationSources > enabled</strong>
+</summary>
+<blockquote>
+
+|              |           |
+| ------------ | --------- |
+| **Type**     | `boolean` |
+| **Required** | No        |
+
+**Description:** Enable/disable dynamic configuration loading
+
+</blockquote>
+</details>
+
+<details>
+<summary>
+<strong> <a name="configurationSources_reloadIntervalSeconds"></a>19.2. [Optional] Property configurationSources > reloadIntervalSeconds</strong>
+</summary>
+<blockquote>
+
+|              |          |
+| ------------ | -------- |
+| **Type**     | `number` |
+| **Required** | No       |
+
+**Description:** How often to check for configuration updates (in seconds)
+
+</blockquote>
+</details>
+
+<details>
+<summary>
+<strong> <a name="configurationSources_merge"></a>19.3. [Optional] Property configurationSources > merge</strong>
+</summary>
+<blockquote>
+
+|              |           |
+| ------------ | --------- |
+| **Type**     | `boolean` |
+| **Required** | No        |
+
+**Description:** When true, merges configurations from all enabled sources. When false, uses the last successful configuration load
+
+</blockquote>
+</details>
+
+<details>
+<summary>
+<strong> <a name="configurationSources_sources"></a>19.4. [Optional] Property configurationSources > sources</strong>
+</summary>
+<blockquote>
+
+|              |         |
+| ------------ | ------- |
+| **Type**     | `array` |
+| **Required** | No      |
+
+**Description:** Array of configuration sources to load from
+
+Each item in the array must be an object with the following properties:
+
+- `type`: (Required) Type of configuration source (`"file"`, `"http"`, or `"git"`)
+- `enabled`: (Required) Whether this source is enabled
+- `path`: (Required for `file` type) Path to the configuration file
+- `url`: (Required for `http` type) URL of the configuration endpoint
+- `repository`: (Required for `git` type) Git repository URL
+- `branch`: (Optional for `git` type) Branch to use
+- `path`: (Required for `git` type) Path to configuration file in repository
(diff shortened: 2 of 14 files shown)


dcoric:
I will rebase to the latest main branch, check the failing test, and post an update afterward.

src/config/ConfigLoader.js
```diff
@@ -0,0 +1,167 @@
...
+
+    const tempDir = path.join(process.cwd(), '.git-config-cache');
+    await fs.promises.mkdir(tempDir, { recursive: true });
+
+    const repoDir = path.join(tempDir, Buffer.from(source.repository).toString('base64'));
+
+    // Clone or pull repository
+    if (!fs.existsSync(repoDir)) {
+      if (source.auth?.type === 'ssh') {
+        process.env.GIT_SSH_COMMAND = `ssh -i ${source.auth.privateKeyPath}`;
```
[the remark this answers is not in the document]
06kellyjac: Also rather than setting this on the `process.env` which will stick around globally for the rest of `git-proxy`'s execution it'd be best to set `env:` inside of `execFileAsync` `options` e.g.:

```js
await execFileAsync(
  'git',
  ['clone', source.repository, repoDir],
  {
    env: {
      ...process.env,
      GIT_SSH_COMMAND = `ssh -i ${source.auth.privateKeyPath}`,
    },
  },
);
```

src/config/ConfigLoader.js
```diff
@@ -0,0 +1,167 @@
...
+
+  async start() {
+    const { configurationSources } = this.config;
+    if (!configurationSources?.enabled) {
+      return;
+    }
+
+    // Start periodic reload if interval is set
+    if (configurationSources.reloadIntervalSeconds > 0) {
+      this.reloadTimer = setInterval(
```
06kellyjac, line 25: if `this.reloadTimer` happens to already have an interval (e.g. start is called twice by accident) maybe it should clear it?
Since we don't intend to do that maybe we don't need to account for it, just a defensive suggestion.

src/config/ConfigLoader.js
```diff
@@ -0,0 +1,167 @@
...
+  async loadFromGit(source) {
+    // Validate inputs
+    if (!source.repository || typeof source.repository !== 'string') {
+      throw new Error('Invalid repository URL');
+    }
+    if (source.branch && typeof source.branch !== 'string') {
+      throw new Error('Invalid branch name');
+    }
+
+    const tempDir = path.join(process.cwd(), '.git-config-cache');
```
06kellyjac, line 118: It'd be ideal if we could follow XDG conventions (and others on other OSes) rather than `$PWD`

https://github.com/sindresorhus/env-paths

src/config/ConfigLoader.js
06kellyjac: Looks like this could possibly happen with the `// Attempt to restart with existing config if reload fails` case

proxy.config.json
```diff
@@ -2,17 +2,44 @@
   "proxyUrl": "https://github.com",
   "cookieSecret": "cookie secret",
   "sessionMaxAgeHours": 12,
+  "configurationSources": {
+    "enabled": false,
```
06kellyjac, line 6: Current default behavior is to read `$PWD/proxy.config.json` automatically. Does it still do that?

06kellyjac (changes requested):
Some suggestions and questions

src/config/ConfigLoader.js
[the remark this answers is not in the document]
dcoric: This is a good tip. I have added validations for both paths and git URL & branch name

src/config/ConfigLoader.js
dcoric: You are correct. It is fixed now.

src/config/ConfigLoader.js
dcoric: I have added env-paths to use for this. Looks like a good approach.
I was thinking about using os.tmpdir approach as well, but I'm not sure if it would behave correctly in all envs:

```javascript
const os = require('os');

const tempDir = os.tmpdir();
```

proxy.config.json
dcoric: Yes, original behavior is kept. On load, the app loads the default `proxy.config.json` and if remote configs are enabled, it will try to fetch them. Once they are fetched, if the merge option is enabled, it will do the merging (with local and remote configs) and validate the new configuration before applying them.
If remote configs are set to false, everything works as it did previously. There will be no changes to functionality of the configuration file.

JamieSlome:
@dcoric - can we bolster up the code coverage? 🚀

dcoric:
> @dcoric - can we bolster up the code coverage? 🚀

Done. 
This helped to identify (and fix) some potential issues with validation functions 👏

JamieSlome:
@dcoric - thanks for attending the call today! As discussed, let's take a look at the failing CI and we can get this merged 👍

JamieSlome (approved):
LGTM! 🍰

JamieSlome:
@dcoric - are you seeing merge conflicts?

dcoric:
> @dcoric - are you seeing merge conflicts?

Yes, on it right now.

website/docs/configuration/reference.mdx
```diff
@@ -648,5 +648,105 @@ description: JSON schema reference documentation for GitProxy
 </blockquote>
 </details>
 
-----------------------------------------------------------------------------------------------------------------------------
+<details>
+<summary>
```
kriswest, line 652: @dcoric was this content written manually? Its supposed to be generated from the config schema, but I'm not seeing matching info added to the schema above. Hence, it will all be lost the next time that the ref doc is re-generated (`npm run gen-schema-doc` - requires you to have a python env setup for it to run in). I'm doing that on another PR that adds to the config schema so that will happen soon... A new PR is probably needed to correct this @JamieSlome @sam-holmes2

PR #944, opened 2025-03-21
fix: missing query parameter in file db

fabiovincenzi:
The `getUsers` function in the file db did not correctly use the provided query parameter. 
Adjusted `addGitPushToDb` function by allowing the user parameter to be explicitly passed.

### DIFF
--- a/packages/git-proxy-cli/test/testCli.test.js
+++ b/packages/git-proxy-cli/test/testCli.test.js
@@ -645,10 +645,12 @@ describe('test git-proxy-cli', function () {
 
   describe('test git-proxy-cli :: git push administration', function () {
     const pushId = `0000000000000000000000000000000000000000__${Date.now()}`;
+    const gitAccount = 'testGitAccount1';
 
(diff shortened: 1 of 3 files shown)


JamieSlome:
@fabiovincenzi - can we take a look at the code coverage warning? 👍 

Nice PR otherwise 😄

kriswest:
+1 for this fix I was going to contribute the same as the file-based DB is not usable without it as the failure to pass through the query is breaking push parsing here (when there is more than one user in the DB the lack of the query causes all users to be returned a fails the `list.length == 1` check):
https://github.com/finos/git-proxy/blob/5d24d9dbd0ea22ac81b3b4288391cb7aee978c5d/src/proxy/processors/push-action/checkUserPushPermission.js#L13-L20

I have a few additions to contribute - casing is handled inconsistently - which I'll raise a separate issue/PR for

JamieSlome (commented):
LGTM! 🍰

PR #961, opened 2025-03-28
fix: decouple UI layout from admin routes

### LINKED ISSUE #918: Add UI route authorization and decouple admin routes

jescalada:
Fixes #918, along with #917.

**Note:** The UI has been adapted somewhat to our use case. Feel free to suggest ways to make this more general or reject the changes if you think it would affect current users' workflow more than desired.

This PR aims to make route access more intuitive. All /admin routes have been replaced with /dashboard, and relevant files have been renamed. Actual admin-only routes, such as the list of users, are displayed with the /admin/ prefix. For example: /dashboard/admin/user.

Unit tests execute the same way as before, and E2E tests have been expanded and amended to work with the new setup.

I also changed the login flow to redirect to the repo list (which seems more intuitive than one's profile), and the username update flow to redirect to one's own profile instead of the "user view" admin route (which would error out for non-admins).

When testing, keep in mind that enabling OIDC will make some tests fail. Please set the local auth in proxy.config.json to enabled: true (and openidconnect to enabled: false) to make tests pass. Thank you!

### Changelog
- Refactor "admin" layout/routes into "dashboard"
- Fix redirect on login (now redirects to repo list, which requires a logged-in user)
- Fix error handling on repo page
  - Catches GitHub repo retrieval error and makes it more descriptive
- Add new E2E test for basic login flow (with admin/admin account) and redirect
- Fix login Cypress command
  - Existing repo E2E test now requires a logged-in user


JamieSlome:
@jescalada - can you resolve conflicts and we will get this merged? 👏

jescalada:
**Note:** I'm working on making the admin routes optional (through a config entry), as Jamie requested in #917.

jescalada:
Hi @JamieSlome, it should be good to review now! I added the config entry, new `RouteGuard` logic, and updated the PR description.

JamieSlome:
Do we agree that this would be considered a minor bump? @jescalada

jescalada:
@JamieSlome Yes, it should be backward compatible, so minor bump is fine!

Also, #963 which is related is ready for a final look/merge. Thank you!

sam-holmes2 (approved):
LGTM - thanks for your contribution 🥳

Issue #950, opened 2025-03-24
Repositories should be identified and retrieved via URL rather than name


kriswest:
**Describe the bug**
This issue relates to/overlaps with two past unresolved issues:
- #66
- #511
  - #545  

Throughout the codebase repositories are assumed to have a unique `name` property. However, repository names are not globally unique and there are two common cases where they will be duplicated:

1) Multiple forks of the same project (e.g. within Github, bearing two different user's names as the 'organisation' (issue #66)
2) Different projects that have the same name, hosted in different platforms.
  - This is part of, but not the whole of issues #545 - the same change is needed.

What we can rely on being unique is the repository URL being pushed to. 

In PR #545 where @msagi was looking at this, it is suggested that we move to a platform/org/reponame structure. However, this requires additional processing of the URL and assumes that the platform/org can be extracted from the URL - the org is a GitHub/GitLab concept which doesn't apply to other hosts/base git repos.

On the other hand, a URL is globally unique and already exists within the data model. Switching to it is to identify repositories is easy enough - except where it impacts the URLs used to contribute to those repos through git proxy, e.g.:
`git clone http://localhost:8000/finos/git-proxy.git`
would need to become:
`git clone http://localhost:8000/https://github.com/finos/git-proxy.git`

This URL is uglier but easier to understand - the git proxy url has simply been prepended to it.

The challenge would be that you can't automatically migrate existing users with a checkout configured with the eexisting URL as origin (at least AFAIK - github manages repo URL migrations so there may be a way in the git protocol - although I suspect they just maintain some redirects). Hence, to solve this without a breaking change, there may need to be some fallback code to handle older checkouts and/or some form of migration. I'd like to figure out the preferred route forward on that before undertaking work to resolve (as I can see two others took this on and didn't bring it to ground). Options:
1. Fallback to existing approach: Switch to repo URL for retrieval and if repo not found/URL didn't contain the full url (so no 'http' in the string) attempt lookup by name.
2. Advise users to migrate: Create a table of the existing repo names at the time of migration. Switch to repo URL for retrieval and if repo not found, check the table and respond with a message advising user to migrate to the new URL.
3. Issue a breaking change and advise firms updating to inform all users that their repos need to migrated to the new origins.

**Expected behaviour**
- Multiple forks of a project can be onboarded, managed and contributed to separately
- Multiple repos with the same name, but oowned by different orgs or hosted on different platforms can be onboarded, managed and contributed to separately.

**Additional context**
[...]

PR #973, opened 2025-04-11
fix: 946 associate commits by email 

### LINKED ISSUE #946: Commits should be associated with users via email rather than the git user.name config
**Describe the bug**
Git commits carry two pieces of identifying information for the users that authored and committed them, that are set in the git client's config: `user.name` and `user.email`. Throughout the Git Proxy codebase it is assumed that `user.name` property relates to a GitHub user id and is used to look up the user in the database - the two most important places are:

https://github.com/finos/git-proxy/blob/5d24d9dbd0ea22ac81b3b4288391cb7aee978c5d/src/proxy/processors/push-action/checkUserPushPermission.js#L10-L20

and 

https://github.com/finos/git-proxy/blob/5d24d9dbd0ea22ac81b3b4288391cb7aee978c5d/src/service/routes/push.js#L45-L62

Whilst the `user.name` _may_ be set to the user's github user account, it doesn't have to be. Github doesn't actually expect or require it and the GIT SCM book actually demonstrates a plan name for a user (i.e. <firstname> <lastname> being set under first time git setup: https://git-scm.com/book/ms/v2/Getting-Started-First-Time-Git-Setup. Further, the assumption that it is the username bakes in reliance on GitHub user accounts and will cause issues applying GitProxy to other platforms such as GitLab (we'll need multiple fields where users contribute into multiple platforms and hence have multiple usernames).

Instead of the GitHub username, commits should be matched to user by email address. GitHub, GitLab and EasyCLA all associate commits with users via email - for example I've had my client setup with user.anem "Kris West", where my user name is @kriswest for years and had not had issues with GitHub or EasyCLA. The user email is readily available on commits (right next to the `user.name`) and already needs to be accurate for association with GitHub user accounts and projects that validate contributions with EasyCLA or CLABot.

**To Reproduce**
1. Setup you git client as advised in the Git SCM book:
    ```bash
    $ git config --global user.name "John Doe"
    $ git config --global user.email johndoe@example.com
    ```
2. Create a user in git proxy with the given email and set the GitHub username so that it does not match `user.name`, e.g. @johndoe
3. Give the user permission to push to a repository
4. Raise and push a commit - it'll be refused as 'User @johndoe is not allowed to push on repo <repo>, ending'

**Expected behavior**
User should be identified by email correctly and allowed to push on the repo.

**Additional Context**
I'll raise a PR to resolve this issue - it affects a small number, but critical set of operations and can be resolved with the existing data model and no need to migrate existing installations.

kriswest:
resolves #946

Updates push parsing to extract commiter emails and to associate commits with users via email, rather than the git user.name config.

Note that:

>  The Git username is not the same as your GitHub username.

https://docs.github.com/en/get-started/git-basics/setting-your-username-in-git

and

>GitHub uses your commit email address to associate commits with your account on GitHub. 

https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-email-preferences/setting-your-commit-email-address

### DIFF
--- a/test/processors/checkUserPushPermission.test.js
+++ b/test/processors/checkUserPushPermission.test.js
@@ -14,7 +14,6 @@ describe('checkUserPushPermission', () => {
 
   beforeEach(() => {
     logStub = sinon.stub(console, 'log');
-
     getUsersStub = sinon.stub();
     isUserPushAllowedStub = sinon.stub();
 
@@ -108,10 +107,14 @@ describe('checkUserPushPermission', () => {
 
     it('should return error when no user is set in the action', async () => {
       action.user = null;
+      action.userEmail = null;
+      getUsersStub.resolves([]);
       const result = await exec(req, action);
       expect(result.steps).to.have.lengthOf(1);
       expect(result.steps[0].error).to.be.true;
-      expect(result.steps[0].errorMessage).to.include('Push blocked: User not found. Please contact an administrator for support.');
+      expect(result.steps[0].errorMessage).to.include(
+        'Push blocked: User not found. Please contact an administrator for support.',
+      );
     });
   });
 });


kriswest:
Looks like I didn't check the CLI test cases and need to sort something out there

kriswest:
Fixed the failing tests but am waiting on a review (in git proxy), will update monday

kriswest:
Code coverage is as good as I can get it, will pick up a little more when #979 is merged

kriswest:
@JamieSlome This is the PR we discussed expediting the review of - it'll be needed for supporting git repositories other than Github (work that is going well this week - but will generate another fairly large PR soon).

src/service/routes/push.js
```diff
@@ -86,6 +83,8 @@ router.post('/:id/authorise', async (req, res) => {
   const questions = req.body.params?.attestation;
   console.log({ questions });
 
+  // TODO: compare attestation to configuration and ensure all questions are answered
+  // - we shouldn't go on the definition in the request!
   const attestationComplete = questions?.every((question) => !!question.checked);
   console.log({ attestationComplete });
```
sam-holmes2, line 89: Good spot, should this be created as a separate issue?

sam-holmes2 (approved):
Thanks for your contribution! LGTM but may need additional approval before being merged.

src/service/routes/push.js
kriswest: yes, will create that now. Thanks for flagging, had forgotten about that.

kriswest: Here you go @sam-holmes2: https://github.com/finos/git-proxy/issues/1041
Thanks for the reminder

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -35,9 +35,10 @@ async function exec(req: any, action: Action): Promise<Action> {
       action.commitFrom = action.commitData[action.commitData.length - 1].parent;
     }
 
-    const user = action.commitData[action.commitData.length - 1].committer;
-    console.log(`Push Request received from user ${user}`);
-    action.user = user;
+    const {committer, committerEmail} = action.commitData[action.commitData.length - 1];
```
jescalada: Although we aren't enforcing it in the CI, we might want to run `npm run format` to keep things clean! 👍🏼

JamieSlome: @kriswest ^

kriswest: Isn't husky applying prettier on save already?

hmm... apparently not as running that touches 129 files! I'll check that in.

JamieSlome: Hmm, I wonder if there is an incompatibility between developer machines as we seem to get prettier changes drifting back and forth between pull requests?

jescalada: That's weird, when pulling your branch and running `npm run format`, I only get about ~20 files changed...

kriswest: I think all the rest have whitespace changes only

kriswest:
@JamieSlome AFAIK this is good to go. We've been using it for more than a month.

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -111,6 +112,9 @@ const getCommitData = (contents: CommitContent[]) => {
       const authorEmail = author?.split(' ').reverse()[2].slice(1, -1);
       console.log({ authorEmail });
 
+      const committerEmail = committer?.split(' ').reverse()[2].slice(1, -1);
```
06kellyjac: since committer has already been split and reversed on line 109 can this be reused please. And grouping the committer stuff is easier to read.

```js
const committerArr = committer?.split(' ').reverse()
const commitTimestamp = committerArr[1]
// slice to trim start and end from `<an-email@address.whatever>`
const committerEmail = committerArr[2].slice(1, -1)
```

Also a comment for why the slice is relevant. I'm assuming it's for `<>` or something else wrapping the email so I've proposed one.

06kellyjac (commented):
Thanks for fixing the typescript build error. One more comment but otherwise we're good to merge.

src/proxy/processors/push-action/parsePush.ts
kriswest: Good call, theres some other similar rationalisation that can be applied to that function (cleanup of `author` and `committer`, grouping the operations better etc.) which I've also applied.

I've reduce some of the (excessive and duplicative IMHO) log messaging to the debug level. TBH I think we could remove all but the first and last log message (input data and output data). WDYT @06kellyjac ?

06kellyjac: Sounds good. Yeah I've wanted to go through and cut down a lot of the logging for a while now.

06kellyjac:
Thanks for the fixes. Just to double check, this change would force the email tied to the user in git-proxy must also be the email they're using for making their git commits?

I'm trying to think of ways to facilitate cases where either:
* the login email is one you'd keep private & you have a more public email for commits, or
* you have one email for some projects and another email for other projects

For the 2nd usecase I'm less bothered, as with the current username method of tying commits to users you can only input one "git username" anyway.

Regarding alternative emails an "easy" option may be to allow self reporting of emails for now but there wouldn't necessarily be verification (especially to start with).
Or maybe admins can add additional emails for users? :shrug: 

Looking at passport some auth backends might report [multiple emails](https://www.passportjs.org/reference/normalized-profile/) which could be a more trusted source. For AD it looks like just the 1 primary email https://github.com/bhoriuchi/passport-activedirectory/blob/master/src/strategy.js#L90

kriswest:
@06kellyjac 

>Thanks for the fixes. Just to double check, this change would force the email tied to the user in git-proxy must also be the email they're using for making their git commits?

Yes it would. With #950 / #1043 adding support for other SCM host (e.g. GitLab, GitLab dedicated instances and bare git repositories), email is the only reliable field we can use to associate commits with users. It has the advantage that users don't need to set their GitHub id (which is of course irrelevant to other SCM providers) and works with bare git repos. It also mirrors GitHub and EasyCLA's use of email to identify a user.

> Or maybe admins can add additional emails for users? 🤷

This is likely what we'd prefer (certainly not self-reporting as that would be too easy to abuse). Although auto-population from an auth backend that supports multiple emails would be easy enough to support alongside that I imagine.

> Looking at passport some auth backends might report [multiple emails](https://www.passportjs.org/reference/normalized-profile/) which could be a more trusted source. For AD it looks like just the 1 primary email https://github.com/bhoriuchi/passport-activedirectory/blob/master/src/strategy.js#L90

Supporting multiple emails will require a few tweaks: support in the DB (isUserPushAllowed), API and UI for administration.  It is achievable (although I'd suggest its a separate issue / PR after this is adopted).

src/proxy/processors/push-action/parsePush.ts
kriswest: removed the debug level logs!

JamieSlome:
@06kellyjac - I'd like to get this merged and released today 🙌 Do we have anything else we need to clarify on your comments above or can we go ahead with a minor release?

kriswest:
Resolved conflicts again @JamieSlome.

jescalada:
@06kellyjac I agree with making `-rc` releases until we get all the big PRs rounded up. It'll be a lot easier to manage just v1 and v2 for now, and less confusing for users 🙂

06kellyjac:
Sounds like a plan 🚀

Feel free to kick off the discussion, that'd be much appreciated.

kriswest:
Looking closer at getMissingData I can see there is an attempt to validate a user has push permissions, but it is only the author of the last commit. Thats not the same as a check on committer (as per checkUserPushPermissions) nor is it ensuring that the all the authors were allowed (which we only check for 'legality' in checkAuthorEmails, right?, we don't require them all to have push perms).

kriswest:
With getMissingData and the [exception in checkUserPushPermissions](https://github.com/finos/git-proxy/blob/79126ffe25e61b1c9125583338c421f6df973277/src/proxy/processors/push-action/checkUserPushPermission.ts#L10-L13), I suspect I could construct an artificial push with missing commitData, cherry-pick a commit from someone approved to push into the first (or possibly last) slot authored by an approved committer and pass the validation of pushPermissions, without actually having that permission.

If thats the not the case, I'd love to se some comments added explaining.

jescalada:
@kriswest Thanks for letting me know about the `getMissingData` issues! It's meant to intercept certain cases where `commitData` is missing, such as:
- Empty branch pushes
- Fast-forward merges
- New branches from unapproved/unpushed commits

This was the PR description for the fix:

> This PR builds on the previous vulnerability fix https://github.com/jescalada/git-proxy-security-fixes/pull/5 and fixes an error when making a new branch from an unapproved commit.
>
> The flow to reproduce the issue is:
>
>   - Make a commit in branch `a`
>   - Make a new branch `b` from that commit without approving it
>   - Make a new commit in `b`, then approve it
>   - Go back to `a`, and attempt to push this commit to the proxy
>
> Alternatively: You can make multiple commits to branch `a` in 1), then continue the flow. The new code is capable of handling multiple unapproved prior commits.
>
> **This was causing an `undefined` error due to the `commitData` being empty.** Git assumes that the proxy already has the required commit in its history (which is true). **However, we don't actually store this `commitData` anywhere, and the pulled repo gets deleted in clearBareClone.**
>
> My solution was to create an additional action `getMissingData` which activates only when the `commitData` is empty. **It fetches the `commitData` directly from the git log to populate the missing commits.** It also fetches and validates the user data, which relies on the latest commit.
>
> ### Changelog
>
>   - Add getMissingData for fetching `action.commitData` and `action.user` for fast-forward pushes (or any push missing the `commitData` in the pack file)
>   - Extend `checkUserPushPermission` action to export user validation logic
>   - Patch `parsePush` and `checkUserPushPermission` actions to defer `commitData`-reliant checks to `getMissingData` if missing rather than erroring out

kriswest:
np @jescalada, conflicts resolved and ready for one last look!

I've had a go at resolving the conflicts between #973 and #1043 and have some commits lined up to resolve. I agree on the version numbering and can add the version number change to #1043 if desired.

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -36,33 +30,25 @@ async function exec(req: any, action: Action): Promise<Action> {
...
+      throw new Error('No body found in request');
     }
     const [packetLines, packDataOffset] = parsePacketLines(req.body);
     const refUpdates = packetLines.filter((line) => line.includes(BRANCH_PREFIX));
 
     if (refUpdates.length !== 1) {
       step.log('Invalid number of branch updates.');
       step.log(`Expected 1, but got ${refUpdates.length}`);
-      step.setError(
+      throw new Error(
```
jescalada, line 41: Is there a reason (other than simplifying the code somewhat) to use `throw new Error` instead of `step.setError` here and in other spots in `parsePush.ts`?

jescalada (approved):
Took a quick look at the code changes, and everything seems good to go!

src/proxy/processors/push-action/parsePush.ts
jescalada: Either way, it seems the try/catch in `parsePush` handles it, I'm just wondering if the errors will get displayed properly even if we don't set them directly in the `step`.

kriswest: Indeed just simplification. Easier to follow the control flow.

PR #975, opened 2025-04-11
fix: repository listing default query

### LINKED ISSUE #947: Repositories page specifies an incorrect DB query and only works because it is ignored
**Describe the bug**
The repositories page route specifies a default query with a filter that is not correct for repositories (as they don't have the `type` field specified:

https://github.com/finos/git-proxy/blob/5d24d9dbd0ea22ac81b3b4288391cb7aee978c5d/src/service/routes/repo.js#L8-L23

Meanwhile the DB implementations ignore the query passed in, which is the only reason that the page displays any repos (as they all would have been hidden by the filter):

File DB client:
https://github.com/finos/git-proxy/blob/5d24d9dbd0ea22ac81b3b4288391cb7aee978c5d/src/db/file/repo.js#L9-L11

Mongo:
https://github.com/finos/git-proxy/blob/5d24d9dbd0ea22ac81b3b4288391cb7aee978c5d/src/db/mongo/repo.js#L8-L11

Both errors should be fixed. The query parameter should be retained for future use in paginating and filtering results.

**Expected behavior**
- The default query for the repos page should be empty so all results are returned
- The DB client implementations should handle query parameters to support pagination etc. in future.

**Additional context**
I'll raise PR to correct this in the repos page and at least the file client, perhaps also mongo.

kriswest:
resolves #947

The repository page passes an incorrect query by default. This is currently ok as the database clients ignore it, but to fix them we first need to fix the query in this page.

### DIFF
--- a/src/db/file/repo.ts
+++ b/src/db/file/repo.ts
@@ -7,7 +7,7 @@ if (!fs.existsSync('./.data/db')) fs.mkdirSync('./.data/db');
 
 const db = new Datastore({ filename: './.data/db/repos.db', autoload: true });
 
-export const getRepos = async (query: any = {}) => {
+export  const getRepos = async (query: any = {}) => {
   return new Promise<Repo[]>((resolve, reject) => {
     db.find({}, (err: Error, docs: Repo[]) => {
       if (err) {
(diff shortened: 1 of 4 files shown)


kriswest:
Fixed teh cypress test but am waiting on a review (in git proxy), will update monday with working cypress tests

kriswest:
Will get code coverage when #979 is merged

PR #977, opened 2025-04-13
feat(auth): add role mapping for JWT auth claims

jescalada:
This PR builds on #967 and adds the possibility of defining mappings for JWT claims into app roles, such as the admin role. This adds flexibility to the API, allowing automated tools to use admin-only endpoints.

It also adds a test suite for the JWT-related additions in both PRs.

Here's a sample `apiAuthentication` entry in `proxy.config.json` to understand the structure of the mapping definition. In this case, all users with the claim `"name"` set to a value of `"Juan E"` will have their `req.user.admin == true`:

```json
"apiAuthentication": [
    {
      "type": "jwt",
      "enabled": true,
      "jwtConfig": {
        "clientID": "<my-client-id>",
        "authorityURL": "https://accounts.google.com",
        "roleMapping": {
          "admin": {
            "name": "Juan E"
          }
        }
      }
    }
  ]
```

### Changelog:
- Add JWT claim -> role mapping functionality
- Add JWT mappings to config
  - `admin` role set by default
- Add unit tests for `jwtAuthHandler` and `jwtUtils` files


cypress/e2e/login.cy.js
```diff
@@ -19,6 +19,18 @@ describe('Login page', () => {
     cy.get('[data-test="login"]').should('exist');
   });
 
+  it('should redirect to repo list on valid login', () => {
```
dcoric: The new test "should redirect to repo list on valid login" is good, but consider adding a negative test case (e.g., invalid login) to ensure the login flow handles errors gracefully.

proxy.config.json
```diff
@@ -98,6 +109,22 @@
   "contactEmail": "",
   "csrfProtection": true,
   "plugins": [],
+  "apiAuthentication": [
```
dcoric: The new apiAuthentication property is well-defined, but consider adding examples or additional documentation to clarify how this property should be used in practice. This will help other developers understand its purpose and usage.

jescalada: I ended up adding a documentation page for auth in general to make it clear enough. 😃 It doesn't explain how to release a JWT, but I think that anybody who actually needs that feature will probably know how to get one...

JamieSlome:
@jescalada - would this be suitable for a minor bump rather than patch bump?

JamieSlome (approved):
LGTM! 🍰

PR #979, opened 2025-04-15
948 neDB implementation issues

### LINKED ISSUE #948: File-based DB support implementation is not usable out-of-the-box and deviates from the mongo implementation
**Describe the bug**
There appear to be a number of flaws in the file-based DB implementation that impede its use. These include some small deviations from the mongo implementation AND some deviations in behaviotr between @seald/nedb and mongo that should be handled in the implementation.

1. lowercasing of inputs (particularly of usernames and emails) is inconsistently applied and differs from mongo implementation. In places, inputs are lowercased before being sent to the database client and hence apply to both client implementations, however, the mongo clients applies lowercasing on its own in other situations. The database clients should consistently lowercase inputs where appropriate (in both record creation and query) to ensure consistent application. Example of deviation (there are many more cases):
    - Mongo: https://github.com/finos/git-proxy/blob/5d24d9dbd0ea22ac81b3b4288391cb7aee978c5d/src/db/mongo/repo.js#L41-L45 
    - File: https://github.com/finos/git-proxy/blob/5d24d9dbd0ea22ac81b3b4288391cb7aee978c5d/src/db/file/repo.js#L54-L73
 
2. Queries are being ignored in user retrieval, causing all users to be retrieved and push parsing to fail. Query being ignored: https://github.com/finos/git-proxy/blob/5d24d9dbd0ea22ac81b3b4288391cb7aee978c5d/src/db/file/users.js#L78-L81
(note that query is not passed in on last nline)
Push parsing that requires DB filtering of users that is not happening: https://github.com/finos/git-proxy/blob/5d24d9dbd0ea22ac81b3b4288391cb7aee978c5d/src/proxy/processors/push-action/checkUserPushPermission.js#L13-L20
(note exactly one user must be found on line 17 - yet we are not actually filtering users)

3. The implementation does not merge records on update, but replaces them. This causes fields not supplied in the update to be  dropped (e.g. if a user is updated on login via an SSO integration, then the Github user name is dropped as that is manually entered later). 
https://github.com/finos/git-proxy/blob/5d24d9dbd0ea22ac81b3b4288391cb7aee978c5d/src/db/file/users.js#L65-L76
(this looks correct by @seald/nedb does not actually merge records when called like this - it finds the existing record and replaces it with the `user` passed in. This needs rewriting to either use the $set keyword or to manually retrieve the existing doc and merge into before calling `db.update`.

4. neDB implementations should be set to compact their data files periodically, unless they are restarted regularly (compacts on read) - neDB uses an append-only file format so it will create many duplicate lines, See https://github.com/seald/nedb?tab=readme-ov-file#persistence. Add calls to set-up compaction to avoid infinite growth and very redundant data files.

5. No indexes are applied, which will result in performance losses as the dataset grows. Add calls to create indexes supporting common queries - these calls should be made on every start-up. See: https://github.com/seald/nedb?tab=readme-ov-file#indexing for details
**Expected behavior**
The behaviour of the file-based DB support should match that of the mongo implementation and it should work out-of-the-box.

**Additional context**
I'll raise a PR to address these issues in the implementation.

kriswest:
resolves #948

Fixes a number of shortcomings in the file DB implementation, including:
- Merge documents when updating user records instead of replacing (to match behaviour of mongo - was dropping gitAccount when record was updated)
- Apply basic validation to repository records (to match behaviour of mongo)
- Reduce username and email to lowercase consistently in DB clients
- Add indexes and compaction settings (db file format is append only and needs periodic compaction)

Changes were also applied to the mongo DB classes:
- Be more consistent about lowercasing user details (username and email)
- Lowercase repo name consistently
  - this change may be controversial (may need migration)
  - however a number of functions already lowercase the repo name... e.g. `addUserCanPush` and `addUserCanAuthorise`, its just not being applied consistently

Finally, a test was updated to ensure it uses unique usernames and emails as this is necessary after the indexes are added.


kriswest:
This PR has one more push to come that will hopefully complete the code coverage. Should be reviewed and pushed tomorrow

kriswest:
One last commit to push on this PR that should get coverage on /src/db/file up to near 100% - might get pushed after easter now.

kriswest:
@JamieSlome @coopernetes @grovesy this PR should be ready to go now

kriswest:
@JamieSlome I was set-up as per the project config, however the project config doesn't cover packages/git-proxy-cli! I've tweaked the format command so that it does apply and run it. I tried to add prettier to the child project directly but it's not working there, so I extended the parent project's format command to cover it instead.

This adds a bunch of files to the PR with minor changes - waiting on a review then will push

kriswest:
@JamieSlome prettier was run on everything and should be applying to packages/git-proxy-cli. It looks like that was previously formatted with different settings / narrower width. Does it look right to you?

coopernetes (approved):
These are all welcome changes and LGTM on the database side. I'm leaving the UI review to @JamieSlome

sam-holmes2 (approved):
LGTM! Thanks for your contribution :)

JamieSlome (approved):
LGTM! 🍰

Issue #27, opened 2020-09-29
SSH support


adamj-codethink:
**Is your feature request related to a problem? Please describe.**
I may be mistaken, but i am wondering how GitProxy may work if someone uses SSH instead of HTTP when pulling/pushing to a Git repo.

Is this something that is already supported, if not could this be added?

**Describe the solution you'd like**
Support SSH when pulling/pushing to Github.

**Describe alternatives you've considered**
Of course HTTP/S will work for now.

**Additional context**
I did check the README and previous issues, but i could not find anything related to SSH, so this is just an issue to investigate.

grovesy:
[+ @maoo] 
We had some discussions around this - right now, it doesn't support SSH, certainly on the list to do - need to look into how we can use the proxy as an SSH tunnel (certainly doable!)

maoo:
The ssh2 library seems interesting, especially the server example posted on https://github.com/mscdex/ssh2#server-examples

Also found https://www.npmjs.com/package/tunnel-ssh , which is based on `ssh2`, though I could not find a way to intercept the connection payload.

The challenge, as far as I can see, is to _connect_ SSH incoming connections with body parsing and routing currently developed for the http endpoint:
```
const proxyApp = require('express')();
...
proxyApp.use(bodyParser.raw(options));
proxyApp.use('/', routes);
```
See https://github.com/finos/git-proxy/blob/main/src/proxy/index.js#L16

JamieSlome:
@maoo - can we close this off now?

maoo:
> @maoo - can we close this off now?

Don't think that SSH is supported by Git Proxy right now, unless I missed some developments. If that's the case, I'd keep this issue open.

PR #987, opened 2025-04-28
feat: adds ssh support for git operations

dcoric:
This PR implements SSH support for Git operations, resolving #27. The implementation allows users to perform Git operations over SSH protocol, providing an alternative to HTTPS for repository access.

## Changes

1. Added SSH server implementation using `ssh2` library
2. Implemented two authentication methods:
   - Public key authentication
   - Password authentication
3. Added SSH configuration options in `config.schema.json`
4. Created comprehensive SSH documentation in `docs/SSH.md`
5. Added SSH server tests in `test/ssh/sshServer.test.js`

## Configuration

SSH support can be configured in the main configuration file:

```json
{
  "ssh": {
    "enabled": true,
    "port": 2222,
    "hostKey": {
      "privateKeyPath": "./.ssh/host_key",
      "publicKeyPath": "./.ssh/host_key.pub"
    }
  }
}
```

## Usage

Users can connect using standard Git SSH commands. The command format depends on the configured port:

- If port is set to 22 (default SSH port):
  ```bash
  git clone git@your-proxy:username/repo.git
  ```

- If using a custom port (e.g., 2222):
  ```bash
  git clone ssh://git@your-proxy:2222/username/repo.git
  ```

## Testing

Added comprehensive tests for SSH functionality:
- Server initialization
- Authentication methods
- Connection handling
- Git protocol support

## Documentation

Added detailed documentation in `docs/SSH.md` covering:
- Configuration options
- Authentication methods
- Security features
- Usage instructions
- Troubleshooting guide

## Future Improvements

1. Move SSH configuration options (keep alive, timeouts, and other params) to config file
2. Enhance actions for SSH functionality
3. Increase test coverage
4. Improved error reporting
5. Additional security features

This implementation provides a secure and reliable way to perform Git operations over SSH, giving users more flexibility in how they interact with their repositories.

Below is a summary of the most important changes:

### SSH Feature Implementation:
* Added a new `ssh` configuration section in `config.schema.json` and `proxy.config.json` to enable SSH support, configure the port, and specify host key paths. (`[[1]](diffhunk://#diff-be1695b1e63a508d59982601f9e1fb7f58247deecb1e427adb77bcad758ae5e5R82-R111)`, `[[2]](diffhunk://#diff-c465aafa0fe603e2d28b017938f55e5ce3253aac7aa303efeabfc06a4ad52d5fR105-R112)`)
* Created `docs/SSH.md` to provide comprehensive documentation for configuring, using, and troubleshooting the SSH feature. (`[docs/SSH.mdR1-R165](diffhunk://#diff-1d0301881738da7c699d7634669c5156c345a5094d6c012cfa7254cbdf15cbd2R1-R165)`)
* Integrated SSH key management into the CLI, including commands to add and remove SSH keys (`addSSHKey` function and `ssh-key` command). (`[[1]](diffhunk://#diff-ee51eb1c2264303569f2a8fa9f5bb2de1b6b51eed24eed943d079f341f4139b0R310-R363)`, `[[2]](diffhunk://#diff-ee51eb1c2264303569f2a8fa9f5bb2de1b6b51eed24eed943d079f341f4139b0R494-R524)`, `[[3]](diffhunk://#diff-c3c551630c7afb35431840ab36bd8cc771deb1fa385b73804eaf8da4b8db7e0cR1-R122)`)

### Database Enhancements:
* Updated the user schema to include SSH public keys and added functions for managing SSH keys (`addPublicKey`, `removePublicKey`, `findUserBySSHKey`). (`[[1]](diffhunk://#diff-d90c8ba033b20fb64a27fb6f16a94afa5ebc2f25bb7af6ec42cd260dcbf46710R43-R46)`, `[[2]](diffhunk://#diff-d90c8ba033b20fb64a27fb6f16a94afa5ebc2f25bb7af6ec42cd260dcbf46710R98-R155)`, `[[3]](diffhunk://#diff-c7d2c320d1f7fc0aad36b8bf616a7289ca4dcf00e6e057ba8a1d354fb25b74c0R84-R86)`)

### Configuration and Code Updates:
* Introduced new methods in `src/config/index.ts` to retrieve SSH-related settings (`getSSHConfig`, `getSSHProxyUrl`). (`[[1]](diffhunk://#diff-198a1431d7c5e26ea78bc6e54b34b54e6f8ab342dd48ea11013877e8466a87c5R33)`, `[[2]](diffhunk://#diff-198a1431d7c5e26ea78bc6e54b34b54e6f8ab342dd48ea11013877e8466a87c5R48-R58)`)
* Updated `package.json` to include the `ssh2` library and its TypeScript types for SSH server implementation. (`[[1]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R75)`, `[[2]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R90)`)

### Documentation Enhancements:
* Updated `README.md` to include a link to the new SSH documentation. (`[README.mdR88-R95](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R88-R95)`)

These changes collectively enable the SSH feature, providing an alternative to HTTPS for secure Git operations while enhancing user control and security.


dcoric:
@JamieSlome @coopernetes, please take a look when you get a chance

src/proxy/ssh/server.js
```diff
@@ -0,0 +1,683 @@
...
+          // Create SSH connection to GitHub using the Client approach
+          const { Client } = require('ssh2');
+          const remoteGitSsh = new Client();
+
+          console.log('[SSH] Creating SSH connection to remote');
+
+          // Get remote host from config
+          const remoteUrl = new URL(config.getProxyUrl());
+
+          // TODO: Connection options could go to config
```
jescalada, line 231: Bringing attention to this `TODO`. Is there a reason why the default port is `22` in some places, but `2222` in the sample `proxy.config.json`? 🤔

src/proxy/ssh/server.js
```diff
@@ -0,0 +1,683 @@
...
+              const proxyGitSsh = new Client();
+
+              // Set up connection options with proxy key
+              const proxyConnectionOptions = {
+                ...connectionOptions,
+                privateKey: require('fs').readFileSync(
+                  config.getSSHConfig().hostKey.privateKeyPath,
+                ),
+                // Ensure these settings are explicitly set for the proxy connection
+                windowSize: 1024 * 1024, // 1MB window size
```
jescalada, line 475: Might be best to extract these (and the repeated ones from earlier in the file) into constants so they can be reused.

src/service/routes/users.js
```diff
@@ -29,4 +36,68 @@ router.get('/:id', async (req, res) => {
   res.send(user);
 });
 
+// Add SSH public key
+router.post('/:username/ssh-keys', async (req, res) => {
+  if (!req.user) {
+    res.status(401).json({ error: 'Authentication required' });
```
jescalada, line 42: I'm wondering if this could be `'Login required'` to make it a bit more clear that the user should login first. Since it will mainly be used as an API, the flow may not be obvious.

Feel free to ignore this if the flow (including login) is documented!

src/service/routes/users.js
```diff
@@ -29,4 +36,68 @@ router.get('/:id', async (req, res) => {
...
+  }
+
+  const { publicKey } = req.body;
+  if (!publicKey) {
+    res.status(400).json({ error: 'Public key is required' });
+    return;
+  }
+
+  // Strip the comment from the key (everything after the last space)
+  const keyWithoutComment = publicKey.split(' ').slice(0, 2).join(' ');
```
jescalada, line 61: Should we add some validation here to make sure that the key is in a valid format? Or would this be more appropriate for the db functions?

Right now, invalid formatting would get caught in the try/catch below, but the user-facing error wouldn't be descriptive.

jescalada (changes requested):
Just a preliminary review of the code changes. I'll run it and test more thoroughly afterwards!

src/proxy/ssh/server.js
```diff
@@ -0,0 +1,683 @@
...
+    // Handle keepalive requests
+    client.on('global request', (accept, reject, info) => {
+      console.log('[SSH] Global request:', info);
+      if (info.type === 'keepalive@openssh.com') {
+        console.log('[SSH] Accepting keepalive request');
+        // Always accept keepalive requests to prevent connection drops
+        accept();
+      } else {
+        console.log('[SSH] Rejecting unknown global request:', info.type);
+        reject();
```
jescalada: Would it be better if we sent the error string as a `reject` argument? Might be better for more granular testing.

test/ssh/sshServer.test.js
```diff
@@ -0,0 +1,341 @@
...
+      server.handleClient(mockClient);
+      expect(mockClient.on.calledWith('error')).to.be.true;
+      expect(mockClient.on.calledWith('end')).to.be.true;
+      expect(mockClient.on.calledWith('close')).to.be.true;
+      expect(mockClient.on.calledWith('global request')).to.be.true;
+      expect(mockClient.on.calledWith('ready')).to.be.true;
+      expect(mockClient.on.calledWith('authentication')).to.be.true;
+    });
+
+    describe('authentication', () => {
```
jescalada: Would be great to add extra tests for authentication failure cases. These are a few passing sample tests I wrote:

```suggestion
    describe('authentication', () => {
      it('should reject invalid public keys', async () => {
        const mockCtx = {
          method: 'publickey',
          key: {
            algo: 'ssh-rsa',
            data: Buffer.from('invalid-key-data'),
            comment: 'invalid-key',
          },
          accept: sinon.stub(),
          reject: sinon.stub(),
        };

        mockDb.findUserBySSHKey.resolves(null);

        server.handleClient(mockClient);
        const authHandler = mockClient.on.withArgs('authentication').firstCall.args[1];
        await authHandler(mockCtx);

        expect(mockCtx.reject.calledOnce).to.be.true;
        expect(mockClient.username).to.be.null;
        expect(mockClient.userPrivateKey).to.be.null;
      });
      
      it('should reject invalid passwords', async () => {
        const mockCtx = {
          method: 'password',
          username: 'test-user',
          password: 'invalid-password',
          accept: sinon.stub(),
          reject: sinon.stub(),
        };

        mockDb.findUser.resolves({
          username: 'test-user',
          password: '$2a$10$mockHash',
        });

        const bcrypt = require('bcryptjs');
        sinon.stub(bcrypt, 'compare').resolves(false);

        server.handleClient(mockClient);
        const authHandler = mockClient.on.withArgs('authentication').firstCall.args[1];
        await authHandler(mockCtx);

        expect(mockDb.findUser.calledWith('test-user')).to.be.true;
        expect(bcrypt.compare.calledWith('invalid-password', '$2a$10$mockHash')).to.be.true;
        expect(mockCtx.reject.calledOnce).to.be.true;
      });

      it('should handle a missing private key', async () => {
        const mockCtx = {
          method: 'publickey',
          key: null,
          accept: sinon.stub(),
          reject: sinon.stub(),
        };

        mockDb.findUserBySSHKey.resolves({ username: 'test-user' });

        server.handleClient(mockClient);
        const authHandler = mockClient.on.withArgs('authentication').firstCall.args[1];
        await authHandler(mockCtx);

        expect(mockCtx.reject.calledOnce).to.be.true;
        expect(mockClient.username).to.be.null;
        expect(mockClient.userPrivateKey).to.be.null;
      });
```

jescalada: Extra authentication test case ideas:

- Malformed SSH keys
  - Currently, malformed keys go through the `findUserBySSHKey` function which logs `[SSH] No user found with this SSH key` instead of throwing error
- Invalid auth method (neither `publickey` nor `password`)
- Nonexistent user

test/ssh/sshServer.test.js
```diff
@@ -0,0 +1,341 @@
...
+      mockAccept = sinon.stub().returns(mockSession);
+      mockReject = sinon.stub();
+    });
+
+    it('should handle git-upload-pack command', async () => {
+      const mockInfo = {
+        command: "git-upload-pack 'test/repo'",
+      };
+
+      mockChain.executeChain.resolves({
```
jescalada: Would be great to include some cases with `error: true`/`blocked: true`:

### Sample

I wrote a sample test for this, but it seems to be failing:

```js
    it('should handle git-upload-pack command with error in chain', async () => {
      const mockInfo = {
        command: "git-upload-pack 'test/repo'",
      };
      
      mockChain.executeChain.resolves({
        error: true,
        blocked: true,
      });

      server.handleSession(mockAccept, mockReject);
      const execHandler = mockSession.on.withArgs('exec').firstCall.args[1];
      await execHandler(mockAccept, mockReject, mockInfo);

      expect(mockChain.executeChain.calledWith({
        method: 'GET',
        originalUrl: " 'test/repo",
        isSSH: true,
      })).to.be.true;

      expect(mockStream.end.calledOnce).to.be.true;
      expect(mockStream.exit.calledOnce).to.be.true;
      expect(mockStream.exit.calledWith(1)).to.be.true;
    });
```

#### Error:
```bash

  2) SSHServer
       handleSession
         should handle git-upload-pack command with invalid repo:
     TypeError: stream.write is not a function
      at SSHServer.<anonymous> (src/proxy/ssh/server.js:664:18)
      at fulfilled (src/proxy/ssh/server.js:5:58)
      at processTicksAndRejections (node:internal/process/task_queues:105:5)
```

test/ssh/sshServer.test.js
```diff
@@ -0,0 +1,341 @@
...
+            'user-agent': 'git/2.0.0',
+            'content-type': undefined,
+          },
+        }),
+      ).to.be.true;
+    });
+
+    it('should handle git-receive-pack command', async () => {
+      const mockInfo = {
+        command: "git-receive-pack 'test/repo'",
```
jescalada: Other possible test cases for `handleSession`:

- Invalid repo handling
  - Currently gets stuck on `[GitHub SSH Debug] Client: Trying github.com on port 22 ...` without throwing error. `executeChain` doesn't seem to get called
- Edge case tests:
  - Empty repository name, whitespace inputs
  - SSH injection (example: `command: "git-upload-pack 'repo'; rm -rf /; #",`)

jescalada (commented):
Review (Part 3): Just took a look at the unit tests. It would be great to cover some more edge cases and get more granular testing by passing error messages through the `ctx.reject` (if possible).

If you'd like, I can help out with the documentation refactor and unit test coverage 🙂

src/security/SSHKeyManager.ts
```diff
@@ -0,0 +1,134 @@
...
+   * Securely encrypt an SSH private key for temporary storage
+   * @param {Buffer | string} privateKey The SSH private key to encrypt
+   * @return {object} Object containing encrypted key and expiry time
+   */
+  static encryptSSHKey(privateKey: Buffer | string): {
+    encryptedKey: string;
+    expiryTime: Date;
+  } {
+    const keyBuffer = Buffer.isBuffer(privateKey) ? privateKey : Buffer.from(privateKey);
+    const encryptionKey = this.getEncryptionKey();
```
dgl, line 43: `getEncryptionKey` says it is for development use (if the env variable isn't set), but I don't see a code path calling `generateEncryptionKey` which says it is for production. I'd also consider renaming the methods to be more clear that one is development and production, rather than get vs generate which isn't that clear...

src/proxy/processors/push-action/pullRemote.ts
```diff
@@ -2,9 +2,163 @@ import { Action, Step } from '../../actions';
...
+  return {
+    username: credentials.slice(0, separatorIndex),
+    password: credentials.slice(separatorIndex + 1),
+  };
+};
+
+const buildSSHCloneUrl = (remoteUrl: string): string => {
+  const parsed = new URL(remoteUrl);
+  const repoPath = parsed.pathname.replace(/^\//, '');
+  return `git@${parsed.hostname}:${repoPath}`;
```
dgl, line 50: I would point out in the docs that the username is currently hardcoded to git and/or add a todo (this works fine for github and gitlab I think, but may not work on other services).

src/proxy/processors/push-action/pullRemote.ts
```diff
@@ -2,9 +2,163 @@ import { Action, Step } from '../../actions';
...
+  }
+
+  const keyBuffer = Buffer.isBuffer(privateKey) ? privateKey : Buffer.from(privateKey);
+  const tempDir = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'git-proxy-ssh-clone-'));
+  const keyPath = path.join(tempDir, 'id_rsa');
+
+  await fs.promises.writeFile(keyPath, keyBuffer, { mode: 0o600 });
+
+  const originalGitSSH = process.env.GIT_SSH_COMMAND;
+  process.env.GIT_SSH_COMMAND = `ssh -i ${keyPath} -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null`;
```
dgl, line 86: I'd point out in the docs that the SSH TOFU check is skipped. Better would be generating a known hosts file  somehow (that way the trust in the TLS certificate is used to trust the SSH keys):

For github in curl + jq:
```
curl -sL -H "Accept: application/vnd.github+json" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  https://api.github.com/meta | jq -r '.ssh_keys[]'
```
(From https://docs.github.com/en/rest/meta/meta?apiVersion=2022-11-28#get-github-meta-information)

Could maybe work to have a workflow that generates that as part of a release and ships it (with say github.com and gitlab.com at least).

docs/SSH_KEY_RETENTION.md
```diff
@@ -0,0 +1,199 @@
+# SSH Key Retention for Git Proxy
```
jescalada: Nit: Replace all instances of "Git Proxy" with "GitProxy" without a space for consistency 👍🏼

docs/SSH_KEY_RETENTION.md
```diff
@@ -0,0 +1,199 @@
...
+## Configuration
+
+### Environment Variables
+
+- `SSH_KEY_ENCRYPTION_KEY`: 32-byte hex string for SSH key encryption
+- If not provided, keys are derived from the SSH host key
+
+### SSH Configuration
+
+Enable SSH support in `proxy.config.json`:
```
jescalada, line 90: Wonder if we could make this setup section a bit more "visible"? It'd be great if there was a more user-friendly guide on how to set up the SSH configuration (although this might have been done in the SSH UI PR?).

However, I think this would be best put in a separate file, such as `SETTING_UP_SSH.md` to make it really obvious for busy developers and AI tools scaninng through the docs 😃

docs/SSH_KEY_RETENTION.md
```diff
@@ -0,0 +1,199 @@
...
+- SSH keys are automatically rotated every 24 hours
+- Manual cleanup can be triggered via `SSHKeyManager.cleanupExpiredKeys()`
+
+### Memory Security
+
+- Private keys are stored in Buffer objects that are securely cleared
+- Temporary files are created with restrictive permissions (0600)
+- All temporary files are automatically cleaned up
+
+## API Usage
```
jescalada, line 123: Is there a reason to have this as part of the docs? Perhaps for using the SSH API in custom actions/plugins? Otherwise I feel like we need to add a good explanation here so it's obvious that this doc is for explaining the SSH rationale/implementation rather than the usage.

docs/SSH_KEY_RETENTION.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Future Enhancements
+
+1. **SSH Agent Forwarding**: Implement true SSH agent forwarding instead of key storage
+2. **Key Derivation**: Support for different key types (Ed25519, ECDSA, etc.)
+3. **Audit Logging**: Enhanced audit trail for SSH key usage
+4. **Key Rotation**: Automatic key rotation based on push frequency
+5. **Integration**: Integration with external SSH key management systems
+
+## Troubleshooting
```
jescalada, line 176: Might be good to expand on these and provide more details! I'll add some suggestions for it when going through the setup process again.

packages/git-proxy-cli/index.js
```diff
@@ -176,8 +175,7 @@ async function authoriseGitPush(id) {
     if (error.response) {
       switch (error.response.status) {
         case 401:
-          errorMessage =
-            'Error: Authorise: Authentication required (401): ' + error?.response?.data?.message;
+          errorMessage = 'Error: Authorise: Authentication required';
```
jescalada: Any specific reason to remove the error response content? I feel like keeping that would be good especially for the 401 errors which can happen due to many different reasons (some of which are easier than others for the user to fix).

packages/git-proxy-cli/index.js
```diff
@@ -224,8 +222,7 @@ async function rejectGitPush(id) {
     if (error.response) {
       switch (error.response.status) {
         case 401:
-          errorMessage =
-            'Error: Reject: Authentication required (401): ' + error?.response?.data?.message;
+          errorMessage = 'Error: Reject: Authentication required';
```
jescalada: Same as above, and on the `case 401` line below - check if we ought to keep the error response message or not 👍🏼

src/cli/ssh-key.ts
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env node
+
+import * as fs from 'fs';
+import * as path from 'path';
+import axios from 'axios';
+
+const API_BASE_URL = process.env.GIT_PROXY_API_URL || 'http://localhost:3000';
```
jescalada, line 7: Wondering if this could be obtained from `ServerConfig` instead? There's a few relevant variables in `/src/config/env.ts`. Otherwise there's the `API_BASE` in `/src/ui/apibase.ts` if that's the one we need.

src/cli/ssh-key.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    // Read the cookies
+    const cookies = JSON.parse(fs.readFileSync(GIT_PROXY_COOKIE_FILE, 'utf8'));
+
+    // Read the public key file
+    const publicKey = fs.readFileSync(keyPath, 'utf8').trim();
+    console.log('Read public key:', publicKey);
+
+    // Validate the key format
+    if (!publicKey.startsWith('ssh-')) {
```
jescalada: Wondering if we should have a more robust check here for the SSH key format. 🤔 

Does anything unusual happen if the key were to be set to a private key or something that's just invalid?

src/cli/ssh-key.ts
```diff
@@ -0,0 +1,140 @@
...
+async function addSSHKey(username: string, keyPath: string): Promise<void> {
+  try {
+    // Check for authentication
+    if (!fs.existsSync(GIT_PROXY_COOKIE_FILE)) {
+      console.error('Error: Authentication required. Please run "yarn cli login" first.');
+      process.exit(1);
+    }
+
+    // Read the cookies
+    const cookies = JSON.parse(fs.readFileSync(GIT_PROXY_COOKIE_FILE, 'utf8'));
```
jescalada, line 35: This could be extracted to `getCookies()` and save us a few lines of code in the various functions 🙂

src/cli/ssh-key.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    // Read the public key file
+    const publicKey = fs.readFileSync(keyPath, 'utf8').trim();
```
jescalada, line 38: Reading + validation could be extracted into their own function as well

src/cli/ssh-key.ts
```diff
@@ -0,0 +1,140 @@
...
+  console.log(`
+Usage:
+  Add SSH key:    npx tsx src/cli/ssh-key.ts add <username> <path-to-public-key>
+  Remove SSH key: npx tsx src/cli/ssh-key.ts remove <username> <path-to-public-key>
+  `);
+  process.exit(1);
+}
+
+if (command === 'add') {
+  addSSHKey(username, keyPath);
```
jescalada, line 127: Should these be awaited? Might work anyways, but it'd be good to show that they are async in case someone wants to run a cleanup function afterwards.

src/config/index.ts
```diff
@@ -285,6 +296,47 @@ export const getRateLimit = () => {
   return config.rateLimit;
 };
 
+export const getMaxPackSizeBytes = (): number => {
+  const config = loadFullConfiguration();
+  const configuredValue = config.limits?.maxPackSizeBytes;
+  const fallback = 1024 * 1024 * 1024; // 1 GiB default
```
jescalada: What are the implications of having a 1 GB default max pack size? Would this somehow prevent larger operations from going through SSH?

src/config/index.ts
```diff
@@ -285,6 +296,47 @@ export const getRateLimit = () => {
...
+
+  if (
+    typeof configuredValue === 'number' &&
```
jescalada, line 306: This validation should happen automatically if `maxPackSizeBytes` is added to the `config.schema.json` (and then we import the generated types from `src/config/generated/config.ts`. We can still use fallbacks in the flow by omitting the `"required": []` section in the schema entry.

src/config/index.ts
```diff
@@ -285,6 +296,47 @@ export const getRateLimit = () => {
...
+        return userConfig.ssh || { enabled: false };
+      } catch (e) {
+        console.error('Error loading SSH config:', e);
+      }
+    }
+    return { enabled: false };
+  }
+};
+
+export const getSSHProxyUrl = (): string | undefined => {
```
jescalada, line 336: Is this function being used? 🤔

src/db/file/users.ts
```diff
@@ -172,3 +175,67 @@ export const getUsers = (query: any = {}): Promise<User[]> => {
...
+        user.publicKeys = user.publicKeys.filter((key) => key !== publicKey);
+        updateUser(user)
+          .then(() => resolve())
+          .catch(reject);
+      })
+      .catch(reject);
+  });
+};
+
+export const findUserBySSHKey = (sshKey: string): Promise<User | null> => {
```
jescalada, line 240: Wondering how this handles searching with duplicates - what would happen if for some reason two users shared the same public key (perhaps if two users were logging in from the same device)?

src/db/mongo/users.ts
```diff
@@ -55,9 +58,34 @@ export const updateUser = async (user: User): Promise<void> => {
...
+
+export const removePublicKey = async (username: string, publicKey: string): Promise<void> => {
+  const collection = await connect(collectionName);
+  await collection.updateOne(
+    { username: username.toLowerCase() },
+    { $pull: { publicKeys: publicKey } },
+  );
+};
+
+export const findUserBySSHKey = async function (sshKey: string): Promise<User | null> {
```
jescalada, line 98: Same comment as above, does this correctly handle two users with the same key, or other duplicate key scenarios?

src/proxy/processors/pre-processor/parseAction.ts
```diff
@@ -6,6 +6,27 @@ const exec = async (req: {
   originalUrl: string;
   method: string;
   headers: Record<string, string>;
+  protocol?: 'https' | 'ssh';
```
jescalada, line 9: There might be a way to refactor this to use `Action` instead, since most of the attributes are shared.

src/proxy/processors/push-action/captureSSHKey.ts
```diff
@@ -0,0 +1,99 @@
...
+    const privateKeySource =
+      sshKeyContext?.privateKey ?? sshKeyContext?.keyData ?? action.sshUser.sshKeyInfo.keyData;
+
+    if (!privateKeySource) {
+      step.log('No SSH private key available for capture');
+      action.addStep(step);
+      return action;
+    }
+
+    const privateKeyBuffer = Buffer.isBuffer(privateKeySource)
```
jescalada: It seems these are executing `Buffer.from(privateKeySource)` and such, regardless of whether it's a buffer or not

src/proxy/processors/push-action/captureSSHKey.ts
```diff
@@ -0,0 +1,99 @@
...
+      );
+    }
+
+    const encrypted = SSHKeyManager.encryptSSHKey(privateKeyBuffer);
+    action.encryptedSSHKey = encrypted.encryptedKey;
+    action.sshKeyExpiry = encrypted.expiryTime;
+    step.log('SSH key information stored for approval process');
+    step.setContent(`SSH key retained until ${encrypted.expiryTime.toISOString()}`);
+
+    privateKeyBuffer.fill(0);
```
jescalada: Should we be clearing the `publicKeyBuffer` memory as well?

src/proxy/processors/push-action/captureSSHKey.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    const addedToAgent = SSHKeyForwardingService.addSSHKeyForPush(
+      action.id,
+      Buffer.from(privateKeyBuffer),
+      publicKeyBuffer,
+      action.sshUser.email ?? action.sshUser.username,
+    );
+
+    if (!addedToAgent) {
+      console.warn(
```
jescalada: Wonder if it's fine to continue execution here if SSH key caching fails 🤔

src/proxy/processors/push-action/captureSSHKey.ts
```diff
@@ -0,0 +1,99 @@
...
+    // Check if we have the necessary SSH key information
+    if (!action.sshUser.sshKeyInfo) {
+      step.log('No SSH key information available for capture');
+      action.addStep(step);
+      return action;
+    }
+
+    const authContext = req?.authContext ?? {};
+    const sshKeyContext = authContext?.sshKey;
+    const privateKeySource =
```
jescalada: Is is necessary to have multiple different sources for the private key? If necessary, it'd be nice to have this fallback chain documented somewhere so it's easier for users to configure.

src/proxy/processors/push-action/captureSSHKey.ts
```diff
@@ -0,0 +1,99 @@
...
+    // This is a placeholder for the key capture mechanism
+
+    action.addStep(step);
+    return action;
+  } catch (error: unknown) {
+    const errorMessage = error instanceof Error ? error.message : 'Unknown error';
+    step.setError(`Failed to capture SSH key: ${errorMessage}`);
+    action.addStep(step);
+    return action;
+  }
```
jescalada, line 87: Might be worth doing the buffer cleanup at the end:

```ts
} finally {
  if (privateKeyBuffer) {
    privateKeyBuffer.fill(0);
  }
  if (publicKeyBuffer) {
    publicKeyBuffer.fill(0);
  }
}
```

src/proxy/processors/push-action/pullRemote.ts
jescalada, line 1: We should think about whether it's fine to keep SSH logic in `pullRemote` or move it somewhere else. I feel like this file is going to get way too complex after merging in @fabiovincenzi's push/pull optimizations (#1246).

Any thoughts? 🤔

jescalada: We could try and simplify things by extracting all the util functions perhaps

src/proxy/processors/push-action/pullRemote.ts
```diff
@@ -2,9 +2,163 @@ import { Action, Step } from '../../actions';
...
+
+  await git.clone(cloneOptions);
+};
+
+const cloneWithSSHKey = async (action: Action, privateKey: Buffer): Promise<void> => {
+  if (!privateKey || privateKey.length === 0) {
+    throw new Error('SSH private key is empty');
+  }
+
+  const keyBuffer = Buffer.isBuffer(privateKey) ? privateKey : Buffer.from(privateKey);
```
jescalada, line 79: The key in this buffer isn't being cleared (not sure if intentional since we were clearing the public/private key buffers in `captureSSHKey.ts`.

jescalada: Also, the conversion seems redundant - I think Buffer.from(privateKey) would handle both cases?

src/proxy/processors/push-action/pullRemote.ts
```diff
@@ -17,31 +171,34 @@ const exec = async (req: any, action: Action): Promise<Action> => {
...
-    await git.clone({
-      fs,
-      http: gitHttpClient,
-      url: action.url,
-      dir: `${action.proxyGitPath}/${action.repoName}`,
-      onAuth: () => ({ username, password }),
-      singleBranch: true,
-      depth: 1,
-    });
+    ensureDirectory(action.proxyGitPath);
```
jescalada: It seems we're calling `!fs.existsSync(action.proxyGitPath)` twice in a row since `ensureDirectory` calls it internally.

src/proxy/processors/push-action/pullRemote.ts
```diff
@@ -2,9 +2,163 @@ import { Action, Step } from '../../actions';
...
+
+const buildSSHCloneUrl = (remoteUrl: string): string => {
+  const parsed = new URL(remoteUrl);
+  const repoPath = parsed.pathname.replace(/^\//, '');
+  return `git@${parsed.hostname}:${repoPath}`;
+};
+
+const cleanupTempDir = async (tempDir: string) => {
+  try {
+    await fs.promises.rm(tempDir, { recursive: true, force: true });
```
jescalada: I'm wondering about the use of `fs.promises` along with `fs.existsSync`/`fs.mkdirSync`. Can/should these be consistent?

src/proxy/processors/push-action/pullRemote.ts
```diff
@@ -2,9 +2,163 @@ import { Action, Step } from '../../actions';
...
+      command: `git clone ${action.url}`,
+      strategy: 'ssh-service-token',
+    };
+  }
+
+  step.log('No SSH clone credentials available; attempting anonymous HTTPS clone');
+  try {
+    await cloneWithHTTPS(action, null);
+  } catch (error) {
+    const err =
```
jescalada: Could the error message construction be simplified a bit?

src/proxy/ssh/server.ts
```diff
@@ -0,0 +1,1170 @@
...
+
+interface SSHUser {
+  username: string;
+  password?: string | null;
+  publicKeys?: string[];
+  email?: string;
+  gitAccount?: string;
+}
+
+interface AuthenticatedUser {
```
jescalada, line 21: Could be replaced with `Partial<User>`

src/proxy/ssh/server.ts
```diff
@@ -0,0 +1,1170 @@
...
+        ctx.method,
+        'for user:',
+        ctx.username,
+      );
+
+      if (ctx.method === 'publickey') {
+        // Handle public key authentication
+        const keyString = `${ctx.key.algo} ${ctx.key.data.toString('base64')}`;
+
+        (db as any)
```
jescalada, line 249: Is casting necessary here? I thought that `db` is exporting `findUserBySSHKey` 🤔

src/proxy/ssh/server.ts
```diff
@@ -0,0 +1,1170 @@
...
+      // Clean up keepalive timer
+      const keepaliveTimer = this.keepaliveTimers.get(client);
+      if (keepaliveTimer) {
+        clearInterval(keepaliveTimer);
+        this.keepaliveTimers.delete(client);
+      }
+    });
+
+    // Handle keepalive requests
+    (client as any).on('global request', (accept: () => void, reject: () => void, info: any) => {
```
jescalada, line 224: Why do we sometimes cast `client` to `any` before calling `client.on` here? The `client.on('close', ...)` bit above seems to work without casting.

src/proxy/ssh/server.ts
```diff
@@ -0,0 +1,1170 @@
...
+  }
+
+  private async forwardPackDataToRemote(
+    command: string,
+    stream: ssh2.ServerChannel,
+    client: ClientWithUser,
+    packData: Buffer | null,
+    action?: Action,
+  ): Promise<void> {
+    return new Promise((resolve, reject) => {
```
jescalada, line 730: Should this promise automatically reject after a set amount of time has passed? I'm not sure if it can handle transfer errors without hanging.

src/proxy/ssh/server.ts
```diff
@@ -0,0 +1,1170 @@
...
+      );
+      const errorMessage = chainError instanceof Error ? chainError.message : String(chainError);
+      stream.stderr.write(`Access denied: ${errorMessage}\n`);
+      stream.exit(1);
+      stream.end();
+      return;
+    }
+  }
+
+  private async forwardPackDataToRemote(
```
jescalada, line 723: It seems this function is similar to `connectToRemoteGitServer`, maybe we could extract the connection logic into its own util function to reduce the number of lines?

src/security/SSHAgent.ts
```diff
@@ -0,0 +1,219 @@
...
+  signData(pushId: string, data: Buffer): Buffer | null {
+    const privateKey = this.getPrivateKey(pushId);
+    if (!privateKey) {
+      return null;
+    }
+
+    try {
+      // Create a sign object - this is a simplified version
+      // In practice, you'd need to handle different key types (RSA, Ed25519, etc.)
+      const sign = crypto.createSign('SHA256');
```
jescalada, line 200: Commenting for visibility - do we want to support different key types at this moment? If so, how can we make this a bit more flexible?

src/security/SSHAgent.ts
```diff
@@ -0,0 +1,219 @@
+import { EventEmitter } from 'events';
+import * as crypto from 'crypto';
+
+/**
+ * SSH Agent for handling user SSH keys securely during the approval process
+ * This class manages SSH key forwarding without directly exposing private keys
+ */
+export class SSHAgent extends EventEmitter {
+  private keyStore: Map<
```
jescalada, line 9: Should we have some kind of limit on the number of keys that can be stored?

src/security/SSHAgent.ts
```diff
@@ -0,0 +1,219 @@
...
+      return false;
+    }
+  }
+
+  /**
+   * Remove an SSH key from the agent
+   * @param {string} pushId The push ID associated with the key
+   * @return {boolean} True if key was removed
+   */
+  removeKey(pushId: string): boolean {
```
jescalada, line 83: Might want to wrap this in a try/catch in case `fill`/`delete` fail.

src/security/SSHKeyManager.ts
```diff
@@ -0,0 +1,134 @@
...
+  static generateEncryptionKey(): string {
+    return crypto.randomBytes(32).toString('hex');
+  }
+
+  /**
+   * Clean up expired SSH keys from the database
+   * @return {Promise<void>} Promise that resolves when cleanup is complete
+   */
+  static async cleanupExpiredKeys(): Promise<void> {
+    const db = require('../db');
```
jescalada, line 119: Can be replaced with top-level import

src/service/SSHKeyForwardingService.ts
```diff
@@ -0,0 +1,216 @@
...
+        // Write the private key to a temporary file
+        await fs.promises.writeFile(keyPath, privateKey, { mode: 0o600 });
+
+        // Set up git with the temporary SSH key
+        const originalGitSSH = process.env.GIT_SSH_COMMAND;
+        process.env.GIT_SSH_COMMAND = `ssh -i ${keyPath} -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null`;
+
+        // Execute the git push
+        const gitRepo = simpleGit(push.proxyGitPath);
+        await gitRepo.push('origin', push.branch);
```
jescalada, line 102: Should the git operations have a timeout in case of disconnections? Might want to test if that plays nice with larger repos too.

src/service/SSHKeyForwardingService.ts
```diff
@@ -0,0 +1,216 @@
...
+    }
+  }
+
+  /**
+   * Execute SSH push using the user's private key
+   * @param {any} push The push object
+   * @param {Buffer} privateKey The user's SSH private key
+   * @return {Promise<boolean>} True if successful
+   */
+  private static async executeSSHPushWithUserKey(push: any, privateKey: Buffer): Promise<boolean> {
```
jescalada, line 86: Can we enforce correct typing on the `push` object? Otherwise we might have issues with undefined parameters that we use below (`branch`, `proxyGItPath`, etc.).

src/service/SSHKeyForwardingService.ts
```diff
@@ -0,0 +1,216 @@
...
+      return false;
+    }
+  }
+
+  /**
+   * Execute SSH push using the proxy's SSH key (fallback)
+   * @param {any} push The push object
+   * @return {Promise<boolean>} True if successful
+   */
+  private static async executeSSHPushWithProxyKey(push: any): Promise<boolean> {
```
jescalada, line 135: Same as above, would be great to enforce proper types for `push`.

src/service/SSHKeyForwardingService.ts
```diff
@@ -0,0 +1,216 @@
...
+      return false;
+    }
+  }
+
+  /**
+   * Execute HTTPS push (no SSH key needed)
+   * @param {any} push The push object
+   * @return {Promise<boolean>} True if successful
+   */
+  private static async executeHTTPSPush(push: any): Promise<boolean> {
```
jescalada, line 169: Same as above

src/service/SSHKeyForwardingService.ts
jescalada, line 1: Would be great to define specific errors and throw them from each file/operation. Such as `SSHKeyForwardingError`, etc.

src/service/urls.js
```diff
@@ -1,20 +1,79 @@
...
+
 module.exports = {
   getProxyURL: (req) => {
-    const defaultURL = `${req.protocol}://${req.headers.host}`.replace(
-      `:${UI_PORT}`,
-      `:${PROXY_HTTP_PORT}`,
-    );
+    const protocol = normaliseProtocol(req?.protocol);
+    const host = resolveHost(req);
+    const defaultURL = `${protocol}://${host}`.replace(`:${UI_PORT}`, `:${PROXY_HTTP_PORT}`);
```
jescalada: Might want to use a port replacement function rather than just replacing the first occurrence of the port:

`https://example.com:3000/path:3000`.replace(`:3000`, `:8000`)
Result: "https://example.com:8000/path:3000"

```ts
const replacePort = (url, oldPort, newPort) => {
  try {
    const parsed = new URL(url);
    if (parsed.port === String(oldPort)) {
      parsed.port = String(newPort);
    }
    return parsed.toString();
  } catch (error) {
    console.error('Failed to parse URL:', error);
    return url;
  }
};
```

src/service/urls.ts
jescalada, line 1: Might want to convert this to TS! 😃

jescalada (changes requested):
Code review (Part 2/3). Looking good so far, just a few spots that I found possible ways to improve the implementation! 😃

packages/git-proxy-cli/index.ts
```diff
@@ -436,6 +491,37 @@ yargs(hideBin(process.argv)) // eslint-disable-line @typescript-eslint/no-unused
...
+        demandOption: true,
+        type: 'string',
+      },
+    },
+    handler(argv) {
+      if (argv.action === 'add') {
+        addSSHKey(argv.username, argv.keyPath);
+      } else if (argv.action === 'remove') {
+        // TODO: Implement remove SSH key
+        console.error('Error: SSH key: Remove action not implemented yet');
```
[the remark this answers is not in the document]
jescalada: Confirmed this is present now 👍🏼

src/proxy/ssh/server.js
jescalada: Answer: `22` is the default SSH port for connecting to the Git server, whereas `2222` is an arbitrarily-defined port for the proxy server. We might want to add a constant just to clarify the purposes of each.

src/db/file/users.ts
jescalada: Possible solution: enforce `sshKey` uniqueness in DB to prevent duplicates

src/proxy/processors/push-action/captureSSHKey.ts
jescalada: Check: Which of these are correct - then remove the ones that are no longer relevant

src/proxy/processors/push-action/captureSSHKey.ts
```diff
@@ -0,0 +1,99 @@
...
+      ? Buffer.isBuffer(publicKeySource)
+        ? Buffer.from(publicKeySource)
+        : Buffer.from(publicKeySource)
+      : Buffer.alloc(0);
+
+    // For this implementation, we need to work with SSH agent forwarding
+    // In a real-world scenario, you would need to:
+    // 1. Use SSH agent forwarding to access the user's private key
+    // 2. Store the key securely with proper encryption
+    // 3. Set up automatic cleanup
```
jescalada, line 53: @dcoric Double-check if modifying/improving implementation is required

src/service/urls.ts
jescalada: Conversion done as part of merge conflict resolution.

test/.ssh/host_key
jescalada, line 1: Double-check if this needs to be removed (autogenerated) or if required for test fixture. Also other SSH key sample files

packages/git-proxy-cli/index.ts
```diff
@@ -335,89 +335,61 @@ async function logout() {
...
+          'Content-Type': 'application/json',
+        },
+        withCredentials: true,
       },
     );
 
-    console.log(`User '${username}' created successfully`);
+    console.log(`SSH key added successfully for user ${username}`);
   } catch (error: any) {
-    let errorMessage = `Error: Create User: '${error.message}'`;
```
jescalada, line 400: Note: Might be worth refactoring the error handling here to use promises and possibly reusing the endpoint error message instead of repeating it here.

src/cli/ssh-key.ts
```diff
@@ -0,0 +1,140 @@
...
+        },
+      },
+    );
+
+    console.log('SSH key added successfully!');
+  } catch (error) {
+    const axiosError = error as ErrorWithResponse;
+    console.error('Full error:', error);
+
+    if (axiosError.response) {
```
jescalada, line 60: Note for self: Error handling here should be improved by leveraging the status codes (same as before) and also using promises/`.then` in case we need to do multiple requests.

src/service/routes/users.ts
```diff
@@ -29,4 +36,68 @@ router.get('/:id', async (req, res) => {
...
+  }
+
+  const { publicKey } = req.body;
+  if (!publicKey) {
+    res.status(400).json({ error: 'Public key is required' });
+    return;
+  }
+
+  // Strip the comment from the key (everything after the last space)
+  const keyWithoutComment = publicKey.split(' ').slice(0, 2).join(' ');
```
jescalada: Fixed and removed the old validation in `index.ts` 🚀

src/proxy/ssh/server.js
```diff
@@ -0,0 +1,683 @@
...
+const config = require('../../config');
+const chain = require('../chain');
+const db = require('../../db');
+
+class SSHServer {
+  constructor() {
+    // TODO: Server config could go to config file
+    this.server = new ssh2.Server(
+      {
+        hostKeys: [require('fs').readFileSync(config.getSSHConfig().hostKey.privateKeyPath)],
```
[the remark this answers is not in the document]
jescalada: Added validation for `sshConfig.hostKey.privateKey` which seems to be the only one in use now.

src/cli/ssh-key.ts
jescalada: Fixed this and added a check for whether the key is public or private

src/cli/ssh-key.ts
```diff
@@ -0,0 +1,133 @@
...
+interface ErrorWithResponse {
+  response?: {
+    data: ApiErrorResponse;
+    status: number;
+  };
+  code?: string;
+  message: string;
+}
+
+async function addSSHKey(username: string, keyPath: string): Promise<void> {
```
jescalada, line 26: Just noticed that there's duplicate logic for `addSSHKey` and `removeSSHKey` - the same API calls are happening in `git-proxy-cli/index.ts`. We should figure out if we want to keep the standalone SSH key manager script or have it all included in the CLI.

src/proxy/processors/push-action/captureSSHKey.ts
jescalada: Fixed (refactored/simplified `captureSSHKey.ts`).

src/proxy/processors/push-action/captureSSHKey.ts
jescalada: Fixed (refactored/simplified `captureSSHKey.ts`).

src/proxy/processors/push-action/captureSSHKey.ts
jescalada: Modified this to throw an error instead to improve security. I'll do some extra testing to see if this actually gets triggered under normal circumstances.

src/db/file/users.ts
fabiovincenzi: Fixed in https://github.com/finos/git-proxy/pull/987/commits/f9e5e9df693bc1e0a99ca68873cf04db0e392471

src/proxy/processors/push-action/captureSSHKey.ts
fabiovincenzi: cleaned both buffers at the end in https://github.com/finos/git-proxy/pull/987/commits/d5920a2ef9ac7af2eeea051c416c078fe28bd7cb

jescalada: Solved merge conflicts for this 👍🏼

src/proxy/processors/push-action/pullRemote.ts
jescalada: Fixed in https://github.com/finos/git-proxy/pull/987/commits/f1b4ddbd61c0728182bf68cded780d3781e6ec2b

src/proxy/processors/push-action/pullRemote.ts
jescalada: Fixed in https://github.com/finos/git-proxy/pull/987/commits/f1b4ddbd61c0728182bf68cded780d3781e6ec2b

PR #972, opened 2025-04-11
feat: support direct querying of AD group membership via LDAP

kriswest:
resolves #909 

Adds support for using LDAP directly to confirm AD group membership, which is enabled when the `apis.ls` config is not set. Also adds documentations for the AD related parts of the config.

PR #963, opened 2025-03-28
feat: enable multiple auth methods

jescalada:
Fixes #962.

This PR builds on top of #961 and adds support for multiple auth methods. Now, it's possible to enable different methods simultaneously without erroring out (local, AD, OIDC).

I'm happy to add tests to improve coverage if needed. 🙂 

### Changelog
- Add support for multiple auth methods in `src/service/passport`
  - Update `src/service/routes/auth.js` routes accordingly
- Rename `/userLoggedIn` to more standard `/me`

PR #998, opened 2025-04-30
refactor: migrate configuration system from JSON Schema to QuickType

fabiovincenzi:
Replace JSON-Schema with Zod for config validation and default injection
Add typed `config` export, `loadConfig()`, `validate()` and `setConfigFile()`
Update` index.ts` to use the new loader and pass the validated config.


kriswest:
Zod schema's aren't the easiest to read and duplicate the content of the config schema from which the config reference is generated... There are docs generators that will turn Zod schemas into markdown, e.g. https://github.com/matejchalk/zod2md (not used this - so not a recommendation), which might provide a solution.

You can also generate zod schemas from JSON-schema (which would also negate the need to dual maintain the config schema and zod schemas) using quicktype: https://app.quicktype.io/#l=typescript-zod

That would be worthwhile if the the zod validation provides an advantage over using ajv or similar to validate directly using the JSON schema - I don't know what that advantage would be...

jescalada:
@kriswest I originally requested this change so that we could enforce stronger typing in some places like the database sink:

```ts
let sink: any;
if (config.getDatabase().type === 'mongo') {
  sink = require('./mongo');
} else if (config.getDatabase().type === 'fs') {
  sink = require('./file');
}
```

I also thought loading the config into something strongly-typed would be more reliable. An example is with the `authentication` methods, `json-config` allows having arbitrary names and key/value for their respective `config`s:

```json
"authentication": [
    {
      "type": "local",
      "enabled": true
    },
    {
      "type": "ActiveDirectory",
      "enabled": false,
      "adminGroup": "",
      "userGroup": "",
      "domain": "",
      "adConfig": {
        "url": "",
        "baseDN": "",
        "searchBase": ""
      }
    },
    {
      "type": "openidconnect",
      "enabled": false,
      "oidcConfig": {
        "issuer": "",
        "clientID": "",
        "clientSecret": "",
        "callbackURL": "",
        "scope": ""
      }
    }
  ],
``` 

Strong typing would make sure that each value is truly valid during compilation, not runtime.

Another use case for this is preventing things from blowing up given "certain" config rules. For example, prior to #963, enabling multiple auth methods would result in runtime errors. Other "unexpected" combinations or values will also cause runtime errors instead of failing fast.

> Zod schema's aren't the easiest to read and duplicate the content of the config schema from which the config reference is generated... There are docs generators that will turn Zod schemas into markdown, e.g. https://github.com/matejchalk/zod2md (not used this - so not a recommendation), which might provide a solution.

I love the idea of generating documentation for the schema! It'd be great UX-wise, since the current `config.schema.json` is not that intuitive. We could have examples and explanations for each entry. And users can rest assured that there isn't unexpected behavior due to the config not being validated 😃

kriswest:
A few more thoughts on alternatives:

> Strong typing would make sure that each value is truly valid during compilation, not runtime.

Fair enough - what do you think about using zod schemas generated from the json-schema to achieve that? One reason to handle it that way would be to continue using the config reference docs generated from the schema (i.e. https://git-proxy.finos.org/docs/configuration/reference - note gaps in those docs are due to gaps in the schema itself).

You can also generate standard Typescript types with quicktype, which come with convert functions to read the JSON and validate it. We use it extensively on the [FDC3 project](https://github.com/finos/fdc3), e.g.:

- [A schema (part of a fairly large pack of them)](https://github.com/finos/FDC3/blob/main/packages/fdc3-schema/schemas/api/WCP1Hello.schema.json)
- [A generated type (with docs from the schema)](https://github.com/finos/FDC3/blob/63eceaa9d1f1790c0c6d0f463b010965a187dd07/packages/fdc3-schema/generated/api/BrowserTypes.ts#L90-L150)
- [Example convert function that parses the JSON, validates and returns strongly typed class](https://github.com/finos/FDC3/blob/63eceaa9d1f1790c0c6d0f463b010965a187dd07/packages/fdc3-schema/generated/api/BrowserTypes.ts#L3937-L3940)

FDC3s use of schemas is much more complex - with only one schema file here things would be much simpler to add. Something I'd be happy to contribute if it was desired.

The same tooling can generate zod schemas instead of TS types and convert functions.

> Another use case for this is preventing things from blowing up given "certain" config rules. For example, prior to https://github.com/finos/git-proxy/pull/963, enabling multiple auth methods would result in runtime errors. Other "unexpected" combinations or values will also cause runtime errors instead of failing fast.

You can express some quite complex restrictions in json-schema, then validate at runtime with ajv or a generated convert class (as above) and fail fast (exit with a complaint and detail as to what failed validation). 

> I love the idea of generating documentation for the schema! It'd be great UX-wise, since the current config.schema.json is not that intuitive. We could have examples and explanations for each entry. And users can rest assured that there isn't unexpected behavior due to the config not being validated 😃

The existing reference that gets generated: https://git-proxy.finos.org/docs/configuration/reference isn't using all the feature available (in json-schema / json-schema-for-humans) yet to due to a lack of descriptions and examples in the schemas (and has quite a few missing sub-property definitions). If you check out the preview on #972 you can see what I propose to add to the authentication schema and what that looks like: https://deploy-preview-972--endearing-brigadeiros-63f9d0.netlify.app/docs/configuration/reference

src/config/config.ts
jescalada, line 1: Is this file entirely auto-generated, or just a part of it?

We might want to separate the auto-generated portion (if it's only partly generated) and add some kind of `//AUTO-GENERATED. DO NOT EDIT.` bit to make it more obvious. It might also help to generate the file into `src/config/generated/config.ts` or something like that.

fabiovincenzi: Yes, it is totally auto-generated, you are right, I wanted to do that, and I forgot it

PR #1006, opened 2025-05-08
feat: report illegal commit messages

### LINKED ISSUE #1005: When illegal commit messages are reported, it can be hard to figure out which message was flagged
When a commit is rejected due to an illegal message, it can be hard to tell which commit message is at fault as the illegal message and commit details are not reported back to the user.

The illegal commit messages should be reported, with any other details of the commit that can be reported (e.g. SHA)

I'll raise a PR to resolve this

kriswest:
resolves #1005 

Adjust the error message for checkCommitMessages to ensure it includes any illegal commit messages found. This should make it easier for the user to track down which commits had illegal messages. 

This is probably the minimum change to improve the situation. A more detailed change might rework the exec function to collect the SHA of illegal messages and output that as well


JamieSlome (approved):
LGTM! 🍰

PR #1007, opened 2025-05-14
ci: generate js and definitions from ts and place in original location

06kellyjac:
- **chore: dont publish experimental or cypress**
- **build: generate js and definitions from ts and place in original location**
- **build: add guardrails and add to publish**


JamieSlome (approved):
LGTM! 🍰

PR #1010, opened 2025-05-16
feat: integrate gitleaks

06kellyjac:
integrate support for gitleaks checks (default off for now)
Users must make the gitleaks binary available on the PATH for git-proxy and enable via the git-proxy settings

### DIFF
--- a/src/proxy/processors/push-action/gitleaks.ts
+++ b/src/proxy/processors/push-action/gitleaks.ts
@@ -1,0 +1,184 @@
+import { Action, Step } from '../../actions';
+import { getAPIs } from '../../../config';
+import { spawn } from 'node:child_process';
+import fs from 'node:fs/promises';
+import { PathLike } from 'node:fs';
+
+const EXIT_CODE = 99;
+
+function runCommand(
+  cwd: string,
+  command: string,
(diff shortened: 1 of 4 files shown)


src/proxy/processors/push-action/gitleaks.ts
```diff
@@ -0,0 +1,146 @@
...
+      reject(err);
+    });
+  });
+}
+
+type ConfigOptions = {
+  enabled: boolean;
+  ignoreGitleaksAllow: boolean;
+  noColor: boolean;
+};
```
JamieSlome, line 45: Should we add this to the schema for the auto docs? @06kellyjac

src/proxy/processors/push-action/gitleaks.ts
```diff
@@ -0,0 +1,146 @@
+import { Action, Step } from '../../actions';
+import { getAPIs } from '../../../config';
+// import { getCommitConfig, getPrivateOrganizations } from '../../../config';
```
JamieSlome, line 3:
```suggestion
```

JamieSlome (approved):
LGTM! 🍰

PR #1025, opened 2025-05-23
test: improve proxy route test coverage

jescalada:
This PR improves test coverage for `/src/proxy/routes`. Overall project coverage should increase by ~5%.

Originally, I wanted to improve coverage for `/plugin.ts`, however I've been having problems with the tests not executing due to an ESM interop problem when using the `plugin-loader` library.

I added a `skip` on all the problematic tests in `plugin.test.js`, and kept the ones that do work.

I tried a few workarounds (such as executing with regular `mocha` + `tsx` loader, running with `vitest`) without any luck. The only working solution I found so far is to upgrade to Node 22.


JamieSlome (approved):
LGTM! 🍰

PR #1033, opened 2025-05-28
feat: SSH UI

fabiovincenzi:
Screenshot 2025-05-28 at 13 30 09
Screenshot 2025-05-28 at 13 29 56
Screenshot 2025-05-28 at 13 29 50


src/ui/components/CustomButtons/CodeActionButton.jsx
```diff
@@ -6,14 +7,29 @@ import {
...
   const [isCopied, setIsCopied] = useState(false);
+  const [protocol, setProtocol] = useState('https');
+
+  const getSSHUrl = () => {
+    try {
+      const urlObj = new URL(cloneURL);
+      const host = urlObj.host;
+      let path = urlObj.pathname;
+      if (path.startsWith('/')) path = path.substring(1);
+      return `git@${host}:${path}`;
```
dcoric: During manual testing, I noticed a potential issue with SSH port handling. 

The current implementation works correctly when:
1. The server is mapped to a domain name (no port in URL)
2. The SSH port is the default (22)

However, if a custom SSH port is configured (not 22), the clone command should be formatted as:
`git clone ssh://git@${host}:${SSHport}/${path}`

**Suggested Improvement:**
Let's make this more robust by:
1. Checking the configured SSH port from settings
2. Dynamically adjusting the clone URL format based on whether:
   - It's the default port (22): `git@${host}:${path}`
   - It's a custom port: `ssh://git@${host}:${SSHport}/${path}`

This would handle all cases correctly while maintaining backward compatibility.

dcoric (changes requested):
**Review Note:**

The implementation looks good overall, but we need to address the SSH port handling in clone URLs.

**Requested Changes:**

1. Currently when a non-standard SSH port is configured, the clone URL format needs to be adjusted to:
   ```sh
   git clone ssh://git@${host}:${SSHport}/${path}

src/db/file/users.ts
```diff
@@ -25,11 +25,7 @@ export const findUser = (username: string) => {
       if (err) {
         reject(err);
       } else {
-        if (!doc) {
-          resolve(null);
-        } else {
-          resolve(doc);
-        }
+        resolve(doc || null);
```
jescalada, line 28: Good catch! It's much more readable now.

src/ui/components/CustomButtons/CodeActionButton.jsx
```diff
@@ -6,14 +8,50 @@ import {
   CodeIcon,
   CopyIcon,
   TerminalIcon,
+  KeyIcon,
 } from '@primer/octicons-react';
-import React, { useState } from 'react';
+
+const API_BASE = import.meta.env.VITE_API_URI ?? '';
```
jescalada: I wonder if there are edge cases where the `VITE_API_URI` is not present? Other parts of the app seem to have a fallback value (current origin):

```ts
const baseUrl = import.meta.env.VITE_API_URI
  ? `${import.meta.env.VITE_API_URI}`
  : `${location.origin}`;
```

src/ui/components/SSHKeysManager/SSHKeysManager.jsx
```diff
@@ -0,0 +1,169 @@
...
+  modalContent: {
+    backgroundColor: 'white',
+    padding: theme.spacing(4),
+    borderRadius: 8,
+    width: 400,
+  },
+  formField: { marginBottom: theme.spacing(2) },
+}));
+
+const API_BASE = `${import.meta.env.VITE_API_URI}/api/v1/user`;
```
jescalada: Same as before: Are we guaranteed to have that `VITE_API_URI` value or should we set some fallback value? Maybe a unified `API_BASE_URL` constant would be great throughout the frontend.

src/ui/components/CustomButtons/CodeActionButton.jsx
```diff
@@ -6,14 +8,50 @@ import {
...
   const [placement, setPlacement] = useState();
   const [isCopied, setIsCopied] = useState(false);
+  const [protocol, setProtocol] = useState('https');
+
+  const [sshCfg, setSshCfg] = useState({ enabled: true, port: 22 });
+  useEffect(() => {
+    axios
+      .get(`${API_BASE}/api/v1/config/ssh`, { withCredentials: true })
+      .then((res) => {
+        const { enabled = true, port = 22 } = res.data || {};
```
jescalada: Is it necessary to add default values here if the API already sets the same default values (`/src/config/index.ts`)? 🤔 (Maybe it does help in the rare scenario someone pulls these changes without reloading their backend, since the frontend gets updated dynamically).

src/db/file/users.ts
```diff
@@ -132,10 +132,74 @@ export const getUsers = (query: any = {}) => {
...
+        if (!user) {
+          reject(new Error('User not found'));
+          return;
+        }
+        if (!user.publicKeys) {
+          user.publicKeys = [];
+        }
+        if (!user.publicKeys.includes(publicKey)) {
+          user.publicKeys.push(publicKey);
+          exports.updateUser(user).then(resolve).catch(reject);
```
jescalada: I seem to get an error when adding a key:

```bash
[0] Adding SSH key { targetUsername: 'admin', keyWithoutComment: 'blah' }
[0] Error adding SSH key: TypeError: exports.updateUser is not a function
[0]     at <anonymous> (/home/juan/Desktop/Juan/Projects/git-proxy/src/db/file/users.ts:163:19)
[0]     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
^C[0] npm run server exited with code SIGINT
[1] npm run client exited with code SIGINT
```

This can be fixed by modifying the `addPublicKey` and `removePublicKey` so that we use this line instead of the `exports.(...)` call:

```diff
- exports.updateUser(user).then(resolve).catch(reject);
+ updateUser(user).then(() => resolve(user)).catch(reject);
```

```suggestion
          updateUser(user).then(() => resolve(user)).catch(reject);
```

src/db/file/users.ts
```diff
@@ -132,10 +132,74 @@ export const getUsers = (query: any = {}) => {
...
+          reject(new Error('User not found'));
+          return;
+        }
+        if (!user.publicKeys) {
+          user.publicKeys = [];
+          resolve(user);
+          return;
+        }
+        user.publicKeys = user.publicKeys.filter((key) => key !== publicKey);
+        exports.updateUser(user).then(resolve).catch(reject);
```
jescalada: Same as the previous call:

```suggestion
        updateUser(user).then(() => resolve(user)).catch(reject);
```

src/service/routes/users.js
```diff
@@ -29,4 +38,142 @@ router.get('/:id', async (req, res) => {
...
+  }
+
+  const { publicKey } = req.body;
+  if (!publicKey) {
+    res.status(400).json({ error: 'Public key is required' });
+    return;
+  }
+
+  // Strip the comment from the key (everything after the last space)
+  const keyWithoutComment = publicKey.split(' ').slice(0, 2).join(' ');
```
jescalada: It would be nice to make the validation here a bit more robust. Invalid strings such as "blah" end up getting added as keys, which then get parsed through the `GET /:username/ssh-keys` as empty strings:

image

In an ideal world, we would have an extra validation step and then catch those errors in the frontend while displaying a descriptive `Snackbar` (for example if the user inputs an SSH key in invalid format). We could also have a description in the "Add New SSH Key" modal explaining the valid format.

src/ui/components/SSHKeysManager/SSHKeysManager.jsx
```diff
@@ -0,0 +1,169 @@
...
+            </Grid>
+            <Grid item>
+              <Typography variant='h6'>{key.name}</Typography>
+            </Grid>
+          </Grid>
+
+          <Grid container alignItems='center' justifyContent='space-between'>
+            <Grid item xs>
+              <Typography variant='body2' color='textSecondary'>
+                {key.hash}
```
jescalada: Might be good to add a title/text to show that the "hash" here is just the SHA-256 fingerprint rather than the actual key.

I feel users would still have a pretty tough time identifying/managing keys, since right now they only add the key itself (which gets hashed). I'm not too knowledgeable on the topic, so I used AI to generate some recommendations:

> The SHA-256 fingerprint can be useful, but by itself, it’s not always helpful from a user experience standpoint, especially when you’re trying to manage or recognize SSH keys at a glance.
>
> The fingerprint (like SHA256:abcd...) is a cryptographic hash of the actual public key. It's primarily used to verify that a public key you received (or stored) is the same one the server has, without exposing the key directly. So yes, it is good from a security perspective, because it allows identification without revealing the key... but! It’s not very user-friendly.
>
> Users often need more context when managing SSH keys. Just a name and a fingerprint is like labeling keys in your drawer with barcodes... sure, it's unique, but what does it mean?
>
> More helpful identifiers could include:
>
> - **Comment field**: Most SSH keys are generated with a comment at the end (like user@host). This is often included in the .pub file and can be a great visual hint. You can extract and show it if available.
>
> - **Key type and length**: Something like ed25519 256-bit or rsa 4096-bit helps users understand what kind of key they’re dealing with. This also helps flag weak keys.
>
> - **Creation time or added date**: Useful when people rotate keys but keep old ones for comparison.
>
> - **Partial public key**: Showing the first ~10 and last ~10 characters of the base64-encoded public key (like GitHub does) can give a recognizable pattern without revealing the whole key.
>
> - **Associated user/device name** (if provided at registration): E.g. “MacBook Pro key”, “CI deploy key”, etc. This is probably what the "name" field in your PR is meant to be, right? Just make sure it’s user-set and not auto-generated.

I guess we could add a few extra optional fields to help UX for key management (even if key details are not actually used internally).

src/ui/components/SSHKeysManager/SSHKeysManager.jsx
```diff
@@ -0,0 +1,169 @@
...
+      </Button>
+
+      {keys.map((key, idx) => (
+        <Paper key={idx} className={classes.keyContainer} elevation={3}>
+          <Grid container alignItems='center' spacing={1} style={{ marginBottom: 8 }}>
+            <Grid item>
+              <VpnKeyIcon color='primary' />
+            </Grid>
+            <Grid item>
+              <Typography variant='h6'>{key.name}</Typography>
```
jescalada, line 139: Are these key names generated automatically? We may want to allow setting the key name in the "Add Key" modal and storing it in the database. The name itself shouldn't affect the execution, it would be purely a UX improvement.

PR #1038, opened 2025-06-02
test: increase action test coverage

jescalada:
PR for increasing coverage for actions (`/src/proxy/processors`).

`parsePush.ts` which will be covered after merging all the recent changes to that file.


test/plugin/plugin.test.js
JamieSlome, line 1: @jescalada - why are we skipping these?

jescalada: The plugins are apparently working correctly, but the tests fail due to incompatibilities with `ts-mocha` and Node 18 (CommonJS -> ESM refactor problem).

Running the tests on Node 22 works without issue, and the plugin feature itself does work when executed via `tsx`. But so as long as our test environment runs Node 18 (chances are, we want to stay at earlier versions for broader compatibility), those tests will fail and clog the CI...

JamieSlome: Cool - can we create an issue to track this issue and assign the two of us to it? Resolving this conversation and merging 👍

jescalada: Done! #1045 

Thanks for the reviews!

Issue #978, opened 2025-04-15
Cypress tests require data created by mocha tests, but should be self-contained


kriswest:
**Describe the bug**
The cypress tests for the repo page depend on data that is created and abandoned by the mocha tests. I.e. the `test-repo` used here: https://github.com/finos/git-proxy/blob/371894323bae83a2aedb8d2944745544359a6275/cypress/e2e/repo.cy.js#L11-L21

The data is created in a mocha test here: https://github.com/finos/git-proxy/blob/371894323bae83a2aedb8d2944745544359a6275/test/addRepoTest.test.js#L44-L49

Which also has to handle it (perhaps) already existing on run: 
https://github.com/finos/git-proxy/blob/371894323bae83a2aedb8d2944745544359a6275/test/addRepoTest.test.js#L26-L29

Ideally tests should handle their own set-up and tear-down (clean-up). As the cypress tests use client-side JS they can't directly import the DB clients and create the data they need, but they could make API calls or a node script be used for setup and tear down before/after the cypress command.

**Expected behaviour**
Each set of tests handles its own data creation and subsequent clean-up.

**Additional context**

PR #1043, opened 2025-06-03
feat(key on repo url): support git hosts other than GitHub + multiple forks

### LINKED ISSUE #66: Provide facility to manage multiple forks of the same upstream repository

### LINKED ISSUE #511: Git Proxy does not support GitLab repos

### LINKED ISSUE #950: Repositories should be identified and retrieved via URL rather than name

kriswest:
resolves #950 
resolves #511
resolves #66 

Refactor (api, proxy & UI) to remove the assumption of GitHub as the git repository host and the use of the repository `name` field as the id of the repository (as this prevents git-proxy instances from supporting multiple forks of a project or projects from multiple hosts with the same name).

Based on #979 - I'll rebase after that merges which should drop the number of affected files down some.

This PR:
- Replaces the use of the repo `name` field in the API with the _id field generated by the database adaptors, 
  - Using the repository URL as a key does not work well with express routing, but _id does in both mongo and neDb
  - allows `names` to be repeated (multiple forks or clashing names from different organisations/repository hosts)
  - UI and CLI were updated accordingly
- Replaces the use of `organisation/repoName.git` in the proxy URLs with the repository url
  - i.e. https://myGitProxyInstance.com:8443/finos/git-proxy.git
  becomes
  https://myGitProxyInstance.com:8443/github.com/finos/git-proxy.git
  - Protocol is omitted from the embedded URL as thats assumed to be https://
  - Refactors the proxy to be able to proxy traffic to multiple different hosts
    - _may_ require a restart after adding a host - further work may need to resolve that by restarting the proxy 
  - Backwards compatibility is provided for old proxy urls by assuming github.com as the default
- Disables GitHub specific functionality in the UI if the host is not Github
- Completes application of Typescript to the database classes
  - Duplicated code reduced
  - A number of minor differences in behaviour (particularly return types) between the DB adaptors were resolved
  - Does NOT refactor all usages of the DB client to use typescript (still many requires to eliminate)
- Deprecates and ignores the config property proxyUrl as the proxied host(s) are now determined from the configured repositories

To Do:
- [ ] Annotate PR for review
- [ ] Check test coverage
- Implement additional tests for the proxy and fallback
  - [x] implement tests for new proxy URLs for github.com
  - [x] implement tests for fallback with legacy proxy urls for github.com
  - [ ] implement tests for gitlab.com
  - [ ] implement tests for non-github/non-gitlab repo
  - [ ] implement tests for multiple forks
- [ ] Add support for GitLab API where repo is hosted at GitLab

(contributed as part of a [GitLab CoCreate collaboration](https://about.gitlab.com/community/co-create/) with help from @StingRayZA)


src/ui/views/RepoList/Components/RepoOverview.jsx
```diff
@@ -578,20 +578,24 @@ export default function Repositories(props) {
     getGitHubRepository();
   }, [props.data.project, props.data.name]);
 
+  // TODO add support for GitLab API: https://docs.gitlab.com/api/projects/#get-a-single-project
```
sam-holmes2: Flagging for awareness

src/proxy/routes/index.ts
```diff
@@ -1,140 +1,140 @@
...
+const proxyErrorHandler: ProxyOptions['proxyErrorHandler'] = (err, res, next) => {
+  console.log(`ERROR=${err}`);
+  next(err);
 };
+
+// eslint-disable-next-line new-cap
+const router = Router();
+
+getAllProxiedHosts().then((originsToProxy) => {
+  // TODO: this will only happen on startup. We'll need to add routes at runtime when new origins are added? Or force a restart for the proxy to work
```
sam-holmes2: Flagging for awareness

sam-holmes2 (approved):
LGTM after an initial scan through :) thanks for your contribution!

src/proxy/routes/index.ts
kriswest: I was reading up on a solution for this... as we're using a router it can be reinitialized. The trick will be figuring out how to tell it to do so!

kriswest: I've added a solution for this, but need to add a test for it, which I'll do today.

src/ui/views/RepoList/Components/RepoOverview.jsx
kriswest: I've had some input from @StingRayZA @ Gitlab on this and will update the PR acordingly. after I've dealt with the test mentioned below.

src/db/file/repo.ts
```diff
@@ -11,14 +13,10 @@ if (!fs.existsSync('./.data')) fs.mkdirSync('./.data');
 if (!fs.existsSync('./.data/db')) fs.mkdirSync('./.data/db');
 
 const db = new Datastore({ filename: './.data/db/repos.db', autoload: true });
-db.ensureIndex({ fieldName: 'name', unique: false });
+db.ensureIndex({ fieldName: 'url', unique: true });
```
jescalada: I found an issue when trying to run the backend with a pre-populated database (same behaviour when running `npm run test`:

```bash
$ npm run start

> @finos/git-proxy@1.17.2 start
> concurrently "npm run server" "npm run client"

[0] 
[0] > @finos/git-proxy@1.17.2 server
[0] > tsx index.ts
[0] 
[1] 
[1] > @finos/git-proxy@1.17.2 client
[1] > vite --config vite.config.js
[1] 
[1] 
[1]   VITE v4.5.13  ready in 287 ms
[1] 
[1]   ➜  Local:   http://localhost:3000/
[1]   ➜  Network: use --host to expose
[0] Configuration sources are disabled
[0] authMethods: [{"type":"local","enabled":true}]
[0] No plugins configured
[0] Found 0 plugin modules
[0] /home/juan/Desktop/Juan/Projects/git-proxy/node_modules/@seald-io/binary-search-tree/lib/avltree.js:261
[0]           const err = new Error(`Can't insert key ${JSON.stringify(key)}, it violates the unique constraint`)
[0]                       ^
[0] 
[0] Error: Can't insert key "https://github.com/finos/git-proxy.git", it violates the unique constraint
[0]     at _AVLTree.insert (/home/juan/Desktop/Juan/Projects/git-proxy/node_modules/@seald-io/binary-search-tree/lib/avltree.js:261:23)
[0]     at AVLTree.insert (/home/juan/Desktop/Juan/Projects/git-proxy/node_modules/@seald-io/binary-search-tree/lib/avltree.js:23:31)
[0]     at Index.insert (/home/juan/Desktop/Juan/Projects/git-proxy/node_modules/@seald-io/nedb/lib/indexes.js:117:40)
[0]     at Index.insertMultipleDocs (/home/juan/Desktop/Juan/Projects/git-proxy/node_modules/@seald-io/nedb/lib/indexes.js:154:14)
[0]     at Index.insert (/home/juan/Desktop/Juan/Projects/git-proxy/node_modules/@seald-io/nedb/lib/indexes.js:108:12)
[0]     at Index.reset (/home/juan/Desktop/Juan/Projects/git-proxy/node_modules/@seald-io/nedb/lib/indexes.js:93:23)
[0]     at Datastore._resetIndexes (/home/juan/Desktop/Juan/Projects/git-proxy/node_modules/@seald-io/nedb/lib/datastore.js:427:13)
[0]     at Persistence.loadDatabaseAsync (/home/juan/Desktop/Juan/Projects/git-proxy/node_modules/@seald-io/nedb/lib/persistence.js:330:15)
[0]     at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
[0]   key: 'https://github.com/finos/git-proxy.git',
[0]   errorType: 'uniqueViolated'
[0] }
[0] 
[0] Node.js v20.19.2
^C[1] npm run client exited with code SIGINT
```

I've done some investigating, and it seems to be triggered by this line.

```diff
+ db.ensureIndex({ fieldName: 'url', unique: true });
```

For reference, my database contains a duplicate URL because a fork of git-proxy (jescalada/git-proxy) was stored with the upstream repo URL (`repos.db`):

```json
{"project":"jescalada","name":"git-proxy","url":"https://github.com/finos/git-proxy.git","users":{"canPush":["admin"],"canAuthorise":["admin"]},"_id":"TOfValNuxq2P4TXJ"}

{"project":"finos","name":"git-proxy","url":"https://github.com/finos/git-proxy.git","users":{"canPush":["admin"],"canAuthorise":["admin"]},"_id":"pGocyKNIReZ73cNO"}
```

Maybe we could sanitize the database on load to prevent these duplicates? People with "old" versions of the database might experience the same issue.

packages/git-proxy-cli/test/testCliUtils.js
```diff
@@ -152,8 +152,9 @@ async function addRepoToDb(newRepo, debug = false) {
   const found = repos.find((y) => y.project === newRepo.project && newRepo.name === y.name);
   if (!found) {
     await db.createRepo(newRepo);
-    await db.addUserCanPush(newRepo.name, 'admin');
-    await db.addUserCanAuthorise(newRepo.name, 'admin');
+    const repo = await db.getRepoByUrl(newRepo.url);
+    await db.addUserCanPush(repo._id, 'admin');
```
jescalada, line 156: Is there a specific reason to use the repo `_id` instead of the `url`? Since both have uniqueness enforced 🤔

src/db/index.ts
```diff
@@ -32,54 +43,142 @@ export const createUser = async (
...
-  removeUserCanAuthorise,
-  removeUserCanPush,
-  deleteRepo,
-  isUserPushAllowed,
-  canUserApproveRejectPushRepo,
-  canUserApproveRejectPush,
-  canUserCancelPush,
-  getSessionStore,
-} = sink;
+export const createRepo = async (repo: AuthorisedRepo) => {
```
jescalada, line 75: Out of curiousity, why were these methods specifically moved to `index.ts`? Was it to improve the typing? 🤔

src/db/mongo/repo.ts
```diff
@@ -1,109 +1,71 @@
...
   const collection = await connect(collectionName);
-  await collection.updateOne({ name: name }, { $push: { 'users.canPush': user } });
+  const doc = collection.findOne({ _id: new ObjectId(_id) });
+  return doc ? toClass(doc, Repo.prototype) : null;
 };
 
-export const addUserCanAuthorise = async (name: string, user: string) => {
-  name = name.toLowerCase();
-  user = user.toLowerCase();
+export const createRepo = async (repo: Repo): Promise<Repo> => {
```
jescalada, line 35: Are we handling missing values before calling `createRepo`? Or does the MongoDB client do this manually? The previous errors were pretty comprehensive:

```diff
- if (isBlank(repo.project)) {
-    throw new Error('Project name cannot be empty');
-  }
-  if (isBlank(repo.name)) {
-    throw new Error('Repository name cannot be empty');
-  }
-  if (isBlank(repo.url)) {
-    throw new Error('URL cannot be empty');
-  }
-
-  repo.users = {
-    canPush: [],
-    canAuthorise: [],
-  };
```

src/service/routes/repo.js
```diff
@@ -124,16 +125,46 @@ router.delete('/:name/delete', async (req, res) => {
...
+
+        // restart the proxy if we're proxying a new domain
+        if (newOrigin) {
+          console.log('Restarting the proxy to handle an additional origin');
+
+          // 1. Get proxy module dynamically to avoid circular dependency
+          const { proxy } = require('../index');
+
+          // 2. Stop existing services
+          await proxy.stop();
```
jescalada: I got an error here when trying to add a new GitHub repository:

```bash
[0] API request to proxy repository https://github.com/jescalada/open-webui.git is for a new origin: true,
[0]     existing origin list was: ["github.com"]
[0] creating new repo {"project":"jescalada","name":"open-webui","url":"https://github.com/jescalada/open-webui.git","maxUser":1,"users":{"canPush":[],"canAuthorise":[]}}
[0] Restarting the proxy to handle an additional origin
[0] Repository creation failed due to error:  Cannot read properties of undefined (reading 'stop')
[0] node:_http_outgoing:655
[0]     throw new ERR_HTTP_HEADERS_SENT('set');
[0]           ^
[0] 
[0] Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
[0]     at ServerResponse.setHeader (node:_http_outgoing:655:11)
[0]     at ServerResponse.header (/home/juan/Desktop/Juan/Projects/git-proxy/node_modules/express/lib/response.js:794:10)
[0]     at ServerResponse.send (/home/juan/Desktop/Juan/Projects/git-proxy/node_modules/express/lib/response.js:174:12)
[0]     at /home/juan/Desktop/Juan/Projects/git-proxy/src/service/routes/repo.js:168:13
[0]     at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
[0]   code: 'ERR_HTTP_HEADERS_SENT'
[0] }
[0] 
[0] Node.js v20.19.2
```

Interestingly, restarting the server does show the newly added repo in the dashboard:

image

jescalada (changes requested):
I went through the approval/rejection flows with a pre-existing repo, and things work well!

There is an issue with backwards compatibility with older, invalid databases from previous versions of GitProxy (unique URL enforcement with repos). This may also cause issues with the other files (pushes, users).

I also tested the Add Repo flow which caused my server to crash, maybe because of something wrong on my end (invalid input maybe?).

kriswest:
> There is an issue with backwards compatibility with older, invalid databases from previous versions of GitProxy (unique URL enforcement with repos). This may also cause issues with the other files (pushes, users).

I can release the unique constraint on the index to avoid this - however, I put it in to catch invalid data as, like the use of the repository name as an ID, it will result in the selection of the wrong repo project at times (although in this case the multiple records would have to be the same repo) and I thought it better for tests to fail etc. if data wasn't cleaned up from a previous run. Where do you think we should go with this @jescalada - an automated migration seems difficult as you'd just have to delete or programmatically edit one of the duplicate records...

> I also tested the Add Repo flow which caused my server to crash, maybe because of something wrong on my end (invalid input maybe?).

I'll retest straight after OSFF, its possible I broke something while writing tests last week. Something wasn't quite right in the proxy restart at the time, which I thought I had corrected.

src/db/file/repo.ts
kriswest: The only ways I can think of to sanitize the DB are to:
- remove one of the records... actions link to the repos by URL so there should be no push records to update.
  - The challenge will be getting the right one to retain - which should probably be the first found.
- update one of the records to change the URL - perhaps artificially breaking it.
  - Same challenge on selecting one to keep.

The alternative is that we don't have a unique index constraint and just let the duplicate exist. If the databases are predictable on the record they select first this allows us to continue as is.

However, I might consider just letting it break and forcing cleanup of invalid database records...

packages/git-proxy-cli/test/testCliUtils.js
kriswest: The reason I used the _id is that it is used in the API & UI where these operations are used heavily. The original repo URL is hard to embed in the application URLs as its number of components (between /s) is variable, so we used the _id there and then replicated the same pattern in the test implementations.

src/db/index.ts
kriswest: Yes and to reduce/eliminate duplicated code. There was a lot of minor variation between the database clients (particularly on data sanitisation but also extending into return types), by moving all the duplicated code to the `index.ts` we get better test coverage and smaller DB sink implementations.

src/db/mongo/repo.ts
kriswest: Yes createRepo and createUser are both wrapped in the `db/index.ts` file so that the handling for missing values is common between the database clients and the code is not duplicated (or missing in one of the DB clients) anymore. I figured consistency is king when it comes to the database.

src/service/routes/repo.js
kriswest: Thats got to be at least two errors in the proxy restart code - its made all the DB changes by that point and it just needs to restart the proxy if we have a new host to handle. In your example, it doesn't actually need to do that as the list of proxied hosts hasn't changed (you've just added another repo at a particular host). I'll need to take another look at that I guess!

kriswest: I've had a go at the relevant fixes, but can't test properly until after OSFF. If you end up giving it a go let me know @jescalada!

BTW I'm tempted to merge your proxy route test PR into this and update 9as the routes change slightly). It looks like you got a handle on testing the API routes with Chai http (although I haven't run them and did see your comment about issues). I was planning to try and test the `getRouter()` and its output, but your tests might be a better approach, as long as we add one that checks routes after adding a new repository host/origin and triggering a restart of the proxy.

Thoughts?

src/config/index.ts
```diff
@@ -47,15 +46,6 @@ let _config = { ...defaultSettings, ...(_userSettings || {}) } as Configuration;
 // Create config loader instance
 const configLoader = new ConfigLoader(_config);
 
-// Get configured proxy URL
-export const getProxyUrl = () => {
-  if (_userSettings !== null && _userSettings.proxyUrl) {
```
coopernetes, line 59: Instead of removing this function entirely, wouldn't it be useful to emit a warning indicating that this field is deprecated? It can be ignored from use but if it's present in `_userSettings`, we want to indicate that a config change is needed.

https://github.com/finos/git-proxy/blob/d13686d14ba700c404eded2d4c828f14b12612da/src/config/index.ts#L214-L225

packages/git-proxy-cli/test/testCliUtils.js
jescalada: Makes sense, thanks!

src/db/file/repo.ts
jescalada: In the previous case the `url` field contains the upstream repo organization (`finos`) for both. We could possibly delete the (invalid) fork entry or regenerate the `url` field only. The problem right now is that the error message isn't very helpful; maybe catching it somewhere and displaying something more user-friendly is a good enough interim solution?

src/db/index.ts
jescalada: The `db` looks great now! I previously worked on the first TS refactor for it and some parts were too confusing to refactor and test.

src/service/routes/repo.js
jescalada: That'd be super helpful if you don't mind the extra work with the tests! I was actually planning on covering any missing/extra code on a follow-up PR to get the overall coverage to 80%. Maybe we could go ahead and merge this as-is so you don't have to keep maintaining the PR?

jescalada:
> I can release the unique constraint on the index to avoid this - however, I put it in to catch invalid data as, like the use of the repository name as an ID, it will result in the selection of the wrong repo project at times (although in this case the multiple records would have to be the same repo) and I thought it better for tests to fail etc. if data wasn't cleaned up from a previous run. Where do you think we should go with this @jescalada - an automated migration seems difficult as you'd just have to delete or programmatically edit one of the duplicate records...

I think catching and displaying a simple error message with the invalid entry/entries could be enough - so that the GitProxy administrator can quickly identify the issue and fix it manually. Thankfully, the error seems to occur on backend (db) startup, so end users wouldn't really have the app suddenly blowing up.

src/config/index.ts
kriswest: Good point! As the function is no longer called, that would need to be done on load

https://github.com/finos/git-proxy/pull/1043/files#diff-198a1431d7c5e26ea78bc6e54b34b54e6f8ab342dd48ea11013877e8466a87c5R19-R25

I suppose we could suck up the warnings and copies for handling `sslKeyPemPath` and `sslCertPemPath` into the same spot if you thought it worth while.

src/db/file/repo.ts
kriswest: I've added a try/catch on each unique index build (they're also in the user and push tables), it'll log to the console and then proceed without the index allowing cleanup through the UI:

https://github.com/kriswest/git-proxy/blob/e059206bc9babcf74cb1c0cfa1c48ee8c2349ff1/src/db/file/repo.ts#L17-L24

src/proxy/routes/helper.ts
```diff
@@ -0,0 +1,176 @@
...
+ * and processing /github.com/finos/git-proxy.git/info/refs?service=git-upload-pack
+ * would produce:
+ * - repoPath: /github.com/finos/git-proxy.git
+ * - gitPath: /info/refs?service=git-upload-pack
+ *
+ * @param {string} requestPath The URL path to process.
+ * @return {GitUrlBreakdown | null} A breakdown of the components of the URL path.
+ */
+export const processUrlPath = (requestPath: string): UrlPathBreakdown | null => {
+  const components = requestPath.match(PROXIED_URL_PATH_REGEX);
```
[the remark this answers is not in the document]
kriswest: fixed by limiting max Url path length to 512 chars

src/proxy/routes/helper.ts
```diff
@@ -0,0 +1,176 @@
...
+ * Processing https://anotherGitHost.com/project/subProject/subSubProject/repo.git
+ * would produce:
+ * - project: project/subProject/subSubProject
+ * - repoName: repo.git
+ *
+ * @param {string} gitUrl The URL  to process.
+ * @return {GitNameBreakdown | null} A breakdown of the components of the URL.
+ */
+export const processGitURLForNameAndOrg = (gitUrl: string): GitNameBreakdown | null => {
+  const components = gitUrl.match(GIT_URL_NAME_ORG_REGEX);
```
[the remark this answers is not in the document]
kriswest: fixed by limiting max URL length to 512 chars

src/db/file/repo.ts
jescalada: Looks good!

src/service/routes/repo.js
kriswest: This should all be resolved and working now (with tests).  However, restarting the proxy after reloading config is (probably) still not working.

jescalada: I tried going through the Add Repo flow, but it seems that the button is missing (even when logged in as admin). Perhaps a mishap happened during the TSX refactor merge? `main` does seem to show the button as expected 🤔

src/ui/views/RepoDetails/RepoDetails.tsx
jescalada, line 1: It seems this file is different in `main`, the TSX PR extracted and refactored the `RepoDetails` component that was originally in the `languageColors.ts`.

jescalada (commented):
The tests look good so far, and test coverage is excellent! 😃 Just wondering if the merge conflicts with main have been resolved correctly as I wasn't able to do the Add Repo flow...

src/service/routes/repo.js
kriswest: I'll check on this today - I was testing with it by adding https://gitlab.com/gitlab-org/gitlab.git (if you want to give it a go). Note that the add repo button has always been missing for me on first load in older versions, it has only ever appeared after I refresh the page. Hence, I'm pretty sure the timing is off somewhere in the code that displays it...

src/ui/views/RepoDetails/RepoDetails.tsx
kriswest: I think it was the other way round: that PR rather annoyingly renamed the existing RepoDetails.tsx to languageColors.ts and moved it, then introduced a new RepoDetails.tsx file. That made merging that PR into this a real PITA, where moving out the languageColors to a new file would have been a much simpler merge. I'll recheck it.

src/service/routes/repo.js
kriswest: Can you confirm that the button doesn't show up when you refresh the page after logging in? It never shows up for me on first load, including on main. Refresh the page and it appears, then is always present thereafter.

That needs fixing, but I don't think its an issue this PR introduces (but could fix I suppose)

image

src/ui/views/RepoDetails/RepoDetails.tsx
kriswest: I've gone back over this and manually compared to main and I think its correct. It was a horrible hand-merge as the file had been renamed and then recreated on another PR (which breaks its history/prevents automated merges) so I could have missed something...

I see the expect changes:
- Referring to repos by _id instead of project/repoName
- Updated clone URLs containing the proxied host
- Using new util fns for retrieving remote data from GitHub AND GitLab
- Replacing predicted URLs at github for the logos, avatar and profile links etc. with URLs from retrieved data

src/service/routes/repo.js
jescalada: Confirmed that it's working as expected now! 🚀

src/ui/views/RepoDetails/RepoDetails.tsx
jescalada: Sorry about the issues with the PR merge conflicts and the false positive on the "missing" changes... 😓 

I've had something similar happen before with the `config` refactors since the file history often breaks when refactoring JS -> TS.

jescalada (approved):

src/ui/views/RepoList/Components/RepoOverview.jsx
kriswest: this was done

kriswest:
@andypols because adding a whole new origin tends to be rare (after GitHub and GitLab.com, I expect to add some GitLab dedicated instances one day, may the occasional other), and its less complex to restart it than maintain the custom router middleware wrapping the proxies. It could be updated to dynamically maintain the `proxies` and `proxyKeys` elements instead, however, I expect to encounter an issue restarting very very rarely and hence haven't invested the effort to avoid it - instead my time goes into maintaining the existing PRs currently.

kriswest:
P.S. There is a second PR that you may need ##973 - relating to linking PRs to authors via email. I doubt that is involved in the issue, but it is necessary for us to make sure the same user can be associated with commits at GitHub, Gitlab, Bitbucket and other SCM providers.

kriswest:
@andypols in the earlier iterations of this PR I was using the proxy middlewares directly in the Router and the only good way to add/remove them again was to restart. I ended up changing that to wrap them in a custom middleware which uses the proxy middlewares in the `proxies` object. That approach can more easily support update without a restart, but it'd take a refactor of the proxy router to support updating (probably wrapping it in a class, or just merge it into the Proxy class) and a new function on the Proxy class to trigger the update (to replace the [stop/start calls made in the api](https://github.com/kriswest/git-proxy/blob/88330d8dbbb7d70a05c30ab97d65afff42bccdd9/src/service/routes/repo.js#L131-L132).

I'm loathe to take that on before we merge the current state, as I think restarting for a new origin is a low (but not no!) impact flaw and we could deal with it in a subsequent issue/PR.

andypols:
@kriswest 

> I'm loathe to take that on before we merge the current state, as I think restarting for a new origin is a low (but not no!) impact flaw and we could deal with it in a subsequent issue/PR.

Fair enough — I can see that. It was just a question, not a criticism. From my (possibly overly simplistic) perspective, the proxy shouldn't care what the Git domain is. Happy to see it handled in a subsequent PR!

kriswest:
Hi @kyet, hopefully 973 will be through review and merged shortly. I'll update this shortly after.

However, I can't think of anything (in the PR) that would result in in `req.body` being `undefined` (I'm assuming its undefined rather than `null`), but there are reasons that it can be including the client not sending a body (or something in between you and the server removing it). Are you testing on localhost, if so we can discount something on your network.

Happy to look at the request and headers in case it tells us something.

test/testProxyRoute.test.js
```diff
@@ -164,41 +184,271 @@ describe('proxy route helpers', () => {
...
+
+    // try (and fail) to proxy a request to gitlab.com
+    const res2 = await chai
+      .request(proxy.getExpressApp())
+      .get(`/${TEST_GITLAB_REPO.proxyUrlPrefix}/info/refs?service=git-upload-pack`)
+      .set('user-agent', 'git/2.42.0')
+      .set('accept', 'application/x-git-upload-pack-request')
+      .buffer();
+
+    res2.should.have.status(404);
```
jescalada, line 462: This test is failing on my local setup:

```bash
  1) proxy express application
       should be restarted by the api and stop proxying requests for a host (e.g. gitlab.com) when the last project at that host is DELETED via the API:

      AssertionError: expected Response{ domain: null, …(34), …(2) } to have status code 404 but got 200
      + expected - actual

      -200
      +404
      
      at Context.<anonymous> (test/testProxyRoute.test.js:452:22)
      at fulfilled (test/testProxyRoute.test.js:5:58)
      at processTicksAndRejections (node:internal/process/task_queues:105:5)
```

test/testRepoApi.test.js
```diff
@@ -0,0 +1,340 @@
...
+  });
+
+  it('Invalid user push permission on repo', async function () {
+    const isAllowed = await db.isUserPushAllowed(TEST_REPO.url, 'test1234');
+    expect(isAllowed).to.be.false;
+  });
+
+  it('Proxy route helpers should return the proxied origin', async function () {
+    const origins = await getAllProxiedHosts();
+    expect(origins).to.eql([TEST_REPO.host]);
```
jescalada, line 266: This test is also failing on my local setup - perhaps due to missing test data? Or is it my local data that's causing the problem?

```bash
 2) add new repo
       Proxy route helpers should return the proxied origin:

      AssertionError: expected [ 'gitlab.com', 'github.com' ] to deeply equal [ 'github.com' ]
      + expected - actual

       [
      -  "gitlab.com"
         "github.com"
       ]
      
      at Context.<anonymous> (test/testRepoApi.test.js:266:24)
      at fulfilled (test/testRepoApi.test.js:5:58)
      at processTicksAndRejections (node:internal/process/task_queues:105:5)
```

src/proxy/routes/helper.ts
```diff
@@ -0,0 +1,188 @@
...
+ * would produce:
+ * - protocol: https://
+ * - host: someOtherHost.com:8080
+ * - repoPath: /repo.git
+ *
+ * @param {string} url The URL to process
+ * @return {GitUrlBreakdown | null} A breakdown of the components of the URL.
+ */
+export const processGitUrl = (url: string): GitUrlBreakdown | null => {
+  if (url.length > 512) {
```
jescalada: Might want to extract into a constant and replace the other usages in this file 👍🏼

src/service/routes/repo.js
```diff
@@ -1,154 +1,203 @@
...
+
+      // determine if we need to restart the proxy
+      const previousHosts = await getAllProxiedHosts();
+      await db.deleteRepo(_id);
+      const currentHosts = await getAllProxiedHosts();
+
+      if (currentHosts.length < previousHosts.length) {
+        // restart the proxy
+        console.log('Restarting the proxy to remove a host');
+        await theProxy.stop();
```
jescalada, line 131: Could there be any unintended side effects when restarting the proxy? Data loss or denial-of-service for users? 

I understand this would be extremely rare, but supposing a user enters an invalid host, the proxy would still be restarted:

Image

src/ui/views/RepoDetails/RepoDetails.tsx
jescalada, line 1: I'm having an issue when viewing pre-existing repo details on the frontend. The page doesn't load at all, and this error shows in the frontend console:

Image

I found that the culprit is that the `trimTrailingDotGit` function accepts only strings, but the frontend (for some reason) doesn't validate correctly whether the `data` exists before attempting to render it.

The fix was to simply add a check on the `trimTrailingDotGit` and extend it so that it accepts `null` values:

```ts
export const trimTrailingDotGit = (str: string | null): string => {
  if (!str) {
    return '';
  }
 
  const target = '.git';
  if (str.endsWith(target)) {
    // extract string from 0 to the end minus the length of target
    return str.slice(0, -target.length);
  }
  return str;
};
```

src/ui/views/RepoList/Components/NewRepo.tsx
```diff
@@ -146,7 +152,7 @@ const AddRepositoryDialog: React.FC<AddRepositoryDialogProps> = ({ open, onClose
                     onChange={(e) => setProject(e.target.value)}
                     value={project}
                   />
-                  <FormHelperText id='project-helper-text'>GitHub Organization</FormHelperText>
+                  <FormHelperText id='project-helper-text'>Organization or path</FormHelperText>
```
jescalada: Minor improvement here - we could fill in the placeholders with sample values instead. This would be especially useful for the URL field, since users might confuse the trailing `.git` URL with the regular URL from the browser (which they will likely copy-paste).

Image

test/testProxyRoute.test.js
kriswest: Do you have a gitlab repo in your data already? The test is assuming you don't. Does the failure persist if you delete the `.data` dir for the fileDB and re-run (and does it come back if you re-run)?

I'm able to run multiple times with the same data dir and do not get that failure.

... however, I note some oddities to the logging:

```
Restarting the proxy to remove a host
No plugins configured
Found 0 plugin modules
HTTP server closed
Initializing proxy router for origins: '["github.com"]'
        setting up origin: 'github.com'
        setting up catch-all route (github.com) for backwards compatibility
proxy keys registered:  ["/github.com/"]
HTTP Proxy Listening on 8000
processing request URL: '/gitlab.com/gitlab-community/meta.git/info/refs?service=git-upload-pack'
proxy keys registered:  ["/github.com/"]
        using fallback
Action processed: Allowed
    Request URL: /gitlab.com/gitlab-community/meta.git/info/refs?service=git-upload-pack
    Host:        127.0.0.1:58153
    User-Agent:  git/2.42.0
Request resolved to https://github.com/gitlab.com/gitlab-community/meta.git/info/refs?service=git-upload-pack
    ✔ should be restarted by the api and stop proxying requests for a host (e.g. gitlab.com) when the last project at that host is DELETED via the API (694ms)
```

Looks like the request was allowed by the fallback. It should have been blocked. I think the 404 must be coming from *GitHub* and making the test pass my end. Not sure why it fails on yours, but clearly it needs a look.

test/testRepoApi.test.js
kriswest: Yes thats likely some existing repo pointing to gitlab.com. Try deleting you .data folder and re-running a few times.

Perhaps I should add an expect at the start of the test for there to be no repos with that origin?

src/service/routes/repo.js
kriswest: I expect adding a repo at a new origin to be rare - but as @andypols pointed out restarting could interrupt a git operation in progress. The same is true on updating the config of course, where we also restart.

It would be possible to update the proxy without a restart - however, I think we could drop a separate issue to do that after the initial adoption of this as it'll require some additional refactoring. See: https://github.com/finos/git-proxy/pull/1043#issuecomment-3154716580

src/ui/views/RepoList/Components/NewRepo.tsx
kriswest: It turns out the it doesn't matter whether you include the trailing .git or not in the URL as both Github and Gitlab seem to redirect you between the bare URL and .get URL depending on your user agent!

However, a prompt for consistency would still be useful, added:

image

Note I had already added validation that .git is present - I'm now tempted to remove as its not technically required in the protocol (at least I couldn't find it and now think its a convention... so might not be used by other SCM providers??). I have NOT actually removed it however, as I'm not certain of the requirement...

src/ui/views/RepoDetails/RepoDetails.tsx
kriswest: I guess out typing or data validation is letting us down somewhere as that field shouldn't be ever be null/undefined. ah... ui/services is still js, which would explain that.

I've added the slightly shorter: 

```
export const trimTrailingDotGit = (str: string): string => {
  const target = '.git';
  if (str && str.endsWith(target)) {
    // extract string from 0 to the end minus the length of target
    return str.slice(0, -target.length);
  }
  return str;
};
```

Which will return the `null` or `undefined` value (so that we're exposing it) but will not blow up. We retain the correct type despite that so that a later developer working on typing in the ui services will see a type mismatch if the property can be null (when it shouldn't be).

kriswest: Does that allow the page to render on your end @jescalada ?

src/proxy/routes/helper.ts
kriswest: Done 👍

src/service/routes/repo.js
jescalada: Added issue for this: #1146

test/testRepoApi.test.js
jescalada: I removed the extra hosts and it worked 👍🏼 

> Perhaps I should add an expect at the start of the test for there to be no repos with that origin?

This would be helpful in case people try to run the tests locally!

test/testProxyRoute.test.js
jescalada: Removing the GitLab repo fixed this as well!

Might be nice to have `before` and `after` functions to prevent these - some developers might assume there's a problem in their code or the tests themselves.

src/ui/views/RepoDetails/RepoDetails.tsx
jescalada: It works on my end now! 👍🏼

test/testProxyRoute.test.js
kriswest: Yeah, but I'm in two minds as to whether to we should clean up other repos on that domain or just fail out earlier and make it clear what the problem is... Possibly a case that runs first with a clear name like "Check that are no existing Gitlab repos in the database"?. Alternatively just prefix the relevant tests with an expect on a clearly named variable?

I can do either today hopefully, I just have to decide which way to go!

jescalada: Looking at these in more detail, I feel like the simplest choice is to make the reason for failure clear, since most contributors won't be testing with a pre-loaded database to begin with...

jescalada:
@kriswest I've taken another look at the tests, and since the failing ones are "end to end" if anything, it'd be harder than I thought to mock out the database dependency (and it only really makes sense for unit/function tests).

As long as the reason for failure is obvious for contributors, that should be enough for now so we can speed up the release. We can make the tests more robust in another issue (#978 and #1143 are related to this).

kyet:
Hi, @kriswest. I found culprit of the `No body found` problem!

I dumped the packets, I found that my git client was actually sending the contents.

The pattern matching fails in the code below, and therefore the body is not generated.

`proxy/routes/index.ts`

``` ts
const isPackPost = (req: Request) =>
  req.method === 'POST' &&
  // eslint-disable-next-line no-useless-escape
  /^\/[^\/]+\/[^\/]+\.git\/(?:git-upload-pack|git-receive-pack)$/.test(req.url);

const teeAndValidate = async (req: Request, res: Response, next: NextFunction) => {
  if (!isPackPost(req)) return next();
  ..
	try {
	  ..
		(req as any).body = buf;
```

In my case, req.url is `/git.msrd.sk/kwangjin.ko/git-proxy.git/git-receive-pack`.

Upon reviewing the regex, I noticed that there is a change history in the commit below.

29e7d2aa9e3da5f46751351bace1ddcee0e1e65b

You may need to revert or modify the regex changes.

I would like to share one more thing related to this issue. As I mentioned earlier, our company has a firewall issue that prevents us from pushing to github.com, so I tested it at home. The test environment is different, but when I set up a github.com proxy like this, `proxy` fails, but `proxy2` succeeds.

``` ini
[remote "proxy"]
        url = http://localhost:8000/github.com/kyet/git-proxy.git
        fetch = +refs/heads/*:refs/remotes/proxy/*
[remote "proxy2"]
        url = http://localhost:8000/kyet/git-proxy.git
        fetch = +refs/heads/*:refs/remotes/proxy2/*
```

If I think about it in relation to the above issue, when falling back to the default proxy, I **assume** that `req.url` probably doesn't contain `github.com` (e.g., `/kyet/git-proxy.git/git-receive-pack`)

kriswest:
@kyet many thanks for this - I think you're right. The regex is inflexible on the number of path components. Replacing it with:
`^(?>\/[^\/]+)*\/[^\/]+\.git\/(?>git-upload-pack|git-receive-pack)$` should resolve that and be much more flexible - it will be needed to support multiple levels of groups and sub-groups in **GitLab**.

@fabiovincenzi I think I wrote the original regex, but its your [commit](https://github.com/finos/git-proxy/commit/29e7d2aa9e3da5f46751351bace1ddcee0e1e65b) adding it. Do you concur on updating? Was there another reason for the current regex on `isPackPost` that I'm missing? I'm happy to fix in an update to this PR.

jescalada:
@kriswest Is this ready to merge in? If we need to modify anything else, we can perhaps do it in a separate PR so that we can unblock the other PRs that are waiting to solve merge conflicts.

@fabiovincenzi If you think this is a one-liner, then feel free to fix it directly, but I'm keen on merging this ASAP to get the `rc.2` prerelease out and dust off our old PRs!

kriswest:
@jescalada almost - apologies for the delay! I've just resolved some conflicts and bad merge with the fuzzing changes - and raised an issue for another of those tests that can fail randomly (occasionally randomly produces a near valid email address).

I've fixed the `isPackPost` issue just now - it did need fixing, that was introduced by merging main and would have prevented working with GitLab dedicated instances.

I think there may be one more test to look at - I don't like seeing the fallback URLs process repos they shouldn't and make requests onto Github.

> processing request URL: '/gitlab.com/gitlab-community/meta.git/info/refs?service=git-upload-pack'
> proxy keys registered:  ["/github.com/"]
>        using fallback
> Action processed: Allowed
>    Request URL: /gitlab.com/gitlab-community/meta.git/info/refs?service=git-upload-pack
>    Host:        127.0.0.1:61485
>    User-Agent:  git/2.42.0
> Request resolved to https://github.com/gitlab.com/gitlab-community/meta.git/info/refs?service=git-upload-pack

I beleive we're passing the test as github returns a 404 itself. We could ignore that, merge and fix under a different issue - but I'm not happy about it TBH

jescalada:
@kriswest From what I see in the `getRouter` function, deleting the `gitlab.com` host (therefore not available when calling `getAllProxiedHosts`) results in the logic here automatically resolving the host to `github.com` instead.

I suppose we could refactor the `getRequestPathResolver` so that it factors in the URLs domain even if it's not registered? Rather than always prepending the GitHub domain. 

```ts
const getRouter = async () => {
  // eslint-disable-next-line new-cap
  const router = Router();
  router.use(teeAndValidate);

  const originsToProxy = await getAllProxiedHosts();
  const proxyKeys: string[] = [];
  const proxies: RequestHandler[] = [];

  console.log(`Initializing proxy router for origins: '${JSON.stringify(originsToProxy)}'`);

  // we need to wrap multiple proxy middlewares in a custom middleware as middlewares
  // with path are processed in descending path order (/ then /github.com etc.) and
  // we want the fallback proxy to go last.
  originsToProxy.forEach((origin) => {
    console.log(`\tsetting up origin: '${origin}'`);

    proxyKeys.push(`/${origin}/`);
    proxies.push(
      proxy('https://' + origin, {
        parseReqBody: false,
        preserveHostHdr: false,
        filter: proxyFilter,
        proxyReqPathResolver: getRequestPathResolver('https://'), // no need to add host as it's in the URL
        proxyReqOptDecorator: proxyReqOptDecorator,
        proxyReqBodyDecorator: proxyReqBodyDecorator,
        proxyErrorHandler: proxyErrorHandler,
      }),
    );
  });

  console.log('\tsetting up catch-all route (github.com) for backwards compatibility');
  const fallbackProxy: RequestHandler = proxy('https://github.com', {
    parseReqBody: false,
    preserveHostHdr: false,
    filter: proxyFilter,
    proxyReqPathResolver: getRequestPathResolver('https://github.com'),
    proxyReqOptDecorator: proxyReqOptDecorator,
    proxyReqBodyDecorator: proxyReqBodyDecorator,
    proxyErrorHandler: proxyErrorHandler,
  });

  console.log('proxy keys registered: ', JSON.stringify(proxyKeys));

  router.use('/', (req, res, next) => {
    console.log(`processing request URL: '${req.url}'`);
    console.log('proxy keys registered: ', JSON.stringify(proxyKeys));

    for (let i = 0; i < proxyKeys.length; i++) {
      if (req.url.startsWith(proxyKeys[i])) {
        console.log(`\tusing proxy ${proxyKeys[i]}`);
        return proxies[i](req, res, next);
      }
    }
    // fallback
    console.log(`\tusing fallback`);
    return fallbackProxy(req, res, next);
  });
  return router;
};
```

kriswest:
@jescalada yes it'll fall back to the fallback... Which isn't much different from the original setup that assumed GitHub. However, the filter function should still apply and check that the repository actually exists in our data, which is how I was expecting this to be controlled.. I don't think that it was this PR that introduced an issue with the pull chain (although I didn't confirm on main before this merged). Regardless, it's a problem for us as it means any repo at GitHub could be fetched or pulled through the proxy, without the auth check applying.

Clearly we need a test that involves a repo that is not in the data but does exist in GitHub. I'll knock something up this morning.

PR #1046, opened 2025-06-05
fix: correct method for finding ad configuration

06kellyjac:
Could never match any string transformed to lowercase to "activeDirectory"


JamieSlome (approved):
LGTM! 🍰

PR #1048, opened 2025-06-10
feat(experimental): fall back to local spdx data and filter deprecated

06kellyjac:
* Added a copy of SPDX data locally to fall back on
  * This updates rarely so not a huge deal
  * Using a library would be another dependency to review/audit
  * can update with `npm run update-local-licenses`
* Added cli option to allow deprecated licenses, by default filter out
* fixed some tsc/linter warnings
* Use aliases where appropriate


sam-holmes2 (approved):
LGTM :)

JamieSlome (approved):
LGTM! 🍰

JamieSlome:
@06kellyjac - all looks good. Can you just confirm the source of the `licenses.json`?

Issue #1008, opened 2025-05-14
(bug) scanDiff fails on force pushes


coopernetes:
**Describe the bug**
It is common for developers to rebase to modify commits such as add missing DCO or commit signing. GitProxy does not handle this scenario properly and will generate opaque or unclear errors due to failures in the diffing steps. Often, this leads to developers having to start over from scratch with a new branch and/or PR which introduces unnecessary toil.

**To Reproduce**
Will add later once I've gathered some additional logs. To reproduce, simply amend an existing (pushed) commit and attempt to use `git push proxy --force`.

**Expected behavior**
GitProxy can handle pushes made forcefully and get the appropriate diffs from the upstream branches.

kriswest:
We've noted that merging main (e.g. to resolve conflicts that have arrive since you raised a PR) will incorporate incoming commits into diffs as the diff is based on the branch's base I think, rather than the current HEAD of main - as is also the case with both GitHub and GitLab). That adds unnecessary weight to the diff. That can be avoided with a rebase - although you will need re-review the PR's commits as their IDs change on rebase but that may prove to be the lesser evil when lots is being merged into main or PRs hang out for while awaiting review). A force push will be needed to support rebase.

PR #1051, opened 2025-06-11
feat: Add Annotated Tag Push Support

### LINKED ISSUE #986: Add support for pushing tags `git push --tags`
Currently, the `parsePush` action only allows pushing branches. As of `v1.10.0`, pushing a tag throws this error:

Image

The recent TS refactors make it throw an `Invalid commit data` error instead, but without successfully pushing and allowing a review:

Image

Add support for pushing tags (`ref/tags/`) and potentially other things such as notes (`ref/notes/`). This could be done through a plugin or by extending the `parsePush` to handle different kinds of pushes.

Adding tag push support would require deeper changes to how commitData is parsed and used in subsequent actions to check their content. We should also think about the edge case of pushing only a tag without the branch ("hidden" commit data must be validated).

Allowing annotated tags would be great too:
```bash
git tag --annotate foo -m ''
git cat-file tag foo
git push --tags
```

Perhaps this could be done through a plugin instead of editing `parsePush.ts`? Though it seems like the best place to add tag handling.

**Additional context**
None

fabiovincenzi:
Description:
This PR introduces end-to-end support for annotated tag pushes 

- Split the push pipeline into separate `branchPushChain` and `tagPushChain`.
- Update `getChain()` to select the correct chain based on `action.type` and presence of `action.tag`.
- Render a dedicated “Tag Details” table when viewing a tag push instead of the commit diff.

Fixes #986


src/proxy/actions/Action.ts
```diff
@@ -38,7 +45,7 @@ class Action {
   rejected: boolean = false;
   autoApproved: boolean = false;
   autoRejected: boolean = false;
-  commitData?: Commit[] = [];
+  commitData?: CommitData[] = [];
```
jescalada, line 64: I wonder if the `Action.ts` file should have an `actionType: ActionType` enum attribute, so we can check more cleanly whether the action is a Commit, Tag or something else entirely.

src/proxy/processors/push-action/parsePush.ts
jescalada, line 1: Just a heads up: looks like this'll have a lot of conflicts with the upcoming changes... Though, we probably want to merge this as a `v2` feature anyways.

src/proxy/actions/Action.ts
jescalada: I just noticed that there is a `type` attribute already with possible values `'push'` and `'pull'`. Perhaps we can name these more accurately and also convert into an enum?

src/proxy/chain.ts
```diff
@@ -87,8 +102,15 @@ export const getChain = async (
     // This is set to true so that we don't re-insert the plugins into the chain
     pluginsInserted = true;
   }
-  if (action.type === 'pull') return pullActionChain;
-  if (action.type === 'push') return pushActionChain;
+  if (action.type === 'pull') {
```
jescalada: We might want to extract this logic into a `getPushChain` and refactor it a bit to use a `switch` instead. There's a request/bug for force push support #1008 which might require adding more chains.

src/proxy/chain.ts
```diff
@@ -21,6 +20,17 @@ const pushActionChain: ((req: any, action: Action) => Promise<Action>)[] = [
   proc.push.blockForAuth,
 ];
 
+const tagPushChain: ((req: any, action: Action) => Promise<Action>)[] = [
```
jescalada, line 23: Is there a specific reason why the push chain for tags is separate from the regular one? Maybe it's possible to share the push actions somehow.

fabiovincenzi: Good point! The separation exists because tags and regular pushes have fundamentally different validation requirements:

  Tag pushes skip:
  - `checkCommitMessages` & `checkAuthorEmails` (tags don't contain commits to validate)
  - `getDiff`, `gitleaks`, `scanDiff` (no code content to analyze)

  We could have a single chain with conditional checks but I think separate chains are cleaner also for maintainability (easy to add or remove steps)

src/proxy/actions/Action.ts
fabiovincenzi: Yes I think this would help, updating it

src/proxy/chain.ts
jescalada: This makes sense, thanks for the explanation!

PR #1024, opened 2025-05-23
test: improve auth test coverage

jescalada:
Builds on top of #963, to improve test coverage for auth methods.

PR #1061, opened 2025-06-25
fix: allow for auth with activedirectory again

06kellyjac:
~Draft because my commit signing seems to be playing up but should otherwise be ready for review.~
After actually posting the PR commit signatures look fine.

---

Main change - allow for auth with activedirectory again

Current /login is forced to use local auth.
This change resolves this regression in a backwards compatible manner while still supporting alternative auth like OIDC.

Auth methods compatible with /login are filtered out from the git-proxy config, if there isn't one /login is essentially disabled to avoid throwing.
If there is multiple methods it'll prioritize the first.
In the future we can support multiple user & password login methods simultaneously with relevant TODOs added.

other fixes:

- **fix: make loadFromGit non-interactive**
- **test: move valid url missing git repo test under our control**


06kellyjac:
cc: @jescalada to double check this doesn't impact OIDC & also may be relevant to #1024

src/service/routes/auth.js
```diff
@@ -23,25 +26,59 @@ router.get('/', (req, res) => {
...
+  '/login',
+  (req, res, next) => {
+    const authType = getLoginStrategy();
+    if (authType === null) {
+      res.status(403).send('Username and Password based Login is not enabled at this time').end();
+      return;
+    }
+    console.log('going to auth with', authType);
+    return passport.authenticate(authType)(req, res, next);
+  },
```
06kellyjac, line 61: This is a callback/middleware to ensure `getLoginStrategy` and more importantly `getAuthMethods` gets the correct proxy config after it's read the user's configuration.

In reality by the time this particular JS file is evaluated the file is already read but this is the more robust method

src/service/routes/auth.js
```diff
@@ -23,25 +26,59 @@ router.get('/', (req, res) => {
...
+// TODO: if providing separate auth methods, inform the frontend so it has relevant UI elements and appropriate client-side behavior
+router.post(
+  '/login',
+  (req, res, next) => {
+    const authType = getLoginStrategy();
+    if (authType === null) {
+      res.status(403).send('Username and Password based Login is not enabled at this time').end();
+      return;
+    }
+    console.log('going to auth with', authType);
```
06kellyjac, line 59: could remove this log, was mostly there for testing

src/service/routes/auth.js
```diff
@@ -23,25 +26,59 @@ router.get('/', (req, res) => {
...
+
+// TODO: provide separate auth endpoints for each auth strategy or chain compatibile auth strategies
+// TODO: if providing separate auth methods, inform the frontend so it has relevant UI elements and appropriate client-side behavior
+router.post(
+  '/login',
+  (req, res, next) => {
+    const authType = getLoginStrategy();
+    if (authType === null) {
+      res.status(403).send('Username and Password based Login is not enabled at this time').end();
+      return;
```
06kellyjac, line 57: This should account for cases where only OIDC auth is enabled without just calling into passport and throwing

PR #1086, opened 2025-07-03
feat: mongo connection string & cookie secret from env vars

### LINKED ISSUE #690: making COOKIE_SECRET an env var

coopernetes:
Closes #690

- allow for the setting of configuration for MongoDB database
  connection string as well as cookie secret via environment
  variables as an alternative to JSON-based configuration

Useful in environments that split config between secrets (ie.
stored in a secret manager) and plain app configurations such
as commitConfig, attestation, etc. that may not be considered
sensitive & wouldn't be appropriate in a secret manager.

### DIFF
--- a/src/config/env.ts
+++ b/src/config/env.ts
@@ -3,6 +3,8 @@ export type ServerConfig = {
   GIT_PROXY_HTTPS_SERVER_PORT: string | number;
   GIT_PROXY_UI_HOST: string;
   GIT_PROXY_UI_PORT: string | number;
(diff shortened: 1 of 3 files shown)


JamieSlome (approved):
LGTM! 🍰

PR #1093, opened 2025-07-07
fix: prevent non-admin users changing another user's gitAccount

andypols:
# Summary

A recent penetration test identified a security issue in the `/api/v1/auth/gitAccount` endpoint:

- Users can add or update their own GitHub/GitLab account names.
- Admins can add or update GitHub/GitLab account names for any user.
- However, **the API does not enforce this restriction** — any user can change the Git account name for any other user, regardless of their permissions.

## Root Cause

The API only checks that the requester is authenticated; it does not verify whether the user is updating their own account or has admin privileges.

## Fix

This PR introduces a guard clause that ensures the requesting user is either:

- Updating their own Git account
- Has admin privileges.

This change brings the backend access control in line with the intended behaviour already enforced in the UI.

I’ve added a new `auth.test.js`. I chose not to expand the existing `testLogin.test.js` because I wanted to test at a lower level, where I can directly assert that the database is updated as expected.

While the current endpoint lives under /auth, updating a user's Git account may conceptually belong on the /user endpoint. However, this change is out of scope for this fix and has been deferred.

### DIFF
--- a/src/service/routes/auth.js
+++ b/src/service/routes/auth.js
@@ -136,7 +136,7 @@ router.post('/gitAccount', async (req, res) => {
       }
 
       const reqUser = await db.findUser(req.user.username);
-      if(username !== reqUser.username && !reqUser.admin) {
+      if (username !== reqUser.username && !reqUser.admin) {
         res.status(403).send('Error: You must be an admin to update a different account').end();
         return;
       }
--- a/src/service/routes/auth.js
+++ b/src/service/routes/auth.js
@@ -135,8 +135,12 @@ router.post('/gitAccount', async (req, res) => {
         return;
       }
 
+      const reqUser = await db.findUser(req.user.username);
+      if(username !== reqUser.username && !reqUser.admin) {
+        res.status(403).send('Error: You must be an admin to update a different account').end();
+        return;
+      }
       const user = await db.findUser(username);
-
       console.log('Adding gitAccount' + req.body.gitAccount);
       user.gitAccount = req.body.gitAccount;
       db.updateUser(user);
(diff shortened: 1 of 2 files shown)


test/services/routes/auth.test.js
```diff
@@ -0,0 +1,116 @@
...
+      expect(
+        updateUserStub.calledOnceWith({
+          username: 'bob',
+          displayName: 'Bob Woodward',
+          email: 'bob@example.com',
+          admin: false,
+          gitAccount: 'UPDATED_GIT_ACCOUNT',
+        }),
+      ).to.be.true;
+    });
```
jescalada, line 114: Maybe we could add this test to make it clear that non-admins should also be able to update their username:

```suggestion
    });
    
    it('POST /gitAccount allows non-admin user to update their own gitAccount', async () => {
      const updateUserStub = sinon.stub(db, 'updateUser').resolves();

      const res = await chai.request(newApp('bob')).post('/auth/gitAccount').send({
        username: 'bob',
        gitAccount: 'UPDATED_GIT_ACCOUNT',
      });

      expect(res).to.have.status(200);
      expect(
        updateUserStub.calledOnceWith({
          username: 'bob',
          displayName: 'Bob Woodward',
          email: 'bob@example.com',
          admin: false,
          gitAccount: 'UPDATED_GIT_ACCOUNT',
        }),
      ).to.be.true;
    });

```

src/service/routes/auth.js
```diff
@@ -135,8 +135,12 @@ router.post('/gitAccount', async (req, res) => {
         return;
       }
 
+      const reqUser = await db.findUser(req.user.username);
+      if(username !== reqUser.username && !reqUser.admin) {
```
jescalada: Formatting issue (may need to run `npm run format`:

```suggestion
      if (username !== reqUser.username && !reqUser.admin) {
```

jescalada (approved):
LGTM! 👍🏼

PR #1094, opened 2025-07-08
fix: only trim trailing .git not any match

06kellyjac:
Currently .git is being replaced in repoNames but `replace` only catches the first match from left to right

```js
> "hello.gitfrom.git".replace(".git", "")
'hellofrom.git'
```

Git repos can have .git included in the name so to correctly get the repoName without a trailing `.git`
it should only be removed from the back (assuming it's there).

Trailing `.git` being trimmed by github:
image

`.git` being allowed mid repo-name:
image

`.git` being allowed at the start of a repo name
image

This code checks the string ends with `.git`. If it does it returns the string up to but not including `.git`, otherwise it returns the whole string.

I've added a new test just for this edgecase.

- **fix: only trim trailing .git not any match**
- **fix: trim ref/heads/ only from the prefix**
- **test: improve visibility of throws in approve/reject push test**
- **test: add approve/reject push test for .git inside name edgecase**

### DIFF
--- a/src/db/helper.ts
+++ b/src/db/helper.ts
@@ -13,3 +13,11 @@ export const trimTrailingDotGit = (str: string): string => {
   return str;
 };
 
+export const trimPrefixRefsHeads = (str: string): string => {
+  const target = 'refs/heads/';
+  if (str.startsWith(target)) {
+    // extract string from the end of the target to the end of str
+    return str.slice(target.length);
+  }
+  return str;
+};
--- a/src/db/helper.ts
+++ b/src/db/helper.ts
@@ -3,3 +3,13 @@ export const toClass = function (obj: any, proto: any) {
   obj.__proto__ = proto;
   return obj;
 };
+
+export const trimTrailingDotGit = (str: string): string => {
+  const target = '.git';
+  if (str.endsWith(target)) {
+    // extract string from 0 to the end minus the length of target
+    return str.slice(0, -target.length);
+  }
+  return str;
+};
+
--- a/src/proxy/processors/push-action/checkUserPushPermission.ts
+++ b/src/proxy/processors/push-action/checkUserPushPermission.ts
@@ -1,11 +1,20 @@
 import { Action, Step } from '../../actions';
 import { getUsers, isUserPushAllowed } from '../../../db';
+import { trimTrailingDotGit } from '../../../db/helper';
 
 // Execute if the repo is approved
 const exec = async (req: any, action: Action): Promise<Action> => {
   const step = new Step('checkUserPushPermission');
 
-  const repoName = action.repo.split('/')[1].replace('.git', '');
+  const repoSplit = trimTrailingDotGit(action.repo.toLowerCase()).split('/');
+  // we expect there to be exactly one / separating org/repoName
+  if (repoSplit.length != 2) {
+    step.setError('Server-side issue extracting repoName');
+    action.addStep(step);
+    return action;
+  }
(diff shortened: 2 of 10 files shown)


src/proxy/processors/push-action/checkUserPushPermission.ts
```diff
@@ -1,11 +1,20 @@
...
 
 // Execute if the repo is approved
 const exec = async (req: any, action: Action): Promise<Action> => {
   const step = new Step('checkUserPushPermission');
 
-  const repoName = action.repo.split('/')[1].replace('.git', '');
+  const repoSplit = trimTrailingDotGit(action.repo.toLowerCase()).split('/');
+  // we expect there to be exactly one / separating org/repoName
+  if (repoSplit.length != 2) {
+    step.setError('Server-side issue extracting repoName');
```
jescalada, line 12: I thought we could use a more descriptive error here, but I don't think this would happen under *normal* usage since Git doesn't allow slashes other than for the repo path.

src/db/helper.ts
```diff
@@ -3,3 +3,21 @@ export const toClass = function (obj: any, proto: any) {
...
+export const trimTrailingDotGit = (str: string): string => {
+  const target = '.git';
+  if (str.endsWith(target)) {
+    // extract string from 0 to the end minus the length of target
+    return str.slice(0, -target.length);
+  }
+  return str;
+};
+
+export const trimPrefixRefsHeads = (str: string): string => {
```
jescalada, line 16: Wonder if we could add a test to cover the missing parts in this function? Just to improve coverage 😃

06kellyjac:
Rebased

src/proxy/processors/push-action/checkUserPushPermission.ts
06kellyjac: yeah, it's essentially unreachable with correct code.
Just to catch the case where some logic changes further down the chain where the wrong data is sent & avoid trying to access outside of bounds on the array created by split.

src/db/helper.ts
06kellyjac: Yeah that's a good point. These are very testable little functions. I'll whip something up now.

06kellyjac: Added test cases:

```
λ npm run test -- -g "db helpers"

> @finos/git-proxy@1.19.0 test
> NODE_ENV=test ts-mocha './test/**/*.test.js' --exit -g db helpers

Configuration sources are disabled
(node:140758) [MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of file:///home/xyz/projects/work/git-proxy/test/ConfigLoader.test.js is not specified and it doesn't parse as CommonJS.
Reparsing as ES module because module syntax was detected. This incurs a performance overhead.
To eliminate this warning, add "type": "module" to /home/xyz/projects/work/git-proxy/package.json.
(Use `node --trace-warnings ...` to show where the warning was created)


  db helpers
    trimPrefixRefsHeads
      ✔ removes `refs/heads/`
      ✔ removes only one `refs/heads/`
      ✔ removes only the first `refs/heads/`
      ✔ handles empty string
      ✔ doesn't remove `refs/heads`
      ✔ doesn't remove `/refs/heads/`
    trimTrailingDotGit
      ✔ removes `.git`
      ✔ removes only one `.git`
      ✔ removes only the last `.git`
      ✔ handles empty string
      ✔ doesn't remove just `git`


  11 passing (11ms)
```

PR #1095, opened 2025-07-08
fix: prevent DOS when checking an unknown repo

andypols:
# Summary

This PR fixes a potential denial-of-service (DoS) vulnerability:

When pushing to an unknown repository, the MongoDB implementation throws a TypeError due to attempting to access properties on a null object:

```
console.log(repo.users.canPush);
// TypeError: Cannot read properties of null (reading 'users')
```

## Root Cause

The file-based database implementation correctly checks for the existence of a repository before accessing its fields. However, the MongoDB implementation does not.

Specifically, `checkUserPushPermission` calls `isUserPushAllowed`, which assumes the repository exists. If the repository is not found, accessing its properties throws a TypeError and stops the service.

## Fix

This PR addresses the issue by:

- Adding  a guard clause in the MongoDB implementation of `isUserPushAllowed` to handle missing repositories safely.

- Adds a unit test to verify behaviour when the repository does not exist.

### DIFF
--- a/src/db/mongo/repo.ts
+++ b/src/db/mongo/repo.ts
@@ -79,19 +79,19 @@ export const deleteRepo = async (name: string) => {
 export const isUserPushAllowed = async (name: string, user: string) => {
   name = name.toLowerCase();
   user = user.toLowerCase();
+  console.log(`checking if user ${user} can push to ${name}`);
   return new Promise(async (resolve) => {
     const repo = await exports.getRepo(name);
-    console.log(repo.users.canPush);
-    console.log(repo.users.canAuthorise);
-
-    if (repo.users.canPush.includes(user) || repo.users.canAuthorise.includes(user)) {
-      resolve(true);
-    } else {
+    if( !repo ) {
+      console.log(`repo ${name} not found`);
       resolve(false);
+      return;
     }
+    resolve(repo.users.canPush.includes(user) || repo.users.canAuthorise.includes(user));
   });
 };
 
+// not used in the codebase, but kept for compatibility
 export const canUserApproveRejectPushRepo = async (name: string, user: string) => {
   name = name.toLowerCase();
   user = user.toLowerCase();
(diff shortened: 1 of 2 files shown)


jescalada (approved):
LGTM! 👍🏼

I looked around the `mongo` handlers and it seems there aren't any other similar bugs.

src/db/mongo/repo.ts
```diff
@@ -79,19 +79,19 @@ export const deleteRepo = async (name: string) => {
...
-
-    if (repo.users.canPush.includes(user) || repo.users.canAuthorise.includes(user)) {
-      resolve(true);
-    } else {
+    if( !repo ) {
+      console.log(`repo ${name} not found`);
       resolve(false);
+      return;
     }
+    resolve(repo.users.canPush.includes(user) || repo.users.canAuthorise.includes(user));
```
kriswest:
```suggestion
    resolve(repo.users?.canPush.includes(user) || repo.users?.canAuthorise.includes(user));
```

The DB clients _should_ make sure that the users element exists (although I think I had to make sure that was happening in #1043, after stumbling over a case where it did not). Doesn't hurt to be defensive here.

src/db/mongo/repo.ts
```diff
@@ -79,19 +79,19 @@ export const deleteRepo = async (name: string) => {
 export const isUserPushAllowed = async (name: string, user: string) => {
   name = name.toLowerCase();
   user = user.toLowerCase();
+  console.log(`checking if user ${user} can push to ${name}`);
```
kriswest: Note that #1043 moves this functions into src/db/index.ts so that it is not duplicated between the clients.

test/db/mongo/repo.test.js
```diff
@@ -0,0 +1,52 @@
+const chai = require('chai');
```
kriswest: This will be orphaned when #1043 merges as the function under test moves to the src/db/index and applies over the underlying db client configured.

There's some, but not total, overlap between these tests and #1043 -these are tighter unit tests on `isUserPushAllowed` and are worth having.

kriswest (changes requested):
Pls rebase and assess against current state of main

andypols:
@finos/git-proxy-maintainers I've merged with latest changes and all is left following @kriswest checking is the test.  Please can you review and merge is all ok.

Thanks

kriswest (approved):
LGTM

Issue #1023, opened 2025-05-21
Upgrade Material UI to v5


jescalada:
**Is your feature request related to a problem? Please describe.**
Since v4 is [deprecated](https://mui.com/material-ui/migration/migration-v4/), it would be good to upgrade MUI to v5. As mentioned in the link, this would improve UI performance and provide the latest React support.

**Describe the solution you'd like**
Maybe an incremental update would be best? There are a quite a few breaking changes in the components, some of which are not handled in the automatic upgrade guide.

**Describe alternatives you've considered**
None

**Additional context**
This issue was originally brought up by @sam-holmes2 in #963:

> I received the following deprecated version number warning from npm:
>
> npm warn deprecated @material-ui/styles@4.11.5: Material UI v4 doesn't receive active development since September 2021. See the guide https://mui.com/material-ui/migration/migration-v4/ to upgrade to v5.
> npm warn deprecated @material-ui/core@4.12.4: Material UI v4 doesn't receive active development since September 2021. See the guide https://mui.com/material-ui/migration/migration-v4/ to upgrade to v5.

jescalada:
### Update

I tried upgrading to MUI v5 and I started getting these frontend errors:

<img width="1418" height="1567" alt="Image" src="https://github.com/user-attachments/assets/077d8c5c-1196-49ae-b28f-16864257be6f" />

I think most of these can be fixed by refactoring the `makeStyles` calls and style files with the newer `styled` components. Unfortunately, there are 33 of those files in total, and the `styled` version is essentially the same thing, except that it clutters the component files (`.tsx`) instead. Perhaps the best solution here is a complete refactor of the UI styling, maybe through something cleaner and more scalable like Tailwind...

PR #1096, opened 2025-07-09
feat: JWT `apiAuthentication` UI integration

### LINKED ISSUE #1091: Implement JWT API authorization support for UI
Currently, the UI doesn't support the `apiAuthentication` feature for JWT (and other auth methods). When I first implemented the JWT API auth, I thought of it as an API-only feature (to be used with automation tools).

We should add support for JWT in the UI, so that any user can improve their backend security while also letting them use the UI as normal. 

This is what is currently shown in the UI when activating the JWT `apiAuthentication`:

Image

We would like to fetch the repos/users/etc. WITHOUT bypassing the JWT token check in the API. This means we need an input for the JWT, as well as handling expired/invalid tokens (may reuse the existing error handling code).

Bypassing the JWT check for UI requests (not a good idea, as the whole point of JWT is an extra layer of security).

**Additional context**
Related PRs: #967 #977

jescalada:
Fixes #1091.

This PR adds a Settings page that allows configuring the JWT token to authenticate UI requests to the API:

image

Previously, requests were bypassing the JWT check if they came from an authenticated user, and being completely blocked when not logged in.

This PR also improves the error messages shown when JWT authentication fails:

##### Missing token
image

##### Bad JWT format
image

##### Expired JWT
image


JamieSlome:
@jescalada - thanks for the PR ❤️ 

Perhaps it makes sense to display the error message outside of the table in a pop-up? Open until closed by the user?

jescalada:
I modified this so that it uses one of the predefined error components (and got the message out of the tabs box):

image

There's still an issue with a non-descriptive error when JWT is enabled but the config is missing.

jescalada:
I've tried to find better-looking error displays (such as custom-styled Snackbars, and other components), but it seems MUI v4 doesn't support some of the best ones.

For example, in later versions of MUI we can use these [Alert components](https://mui.com/material-ui/react-alert/) or combine them with Snackbars to make it look better.

There's this issue for upgrading MUI to v5 #1023 which I'll take a look at soon. It might be worth modernizing the frontend a bit once the dependency upgrades are done. (Note: Upgrading directly to v7 is not possible as each new major version introduces various breaking changes - plus the automatic migration tools only go up one major version at a time)

src/ui/services/auth.js
```diff
@@ -20,3 +22,34 @@ export const getUserInfo = async () => {
     return null;
   }
 };
+
+/**
+ * Gets the Axios config for the UI
+ * @return {Object} The Axios config
+ */
+export const getAxiosConfig = () => {
+  console.log('getAxiosConfig', getCookie('csrf'), localStorage.getItem('ui_jwt_token'));
```
dcoric: Remove logging to prevent leaking sensitive info to the console

src/ui/components/Typography/Danger.jsx
```diff
@@ -8,7 +8,7 @@ const useStyles = makeStyles(styles);
 export default function Danger(props) {
   const classes = useStyles();
   const { children } = props;
-  return <div className={classes.defaultFontStyle + ' ' + classes.dangerText}>{children}</div>;
+  return <div className={classes.primaryText + ' ' + classes.dangerText}>{children}</div>;
```
dcoric: This one is just a suggestion - consider using [clsx](https://www.npmjs.com/package/clsx) or similar utils for class construction

src/ui/components/Navbars/DashboardNavbarLinks.tsx
```diff
@@ -51,21 +51,17 @@ const DashboardNavbarLinks: React.FC = () => {
...
-        },
-      );
-
-      if (!response.data.isAuth && !response.data.user) {
-        setAuth(false);
-        navigate(0);
-      }
+        getAxiosConfig(),
+      )
+      .then((res) => {
```
dcoric: I'm not sure if we should combine `async/await` with `.then` promise style. For readability, it is usually recommended to stick with one or another, so I'm curious if there's a specific reason for the mix here?

src/ui/services/auth.js
jescalada: Whoops, I missed this one!

src/ui/components/Typography/Danger.jsx
jescalada: Added `clsx` and replaced the previous `classnames` calls too, since `clsx` is more lightweight.

src/ui/components/Navbars/DashboardNavbarLinks.tsx
jescalada: Just an oversight during my refactoring 😅

PR #1109, opened 2025-07-17
test: improve repo DB tests and CheckRepoInAuthList tests

andypols:
# Summary

This PR fixes a bug where pushes to whitelisted repositories were rejected because the entered repository name didn’t exactly match the actual repository URL.

It’s related to the broader feature in PR #1043 (support for non-GitLab repos), but has been separated into a smaller, focused fix to unblock the security team, who are currently running penetration tests using GitLab. It also improves efficiency by loading a repository directly from the database instead of scanning all repositories on each commit.

## Root Cause

The code previously assumed that repository names followed a strict `${project}/${name}` naming convention (e.g., `finos/git-proxy`). However, this assumption doesn't hold in our case. For example:

* Project: `Security & Compliance`
* Name: `GitProxy`
* Actual repo URL: `security-and-compliance/git-proxy`

Since the naming did not match the expected format, the push failed.

## Fix

This PR removes the naming convention assumption and instead matches repositories based on their full URL. Specifically:

* Introduces `getRepoByUrl` (with unit tests) for both file-based and MongoDB-backed repo sources.
* Simplifies `checkRepoInAuthorisedList` by directly loading a matching repo via its URL instead of looping through all repos.
* Updates the checkRepoInAuthorisedList test.  I also moved it to be with the other processor action tests.
* Adds support for handling mistakenly entered .git suffixes in repo URLs (which we also did by mistake).


test/db/mongo/repo.test.js
```diff
@@ -0,0 +1,85 @@
...
+  afterEach(function () {
+    sinon.restore();
+  });
+
+  describe('getRepo', () => {
+    it('should get the repo using the name', async () => {
+      const repoData = {
+        name: 'sample',
+        users: { canPush: [] },
+        url: 'http://example.com/sample-repo',
```
kriswest: I think the URLs should be to the git endpoint and should include the .git. Non-github/gitlab repos will not necessarily have a webpage (which is a function of those particular SCM providers), but they will have the endpoint URL. Both Github and Gitlab implement redirects for browsers into the webpage, with git clients staying on git endpoint url

kriswest: #1043 includes validation that the URL includes the .git is present in the URL and the current example authorisedRepo in the example config includes it: https://github.com/finos/git-proxy/blob/c7d94a994fcbb39f403ac4282a4f70226b9b4d6f/proxy.config.json#L17

kriswest (commented):
I could do with some clarification of why the repository name didn't exactly match. Until #1043 is merged, git-proxy can't support multiple forks of the same repo (project or organisation name is different, repo name the same). This PR addresses one or two of those but, by including only a subset of the changes from #1043 it doesn't give you the ability work with multiple forks through the API or UI.

If the project is a going to accept #1043 soon (and on the v1 line - as seemed to be the outcome from the last meeting) then I would question if this this should go in first.

@andypols is any part of this PR NOT included in #1043 that I should be considering adding to it?

test/db/mongo/repo.test.js
andypols: Good catch on the .git suffix.  I will change it.

Its a bit ambiguous in the UI - is _'GitHub Repository URL'_ the `'https://github.com/finos/git-proxy'` or `'https://github.com/finos/git-proxy.git'`.  When testing different people have been using different ones! I think the UI should reject ones that don't comply and provide an example as most people won't see the config. I can do this in a different PR.

andypols:
> I could do with some clarification of why the repository name didn't exactly match. Until #1043 is merged, git-proxy can't support multiple forks of the same repo (project or organisation name is different, repo name the same). This PR addresses one or two of those but, by including only a subset of the changes from #1043 it doesn't give you the ability work with multiple forks through the API or UI.

The discrepancy here is due to testing against GitLab, which allows arbitrary display names for repos. For example, I have a repo named Git Proxy Infra, but the actual repo URL in the git push is git-proxy-infra. So the display name and the slug used in the push don't exactly match — which causes issues with how git-proxy currently resolves repositories.  People tend to enter the name they see in the UI.

jescalada:
@kriswest We're waiting for a FINOS admin to help us unblock the PRs we talked about in the call. Maintainers cannot actually merge them directly through GitHub so we would need extended permissions or coordinate with an admin to get them in...

If it's not too much of a hassle to keep maintaining your PRs, I can take a look at #1109, #1110 and the other fix PRs by @andypols and make a patch release so their org can keep pentesting smoothly (which I believe is valuable to all of us).

As for my PRs, a quick look would be appreciated but they are pretty much ready to merge barring the permissions issue 🙂

test/db/mongo/repo.test.js
jescalada: Would love to see the UI ambiguities fixed in another PR! 👍🏼

I just did some testing with that, it seems that the difference is minor since the "git clone URL copy" section gets the right URL after refresh, and the URL on the repo page differs but GitHub and GitLab are able to correctly redirect if the link ends in `.git`. I think the simplest fix would be a grey default text with a .git, and potentially just auto-append it if it's missing.

test/db/file/repo.test.js
```diff
@@ -0,0 +1,60 @@
...
+        url: 'https://github.com/finos/git-proxy.git',
+      };
+
+      sandbox.stub(repoModule.db, 'findOne').callsFake((query, cb) => cb(null, repoData));
+
+      const result = await repoModule.getRepoByUrl('https://github.com/finos/git-proxy.git');
+
+      expect(repoModule.db.findOne.calledWith(sinon.match({ url: 'https://github.com/finos/git-proxy.git'}))).to.be.true;
+      expect(result).to.equal(repoData);
+    });
```
jescalada, line 65: A few more tests for the `file` db (especially for the "ignore case" route):

```suggestion
    });
    
    it('should get the repo using the url, ignoring the case', async () => {
      const repoData = {
        name: 'sample',
        users: { canPush: [] },
        url: 'https://github.com/finos/git-proxy.git',
      };

      sandbox.stub(repoModule.db, 'findOne').callsFake((query, cb) => cb(null, repoData));

      const result = await repoModule.getRepoByUrl('https://github.com/Finos/Git-Proxy.git');
      expect(result).to.equal(repoData);
      expect(repoModule.db.findOne.calledWith(sinon.match({ url: 'https://github.com/finos/git-proxy.git' }))).to.be.true;
    });

    it('should return null if the repo is not found', async () => {
      sandbox.stub(repoModule.db, 'findOne').callsFake((query, cb) => cb(null, null));

      const result = await repoModule.getRepoByUrl('https://github.com/finos/missing-repo.git');
      expect(result).to.be.null;
      expect(repoModule.db.findOne.calledWith(sinon.match({ url: 'https://github.com/finos/missing-repo.git' })),
      ).to.be.true;
    });

    it('should reject if the database returns an error', async () => {
      sandbox.stub(repoModule.db, 'findOne').callsFake((query, cb) => cb(new Error('DB error')));
    
      try {
        await repoModule.getRepoByUrl('https://github.com/finos/git-proxy.git');
        expect.fail('Expected promise to be rejected');
      } catch (err) {
        expect(err.message).to.equal('DB error');
      }
    });
```

test/db/mongo/repo.test.js
```diff
@@ -0,0 +1,69 @@
...
+      };
+      repoCollection.findOne.resolves(repoData);
+
+      const result = await getRepoByUrl('https://github.com/Finos/Git-Proxy.git');
+      expect(result).to.equal(repoData);
+      expect(connectionStub.calledWith('repos')).to.be.true;
+      expect(
+        repoCollection.findOne.calledWith({ url: { $eq: 'https://github.com/finos/git-proxy.git' } }),
+      ).to.be.true;
+    });
```
jescalada, line 53: Oddly, the following test which passes on the `file` db test, will fail here:

```suggestion
    });
    
    it('should handle invalid URLs gracefully', async () => {
      const result = await getRepoByUrl('invalid-url');
      expect(result).to.be.null;
    });
```

This returns, for some reason

```
{
  name: 'sample',
  users: { canPush: [] },
  url: 'https://github.com/finos/git-proxy.git'
}
```

andypols:
@jescalada - thanks for the test suggestion - I've added them

kriswest:
@andypols this can be updated now - although the changes to how DBs are looked up in the DB by URL may have eliminated the need. However, we'd still take test changes.

kriswest:
> @kriswest I think there is a bug in your implementation of getRepoByUrl
> 
> ```
> export const getRepoByUrl = async (repoUrl: string): Promise<Repo | null> => {
>   const collection = await connect(collectionName);
>   const doc = collection.findOne({ name: { $eq: repoUrl.toLowerCase() } });
>   return doc ? toClass(doc, Repo.prototype) : null;
> };
> ```
> 
> Shouldn't this match on by `url` and not `name`?

I'll look into this - there were so many rounds of dealing with conflicts on this PR that there may be a merge issue there.

kriswest:
@andypols nope looks like a straight-up mistake, I'm not setup to test against mongo on my machine and, although it looks like there was a start on creating a mongo testing setup in the repo it was incomplete the last time I checked. Thats clearly something we need to change @finos/git-proxy-maintainers.

I've raised a bug fix PR (#1167), but can't test it until I have time to get some setup done here. If you have a moment to check we could move that forward today.

andypols:
@finos/git-proxy-maintainers : I have updated this, please can you review and merge if happy.  All that is left is a slight cleanup of the checkRepoInAuthorisedList code and additional tests

Thanks

test/db/file/repo.test.js
```diff
@@ -0,0 +1,87 @@
...
+
+  describe('getRepo', () => {
+    it('should get the repo using the name', async () => {
+      const repoData = {
+        name: 'sample',
+        users: { canPush: [] },
+        url: 'http://example.com/sample-repo.git',
+      };
+
+      sandbox.stub(repoModule.db, 'findOne').callsFake((query, cb) => cb(null, repoData));
```
kriswest, line 24: I worry that we don't test the query construction against the dependency when we mock the DB calls... however, there are a many other tests that do roundtrip through the db so it is covered elsewhere.

kriswest: But essentially we end up mocking the only interesting line in the function under test.

test/db/file/repo.test.js
```diff
@@ -0,0 +1,87 @@
...
+        url: 'https://github.com/finos/git-proxy.git',
+      };
+
+      sandbox.stub(repoModule.db, 'findOne').callsFake((query, cb) => cb(null, repoData));
+
+      const result = await repoModule.getRepoByUrl('https://github.com/finos/git-proxy.git');
+      expect(result).to.deep.equal(repoData);
+    });
+
+    it('should get the repo using the url, stripping off the .git', async () => {
```
kriswest: Where does the stripping off the .git come in now?

test/db/file/repo.test.js
```diff
@@ -0,0 +1,87 @@
...
+
+      expect(
+        repoModule.db.findOne.calledWith(
+          sinon.match({ url: 'https://github.com/finos/git-proxy.git' }),
+        ),
+      ).to.be.true;
+      expect(result).to.deep.equal(repoData);
+    });
+
+    it('should return null if the repo is not found', async () => {
```
kriswest, line 44: If we mock the function so that it always returns null, what are we testing? I guess pass through of the null and that it doesn't blow up when it happens ✔️.

test/db/mongo/repo.test.js
```diff
@@ -0,0 +1,55 @@
...
+  findOne: sinon.stub(),
+};
+
+const connectionStub = sinon.stub().returns(repoCollection);
+
+const { getRepo, getRepoByUrl } = proxyqquire('../../../src/db/mongo/repo', {
+  './helper': { connect: connectionStub },
+});
+
+describe('MongoDB', () => {
```
kriswest, line 15: Great to get some coverage on this. A couple of additional tests from the file DB set might be useful, e.g. 'should reject if the database returns an error' and 'should return null if the repo is not found' perhaps.

test/testCheckRepoInAuthList.test.js
```diff
@@ -1,52 +0,0 @@
...
-    expect(result.error).to.be.false;
-  });
-
-  it('Should set ok=false if not in authorised', async () => {
-    const action = new actions.Action('123', 'type', 'get', 1234, TEST_NON_EXISTENT_REPO.url);
-    const result = await processor.exec(null, action);
-    expect(result.error).to.be.true;
-  });
-
-  describe('fuzzing', () => {
```
kriswest, line 40: @jescalada are you ok with removing this fuzz test?

jescalada: I thought perhaps @andypols had a reason to remove (perhaps it no longer reliably passes for some reason). Would be great to know why it was removed 🤔

test/db/file/repo.test.js
andypols: Thats fair - it should test the query.  I

test/db/file/repo.test.js
```diff
@@ -0,0 +1,87 @@
...
+
+      const result = await repoModule.getRepoByUrl('https://github.com/finos/git-proxy.git');
+
+      expect(
+        repoModule.db.findOne.calledWith(
+          sinon.match({ url: 'https://github.com/finos/git-proxy.git' }),
+        ),
+      ).to.be.true;
+      expect(result).to.deep.equal(repoData);
+    });
```
kriswest: This has become identical to the previous test and can be removed.

```suggestion
```

test/db/file/repo.test.js
kriswest: its was a vetigal name from an older test that made more sense when retrieving by name.

test/db/file/repo.test.js
```diff
@@ -0,0 +1,69 @@
...
+        url: 'https://github.com/finos/git-proxy.git',
+      };
+
+      sandbox.stub(repoModule.db, 'findOne').callsFake((query, cb) => cb(null, repoData));
+
+      const result = await repoModule.getRepoByUrl('https://github.com/finos/git-proxy.git');
+      expect(result).to.deep.equal(repoData);
+    });
+
+
```
kriswest:
```suggestion

```

test/testCheckRepoInAuthList.test.js
kriswest: @jescalada I think it might just have been a merge issue - do you want to revert the change and reintroduce that test?

test/db/mongo/repo.test.js
kriswest: Can be added later for more coverage

test/db/file/repo.test.js
kriswest: Resolving this thread so we can merge - there are other tests in /test/testDB.test.js that cover the queries in fileDB.

test/testCheckRepoInAuthList.test.js
kriswest: (you'll have to resolve this conversation to merge the PR)

kriswest: @jescalada to reintroduce this test

PR #1110, opened 2025-07-17
fix: checkUserPushPermission push failure on name

andypols:
# Summary

This PR builds on #1109 to address a bug in `checkUserPushPermission`. The bug caused permission checks to fail when the repository could not be found by name. It also incorrectly assumed that repository names are globally unique, which prevented users from forking the same repository into different projects/organisations within the same git-proxy platform.

This fix is split into a separate PR to keep the scope focused and to unblock the security team, who are currently running penetration tests using GitLab.

## Fix

This PR:

* Uses the `getRepoByUrl` method (introduced in #1109) to simplify logic and reliably load the correct repository.
* Updates the `getRepoByUrl` tests accordingly.

## Notes

* A potential future improvement would be to load the repository at the start of the action chain and pass it through to each step, avoiding repeated lookups.

* The `isUserPushAllowed` function is no longer used in the main code path, but cannot yet be removed because its tests in `testDb.test` implicitly set up state that other tests depend on. Specifically, each test in `testDb.test` writes to a file-based database, and subsequent tests rely on that data — making them fragile due to implicit ordering. Removing this function breaks unrelated tests. Refactoring these tests to be isolated and self-contained would be required before safely removing `isUserPushAllowed`, but that is out of scope for this PR.


test/db/mongo/repo.test.js
```diff
@@ -0,0 +1,85 @@
...
+        users: { canPush: [] },
+        url: 'https://github.com/finos/git-proxy',
+      };
+      repoCollection.findOne.resolves(repoData);
+
+      const result = await getRepoByUrl('https://github.com/finos/git-proxy');
+      expect(result).to.equal(repoData);
+      expect(connectionStub.calledWith('repos')).to.be.true;
+      expect(
+        repoCollection.findOne.calledWith({ url: { $eq: 'https://github.com/finos/git-proxy' } }),
```
kriswest: as per comments on #1109 I believe the .git is supposed to be included in the URL as this represents the git endpoint NOT the github webpage (which won't exist for standalone git repos or may exist at a different path to the git endpoint, for example). Github will redirect a browser from the git endpoint to the page anyway.

src/proxy/processors/push-action/checkUserPushPermission.ts
```diff
@@ -1,52 +1,31 @@
...
+const getPushUser = async (action: Action): Promise<string | null> => {
+  const list = await getUsers({ gitAccount: action.user });
+  return list.length === 1 ? list[0].username : null;
+};
 
 // Execute if the repo is approved
 const exec = async (req: any, action: Action): Promise<Action> => {
   const step = new Step('checkUserPushPermission');
 
-  const repoSplit = trimTrailingDotGit(action.repo.toLowerCase()).split('/');
```
jescalada: Is the `trimTrailingDotGit` call deferred in the current execution? It seems this is happening implicitly in the `getRepoByUrl` db functions but it's not exactly the same as before:

```ts
  url = url.toLowerCase().replace('.git', '');
```

Perhaps there's a way to modify or reuse the existing `trimTrailingDotGit`?

test/processors/checkUserPushPermission.test.js
jescalada, line 1: Thanks for getting rid of the log stub checks, the tests are much cleaner now! 👍🏼

jescalada (approved):
Looks good to me, just missing the recent changes from #1109. 🚀

src/proxy/processors/push-action/checkUserPushPermission.ts
andypols: I've removed the `.replace('.git', '')`in `getRepoByUrl`.  I have a new PR that forces the correct form of URL so the problem should go away.

It's all a bit nasty having to use functions like `trimTrailingDotGit` to adjust the urls.  The urls in the proxy actions are always the `.git` kind. We should be able to match on the actual url and remove `trimTrailingDotGit`. We can do this incrementally and keep the PRs small.

jescalada: Thank you for letting me know! This makes sense, and I've created an issue to track it: #1113

jescalada:
Will merge this after #1109 if @kriswest is okay with maintaining his PRs for just a bit longer...

kriswest (changes requested):
Needs to be rebased and simplified now that we retrieve both repos and users differently, hopefully eliminating the original issues. Test changes may still be valuable.

andypols:
Checking this - it no longer adds any value not that #1109 has been merged, co closing it

Issue #694, opened 2024-08-15
Bump OSSF score above 9.0 ⬆️ 


JamieSlome:
@rvema contributed the OSSF Scorecard to the repository in https://github.com/finos/git-proxy/pull/676. If possible, it would be great to drive the score about 9.0 to ensure we excel at meeting OSSF's security standards👍 

```[tasklist]
### Tasks
- [ ] Identify the gaps in policy as calculated by the [scorecard](https://scorecard.dev/viewer/?uri=github.com/finos/git-proxy)
- [ ] Make adjustments to repository settings and codebase to bump score up
```

JamieSlome:
@rvema - pinned dependencies is one of them. Do you want to take this task on?

rvema:
> @rvema - pinned dependencies is one of them. Do you want to take this task on?

For sure, let me take a look and work on it

coderhs:
Should we send one PR per item being addressed. Like I saw here (https://scorecard.dev/viewer/?uri=github.com/finos/git-proxy) there is a vulnerability that can be fixed by upgrading the page `body-parser` to 1.20.3 should that be a single a PR.

JamieSlome:
@coderhs - sure, thank you for volunteering 👍 

Are we able to do in a single PR?

coderhs:
Perhaps we can do per action item as reported.

![image](https://github.com/user-attachments/assets/30eb3d93-c1e7-4304-9018-8899ac63bcfa)

I am not sure if we can go above 9, but with each PR we should be getting closer.

coderhs:
@JamieSlome I have to step back from this, i thought it would be a small thing to do but it seems some of the concerns are with express.js itself that I don't think me without more experience in the project should be upgrading them. Kindly unassing this from me so someone else can pick it up.

I could send a PR for the body-parser that i upgraded if needed.

laukik-target:
@JamieSlome Please assign it to me. I will take this up.

laukik-target:
#762 

@JamieSlome  I am a Citi Employee and Participating for Citi Hackathon.
Please approve  my CLA

laukik-target:
@JamieSlome  Can you please review the above PR?

I would be pushing more PRs to cross-score 9.0. Targeting High Priority items from the OSSF Scorecard first.

Sabarivasan-Velayutham:
@JamieSlome Can you please review the above PR? 

[...]

PR #1115, opened 2025-07-21
test: Implement fuzz tests for processors

### LINKED ISSUE #1111: Implement fuzzing for push actions
Currently, our OSSF score is being lowered because we're not using any fuzzing libraries:

Image

Implementing fuzzing through [`fast-check`](https://fast-check.dev/docs/tutorials/quick-start/basic-setup/) is probably the best choice as it's lightweight, works with JS/TS and easy to integrate with Mocha and other test runners.

For the time being, we can scope this to the push actions (proxy code) only.

The alternative is [OSS-Fuzz](https://google.github.io/oss-fuzz/getting-started/accepting-new-projects/) which requires approval and doesn't work out of the box with JS/TS. At the end of the day, both fuzzers are able to provide the full 10 points for the OSSF scorecard, so I figured we can go with the simpler one.

**Additional context**
None

jescalada:
Fixes #1111, #694 (potentially).

Note that the actual impact on the OSSF score will be known *after* merging (may require another PR to fix if it doesn't increase automatically).


jescalada:
Thanks for the review @06kellyjac! I confirmed that the OSSF score went up from 8.0 to 8.5 - a full 10 points for the fuzzing section. 🎉

Issue #1022, opened 2025-05-21
Improve default admin user creation


jescalada:
**Is your feature request related to a problem? Please describe.**
Currently, a default admin user is created when running git-proxy for the first time:

```ts
/**
 * Create the default admin user if it doesn't exist
 */
export const createDefaultAdmin = async () => {
  const admin = await db.findUser('admin');
  if (!admin) {
    await db.createUser('admin', 'admin', 'admin@place.com', 'none', true);
  }
};
```

**Describe the solution you'd like**
One suggestion by @sam-holmes2 in https://github.com/finos/git-proxy/pull/963, is to have a randomly generated password for the default admin user, and then force to change the password on the first login. Another possible solution is to show a signup form to create the admin user on the first UI load.

Another suggestion by @kriswest is to simply make this configurable in `proxy.config.json`. That would get rid of the security concern altogether, unless the GitProxy administrator didn't change the default admin credentials there.

**Additional context**
I modified some of the related code in #963, and made this issue as I thought it's beyond the scope of the PR.

PR #1120, opened 2025-07-26
Fix display repo add errors

andypols:
# Summary

This PR fixes a bug where the client silently ignored server-side errors when adding a new repository. As a result, users were led to believe the repository was added successfully, only to see it disappear upon refreshing the page.

## Root Cause

The `NewRepo.tsx` component (`src/ui/views/RepoList/Components/NewRepo.tsx`) always closed the modal, regardless of whether the server request succeeded or failed.

## Fix

Display server errors returned by addRepo in `src/ui/services/repo.js`.

Extended Cypress tests to:

- Reproduce the bug.
- Verify the fix.
- Upgraded Cypress to version 14.5.2.
- Added data-testid attributes to improve selector stability in Cypress tests (avoids coupling to CSS/JS changes). These are are also useful in the react unit tests (see note)

## Notes

Adding unit tests for the React components is desirable but considered out of scope for this PR.

### DIFF
--- a/cypress/e2e/repo.cy.js
+++ b/cypress/e2e/repo.cy.js
@@ -1,57 +1,119 @@
 describe('Repo', () => {
   beforeEach(() => {
-    cy.login('admin', 'admin');
-
     cy.visit('/dashboard/repo');
 
     // prevent failures on 404 request and uncaught promises
     cy.on('uncaught:exception', () => false);
   });
 
-  describe('Code button for repo row', () => {
-    it('Opens tooltip with correct content and can copy', () => {
-      const cloneURL = 'http://localhost:8000/finos/git-proxy.git';
-      const tooltipQuery = 'div[role="tooltip"]';
-
-      cy
-        // tooltip isn't open to start with
-        .get(tooltipQuery)
+  describe('Anonymous users', () => {
+    it('Prevents anonymous users from adding repos', () => {
+      cy.get('[data-testid="repo-list-view"]')
+        .find('[data-testid="add-repo-button"]')
         .should('not.exist');
+    });
 
-      cy
-        // find the entry for finos/git-proxy
-        .get('a[href="/dashboard/repo/git-proxy"]')
-        // take it's parent row
-        .closest('tr')
-        // find the nearby span containing Code we can click to open the tooltip
-        .find('span')
-        .contains('Code')
-        .should('exist')
-        .click();
+    describe('Code button for repo row', () => {
+      it('Opens tooltip with correct content and can copy', () => {
+        const cloneURL = 'http://localhost:8000/finos/git-proxy.git';
+        const tooltipQuery = 'div[role="tooltip"]';
 
-      cy
-        // find the newly opened tooltip
-        .get(tooltipQuery)
-        .should('exist')
-        .find('span')
-        // check it contains the url we expect
-        .contains(cloneURL)
-        .should('exist')
-        .parent()
-        // find the adjacent span that contains the svg
-        .find('span')
-        .next()
-        // check it has the copy icon first and click it
-        .get('svg.octicon-copy')
-        .should('exist')
-        .click()
-        // check the icon has changed to the check icon
-        .get('svg.octicon-copy')
-        .should('not.exist')
-        .get('svg.octicon-check')
-        .should('exist');
+        cy
+          // tooltip isn't open to start with
+          .get(tooltipQuery)
+          .should('not.exist');
 
-      // failed to successfully check the clipboard
+        cy
+          // find the entry for finos/git-proxy
+          .get('a[href="/dashboard/repo/git-proxy"]')
+          // take it's parent row
+          .closest('tr')
+          // find the nearby span containing Code we can click to open the tooltip
+          .find('span')
+          .contains('Code')
+          .should('exist')
+          .click();
+
+        cy
+          // find the newly opened tooltip
+          .get(tooltipQuery)
+          .should('exist')
+          .find('span')
+          // check it contains the url we expect
+          .contains(cloneURL)
+          .should('exist')
+          .parent()
+          // find the adjacent span that contains the svg
+          .find('span')
+          .next()
+          // check it has the copy icon first and click it
+          .get('svg.octicon-copy')
+          .should('exist')
+          .click()
+          // check the icon has changed to the check icon
+          .get('svg.octicon-copy')
+          .should('not.exist')
+          .get('svg.octicon-check')
+          .should('exist');
+
+        // failed to successfully check the clipboard
+      });
+    });
+  });
+
+  describe('Regular users', () => {
+    beforeEach(() => {
+      cy.login('user', 'user');
+
+      cy.visit('/dashboard/repo');
+    });
+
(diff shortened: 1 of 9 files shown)


cypress/e2e/repo.cy.js
```diff
@@ -1,57 +1,119 @@
...
     cy.on('uncaught:exception', () => false);
   });
 
-  describe('Code button for repo row', () => {
-    it('Opens tooltip with correct content and can copy', () => {
-      const cloneURL = 'http://localhost:8000/finos/git-proxy.git';
-      const tooltipQuery = 'div[role="tooltip"]';
+  describe('Anonymous users', () => {
+    it('Prevents anonymous users from adding repos', () => {
+      cy.get('[data-testid="repo-list-view"]')
```
jescalada, line 11: It might be a good idea to alias some of these long element identifiers to make tests more readable. Not exactly a problem for smaller tests, but definitely good if the frontend component or tests get refactored:

```js
cy.get('[data-testid="repo-list-view"]').as('repoList');
cy.get('@repoList').find('[data-testid="add-repo-button"]'); // chainable
```

cypress/e2e/repo.cy.js
jescalada, line 1: This test file fails on my local setup due to missing repos (I don't have the default finos/git-proxy repo which is present on a clean install).

Although it might be best for another PR, we could have `cy.seedDb()`, `cy.cleanUp()` functions to run some basic API calls and make sure all E2E tests have the required data before running. Ideally we would seed with data that we know won't be present normally, then delete it after the E2E tests.

Related: #978

src/service/passport/local.js
```diff
@@ -42,13 +42,18 @@ const configure = async (passport) => {
...
-    await db.createUser("admin", "admin", "admin@place.com", "none", true);
-  }
+  const createIfNotExists = async (username, password, email, type, isAdmin) => {
+    const user = await db.findUser(username);
+    if (!user) {
+      await db.createUser(username, password, email, type, isAdmin);
+    }
+  };
+
+  await createIfNotExists('admin', 'admin', 'admin@place.com', 'none', true);
```
jescalada, line 55: Just a heads up - there are plans to improve/refactor the way default users are created: #1022 

I believe the reason why user registering wasn't implemented was that the first adopting organizations only needed ActiveDirectory for their user management. But for others, having default accounts could be a security concern (supposing a malicious actor got access to the GitProxy UI).

Happy to discuss other ways to improve this!

package-lock.json
jescalada, line 1: Seems this file got deleted by accident - unless there's a specific reason to delete it? 🤔

andypols: Good catch - I deleted it to resolve a conflict and added the new one back, but hadn't noticed it's actually in `.gitignore` and did not get added. 

Assume thats a mistake? I can add it back and remove from `.gitignore`?

src/service/passport/local.js
andypols: That's fine — I just needed a user to test with, so I used the existing mechanism. Personally, I prefer having a dedicated testing service that can create and delete test users, repos, etc., so you can use something like cy.ensureUserExists(email, isAdmin, ...) for setup. That way, it's behaviorally separate from the app itself and doesn't depend on anything needing to be run first.

cypress/e2e/repo.cy.js
andypols: Agree that would be good - I like having a test service api that lets the tests setup the system however they want - see comment bellow.

andypols: I double checked and they still run.  Maybe related to this issue #978?

package-lock.json
jescalada: Yes, according to the [npm docs](https://docs.npmjs.com/cli/v11/configuring-npm/package-lock-json), committing the `package-lock.json` is good practice.

Some projects add it to `.gitignore` for various reasons (perhaps preventing unnecessary line changes in PRs).

cypress/e2e/repo.cy.js
jescalada: Added issue for this #1143.

cypress/e2e/repo.cy.js
jescalada: The issue I mentioned previously covers this too: #1143

src/service/passport/local.js
jescalada: Thanks for the suggestion! I've added it to the general E2E test improvement issue.

package-lock.json
jescalada, line 1: I find it odd that there are still some dependencies being added/updated to `package-lock.json` even though the `package.json` hasn't changed. This _might_ be due to running `npm audit --fix` which can update internal dependencies like that.

Specifically, the AWS SDK-related packages are the new "addition" here, and I don't think these are actually being used in the code. The only AWS SDK library that was being imported previously, is `@aws-sdk/credential-providers` which is meant for `mongodb`.

I think the easiest solution would be to discard the changes in `package-lock.json` since we don't intend to introduce new libraries in this PR.

kriswest (changes requested):
Needs a rebase, there are a handful of conflicting/similar changes in main.

I think there are some other cases where the errors aren't displayed that would benefit from similar changes, such as cancelling a push in the dashboard when you are not allowed. That's no doubt a separate issue ;-)

cypress/e2e/repo.cy.js
```diff
@@ -1,82 +1,156 @@
...
+        cy.get('[data-testid="repo-name-input"]').type(repoName);
+        cy.get('[data-testid="repo-url-input"]').type(
+          `https://github.com/cypress-test/${repoName}.git`,
+        );
+        cy.get('[data-testid="add-repo-button"]').click();
       });
+
+      cy.contains('a', `cypress-test/${repoName}`, { timeout: 10000 }).click();
+
+      // cy.get('[data-testid="delete-repo-button"]').click();
```
kriswest, line 58: should this be removed?

cypress/e2e/repo.cy.js
```diff
@@ -1,82 +1,156 @@
...
-      .find('span')
-      .next()
-      .get('svg.octicon-copy')
-      .should('exist')
-      .click()
-      .get('svg.octicon-copy')
-      .should('not.exist')
-      .get('svg.octicon-check')
-      .should('exist');
+    it('Prevents adding an existing repo', () => {
```
kriswest:
```suggestion
    it('Displays an error when adding an existing repo', () => {
```

Slightly different from the other 'Prevents' cases where the button doesn't exist

cypress/e2e/repo.cy.js
andypols: I’d like to test that deletion works from within the app.
However, when running Cypress, tests are automatically re-run on file changes. This results in hitting GitHub API rate limits, which then causes the test to fail. 

I think it would be better to move the GitHub API call to the server so that (different PR :):

- A repo can be added with just its URL, and the server fetches the rest of the details from the API.
- The server can handle rate-limit errors gracefully, instead of letting the request fail, which will also make Cypress tests more reliable.

andypols:
@finos/git-proxy-maintainers Please can someone merge this - thanks

PR #1134, opened 2025-08-03
refactor: replace `getMissingData` action with `checkEmptyBranch`

jescalada:
This is a solution to some problems identified by @kriswest in #973 after merging in the latest security fixes.

I originally wrote `getMissingData` to solve the problem when making commits from an unapproved branch, in which case `action.commitData` and `action.user` are empty and must be fetched later. This means the user checks are deferred, and some checks were inadventently skipped as well.

To sum it up, `getMissingData` was unsafe, due to mistakenly setting the `committer` to the `author_name` which is not always true, and it wasn't actually necessary since Fabio's `checkHiddenCommits` action can handle the edge case mentioned above. I decided to keep only the empty branch check which is still useful and _hopefully_ safe.

### Changelog
- Remove `getMissingData` action
- Add `checkEmptyBranch` action to handle missing `commitData` and empty branch identification
  - I decided to do this separately since `parsePush` is quite bloated already
- Add `parsePush` test for demonstrating that `commitData == []` on empty branch pushes
  - There may be other cases where this is true, such as `--delete` commands, but these are not supported yet


jescalada:
@kriswest Would love to get a quick look on this PR - it should deal with the issue that you mentioned in #973. If this clears things up then we should be able to merge your PR for `2.0.0-rc.1`!

src/proxy/processors/push-action/checkEmptyBranch.ts
```diff
@@ -0,0 +1,45 @@
+import { Action, Step } from '../../actions';
+import simpleGit from 'simple-git';
+import { EMPTY_COMMIT_HASH } from '../constants';
+
+const isEmptyBranch = async (action: Action) => {
+  const git = simpleGit(`${action.proxyGitPath}/${action.repoName}`);
```
kriswest: might as well do this inside the if statement below

src/proxy/processors/push-action/checkEmptyBranch.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  if (action.commitFrom === EMPTY_COMMIT_HASH) {
```
kriswest, line 6: would `commitFrom` and `commitTo` ever be equal?

test/processors/checkCommitMessages.test.js
```diff
@@ -97,8 +97,9 @@ describe('checkCommitMessages', () => {
     });
 
     it('should not error when commit data is empty', async () => {
-      // Empty commit data is a valid scenario that happens when making a branch from an unapproved commit
-      // This is remedied in the getMissingData.exec action  
+      // Empty commit data happens when making a branch from an unapproved commit
+      // or when pushing an empty branch or deleting a branch
+      // This is remedied in the checkEmptyBranch.exec action
```
kriswest:
```suggestion
      // This is handled in the checkEmptyBranch.exec action  
```

src/proxy/processors/push-action/checkEmptyBranch.ts
```diff
@@ -0,0 +1,45 @@
...
+    try {
+      const type = await git.raw(['cat-file', '-t', action.commitTo || '']);
+      const known = type.trim() === 'commit';
+      if (known) {
+        return true;
+      }
```
coopernetes: Can be simplified.
```suggestion
      return type.trim() === 'commit';
```

src/proxy/processors/push-action/checkEmptyBranch.ts
coopernetes: IME I don't _believe_ so. When the git client initially reaches out to a remote server using the "smart" protocol, it grabs the list of existing references via `GET https://{proxied-upstream-repository-url}/info/refs?service=git-receive-pack`. If the local repository's branch contains the same ref, the client doesn't send the subsequent request (`POST /git-receive-pack`). 

https://git-scm.com/docs/http-protocol#_discovering_references

PR #1139, opened 2025-08-04
feat: Add Annotated Tag Push Support

### LINKED ISSUE #56: Add support for tags push

### LINKED ISSUE #986: Add support for pushing tags `git push --tags`

fabiovincenzi:
Description:
This PR introduces end-to-end support for annotated tag pushes 

- Split the push pipeline into separate `branchPushChain` and `tagPushChain`.
- Update `getChain()` to select the correct chain based on `action.type` and presence of `action.tag`.
- Render a dedicated “Tag Details” table when viewing a tag push instead of the commit diff.

Fixes #986


src/proxy/processors/push-action/parsePush.ts
```diff
@@ -93,30 +103,79 @@ async function exec(req: any, action: Action): Promise<Action> {
...
       if (action.commitFrom === EMPTY_COMMIT_HASH) {
         action.commitFrom = action.commitData[action.commitData.length - 1].parent;
       }
-      const user = action.commitData[action.commitData.length - 1].committer;
-      action.user = user;
+      action.user = action.commitData.at(-1)!.committer;
+    } else if (action.tagData?.length) {
+      action.user = action.tagData.at(-1)!.tagger;
+    } else {
+      throw new Error('No commit or tag data parsed from packfile');
```
jescalada: It seems that the `commitData` check here is causing this test to fail:

```js
it.only('should return empty commitData on empty branch push', async () => {
  const emptyPackBuffer = createEmptyPackBuffer();

  const newCommit = 'b'.repeat(40);
  const ref = 'refs/heads/feature/emptybranch';
  const packetLine = `${EMPTY_COMMIT_HASH} ${newCommit} ${ref}\0capabilities\n`;

  req.body = Buffer.concat([createPacketLineBuffer([packetLine]), emptyPackBuffer]);

  const result = await exec(req, action);

  expect(result).to.equal(action);

  const step = action.steps.find(s => s.stepName === 'parsePackFile');
  expect(step).to.exist;
  expect(step.error).to.be.false; // <-- Fails here
  expect(action.branch).to.equal(ref);
  expect(action.setCommit.calledOnceWith(EMPTY_COMMIT_HASH, newCommit)).to.be.true;

  expect(action.commitData).to.be.an('array').with.lengthOf(0);
});
```

Ideally, we want to allow empty `commitData` so that we can interpret it later: It might be an empty branch push, a tag push or something else entirely.

Perhaps we could move this check into the another action? We could rename the `checkEmptyBranch` action to `handleEmptyCommitData` and do all the relevant checks there.

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -93,30 +103,79 @@ async function exec(req: any, action: Action): Promise<Action> {
...
-      `Unable to parse push. Please contact an administrator for support: ${e.toString('utf-8')}`,
+      `Unable to parse push. Please contact an administrator for support: ${e.message || e.toString()}`,
     );
   } finally {
     action.addStep(step);
   }
   return action;
 }
 
+function parseTag(x: CommitContent): TagData {
```
jescalada: I noticed that this function is not extracting the timestamp, which is actually present in the `tagger` field within the `content`:

Image

We should extract it and pass it to the frontend, which currently displays N/A:

Image

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -41,8 +35,10 @@ async function exec(req: any, action: Action): Promise<Action> {
       action.addStep(step);
       return action;
     }
+    console.log('RRR');
     const [packetLines, packDataOffset] = parsePacketLines(req.body);
-    const refUpdates = packetLines.filter((line) => line.includes(BRANCH_PREFIX));
+    console.log('PPPP');
+    const refUpdates = packetLines.filter((line) => line.includes('refs/'));
 
     if (refUpdates.length !== 1) {
```
jescalada: The flow described in the issue seems to be failing here, due to "multiple ref updates":

```bash
git tag --annotate foo -m ''
git cat-file tag foo
git push --tags
```

Image

I think this is happening due to tags (or other objects) that haven't been approved yet, and git is trying to push everything at once when calling `git push --tags`.

This will probably be complicated to solve - for the time being we could improve the error message to say something like:

```ts
step.setError(
  'Your push has been blocked. Multi-ref pushes (multiple tags and/or branches) are not supported yet. Please push one ref at a time.',
);
```

jescalada: Note that when I approved my previous tag push, it did reduce the number of refs, so it's just bundling and pushing all the unapproved branches + tags at once.

Image

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -41,8 +35,10 @@ async function exec(req: any, action: Action): Promise<Action> {
       action.addStep(step);
       return action;
     }
+    console.log('RRR');
```
jescalada: Cleanup 👀

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -69,7 +65,21 @@ async function exec(req: any, action: Action): Promise<Action> {
 
     // Strip everything after NUL, which is cap-list from
     // https://git-scm.com/docs/http-protocol#_smart_server_response
-    action.branch = ref.replace(/\0.*/, '').trim();
+    const refName = ref.replace(/\0.*/, '').trim();
+    const isTag = refName.startsWith('refs/tags/');
+    const isBranch = refName.startsWith('refs/heads/');
```
jescalada: Perhaps this could use the `BRANCH_PREFIX` from before? And have a new `TAG_PREFIX` constant for `'refs/tags/`?

jescalada (changes requested):
The core flow seems to work correctly! Just a few edge cases and an integration issue with the `checkEmptyBranch` that might be worth refactoring.

Also, we will be merging #973 soon which will likely cause a few conflicts here... Might want an extra pair of eyes to make sure everything is good to merge.

src/proxy/chain.ts
```diff
@@ -32,16 +43,19 @@ let pluginsInserted = false;
 export const executeChain = async (req: any, res: any): Promise<Action> => {
   let action: Action = {} as Action;
   try {
+    // 1) Initialize basic action fields
     action = await proc.pre.parseAction(req);
+    // 2) Parse the push payload first to detect tags/branches
+    if (action.type === RequestType.PUSH) {
+      action = await proc.push.parsePush(req, action);
```
jescalada: I'm not sure if I misunderstood the flow here, but isn't `parsePush` executing twice when pushing branches? Seems it executes here, and once again as part of the `branchPushChain`.

src/proxy/processors/push-action/parsePush.ts
jescalada: Works properly now! 👍🏼

src/proxy/chain.ts
jescalada: Regardless, it seems the push flow is working as expected - just not sure if this is running twice or if the `parsePush` prevents the double execution somehow.

jescalada (approved):
Looks good to me! I tested the tags and regular push flow and both work as expected with the new updates 👍🏼 

Just a comment on the `parsePush` which I'm not sure if it's executing once or twice on pushes.

jescalada:
As we're modifying the proxy logic, another pair of eyes here would be super helpful! 👀  @finos/git-proxy-maintainers

andypols (approved):
LGTM

src/proxy/processors/pre-processor/parseAction.ts
```diff
@@ -16,21 +16,21 @@
...
-import { Action } from '../../actions';
+import { Action, RequestType } from '../../actions';
 import { processUrlPath } from '../../routes/helper';
 import * as db from '../../../db';
 
 const exec = async (req: Request) => {
   const id = Date.now();
   const timestamp = id;
-  let type = 'default';
+  let type: RequestType | string = 'default';
```
kriswest: RequestType should have a value for default - which applies to every proxied operation that is NOT a push or pull. That would eliminate teh union here as the override (as) on line 55

src/proxy/actions/Action.ts
```diff
@@ -17,13 +17,29 @@
 import { processGitURLForNameAndOrg, processUrlPath } from '../routes/helper';
 import { Step } from './Step';
 import { CompletedAttestation, CommitData, Rejection } from '../processors/types';
+import { TagData } from '../../types/models';
+
+export enum RequestType {
+  PUSH = 'push',
+
+  PULL = 'pull',
+}
```
kriswest, line 28: Theres a third type that I think you need here which covers everything other than push or pull that we proxy. Default isn't the best name for it (perhaps 'other?') but it is the current value.

```suggestion
export enum RequestType {
  PUSH = 'push',
  PULL = 'pull',
  DEFAULT = 'default',
}
```

src/proxy/processors/pre-processor/parseAction.ts
```diff
@@ -52,7 +52,7 @@ const exec = async (req: Request) => {
     );
   }
 
-  return new Action(id.toString(), type, req.method, timestamp, url);
+  return new Action(id.toString(), type as RequestType, req.method, timestamp, url);
```
kriswest: Lose the type override by correcting RequestType to have a 'default' value, which can actually be returned here I think.

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -62,13 +65,13 @@ async function exec(req: Request, action: Action): Promise<Action> {
     }
 
     const [packetLines, packDataOffset] = parsePacketLines(req.body);
-    const refUpdates = packetLines.filter((line) => line.includes(BRANCH_PREFIX));
+    const refUpdates = packetLines.filter((line) => line.includes('refs/'));
```
kriswest: Create a new constant for this? BRANCH_AND_TAG_PREFIX ?

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -62,13 +65,13 @@ async function exec(req: Request, action: Action): Promise<Action> {
...
 
     if (refUpdates.length !== 1) {
-      step.log('Invalid number of branch updates.');
+      step.log('Invalid number of ref updates.');
       step.log(`Expected 1, but got ${refUpdates.length}`);
```
kriswest: This can only handle a single tag push right now, but I feel like it should be able to handle a push with multiple tags... Supporting that would mean a bit more complexity here (checking that all the refs are tags) and some other code changes but I think it might be worth doing...

I suspect users will be more used to pushing all tags:
`git push <remote> --tags`
than pushing a single tag:
`git push <remote> <tagname>`

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -84,7 +87,21 @@ async function exec(req: Request, action: Action): Promise<Action> {
 
     // Strip everything after NUL, which is cap-list from
     // https://git-scm.com/docs/http-protocol#_smart_server_response
-    action.branch = ref.replace(/\0.*/, '').trim();
+    const refName = ref.replace(/\0.*/, '').trim();
+    const isTag = refName.startsWith(TAG_PREFIX);
+    const isBranch = refName.startsWith(BRANCH_PREFIX);
+
+    action.branch = isBranch ? refName : undefined;
+    action.tag = isTag ? refName : undefined;
```
kriswest: This only supports a single tag name, but tagData is an array... I think you should change this to support multiple tag names.

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -84,7 +87,21 @@ async function exec(req: Request, action: Action): Promise<Action> {
...
+
+    // Set actionType based on what type of push this is
+    if (isTag) {
+      action.actionType = ActionType.TAG;
+    } else if (isBranch) {
+      action.actionType = ActionType.BRANCH;
+    } else {
+      action.actionType = ActionType.COMMIT;
```
kriswest: I'm not sure I understand this type -  is this used when pushing commits to an existing branch?

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -105,21 +122,34 @@ async function exec(req: Request, action: Action): Promise<Action> {
...
-    } else {
+    for (const obj of contents) {
+      if (obj.type === GIT_OBJECT_TYPE_COMMIT) ParsedObjects.commits.push(...getCommitData([obj]));
+      else if (obj.type === GIT_OBJECT_TYPE_TAG) ParsedObjects.tags.push(parseTag(obj));
+    }
+
+    action.commitData = ParsedObjects.commits;
+    action.tagData = ParsedObjects.tags;
+
+    if (action.commitData.length) {
```
kriswest: Should probably be based on `action.actionType` instead?

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -105,21 +122,34 @@ async function exec(req: Request, action: Action): Promise<Action> {
...
         action.commitFrom = action.commitData[action.commitData.length - 1].parent;
       }
-
       const { committer, committerEmail } = action.commitData[action.commitData.length - 1];
       // Note: This is not always the pusher's email, it's the last committer's email.
       // See https://github.com/finos/git-proxy/issues/1400
       step.log(`Push request received from user ${committer} with email ${committerEmail}`);
       action.user = committer;
       action.userEmail = committerEmail;
+    } else if (action.tagData?.length) {
```
kriswest: Should probably be based on action.actionType instead

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -105,21 +122,34 @@ async function exec(req: Request, action: Action): Promise<Action> {
...
+      action.user = action.tagData.at(-1)!.tagger;
+      action.userEmail = action.tagData.at(-1)!.taggerEmail;
+    } else {
+      step.log('No commit data found when parsing push.');
```
kriswest:
```suggestion
      step.log('No commit or tag data found when parsing push.');
```

May mean a test needs updating.

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -134,6 +164,44 @@ async function exec(req: Request, action: Action): Promise<Action> {
   return action;
 }
 
+function parseTag(x: CommitContent): TagData {
```
kriswest: minor: could perhaps be called `getTagData` for consistency with `getCommitData` and the type name

src/proxy/chain.ts
```diff
@@ -40,6 +39,17 @@ const pushActionChain: ((req: Request, action: Action) => Promise<Action>)[] = [
...
 ];
 
+const tagPushChain: ((req: Request, action: Action) => Promise<Action>)[] = [
+  proc.push.checkRepoInAuthorisedList,
+  proc.push.checkUserPushPermission,
+  proc.push.checkIfWaitingAuth,
+  proc.push.pullRemote,
+  proc.push.writePack,
+  proc.push.preReceive,
+  // TODO: implement tag message validation?
```
kriswest: Yes should be implemented. You could probably use the existing `commitConfig` to do so.

I was going to suggest also implementing checkAuthorEmails - but I don't think we need to. A tag push is not going to expose IP so I think it less necessary to check who created the tag.

src/proxy/chain.ts
kriswest: A change above removed `proc.push.parsePush` from the `branchPushChain`, but I don't know why./ I think it should just be the first processor in both the `branchPushChain` and `tagPushChain`

src/proxy/chain.ts
```diff
@@ -96,6 +113,22 @@ export const executeChain = async (req: Request, _res: Response): Promise<Action
...
+  switch (action.actionType) {
+    case ActionType.TAG:
+      return tagPushChain;
+    case ActionType.BRANCH:
+    case ActionType.COMMIT:
+    default:
+      return branchPushChain;
+  }
+};
+
```
kriswest: This could just be a 2 line if statement or 1-line ternary down in the switch statement on line 165

src/ui/utils/pushUtils.ts
```diff
@@ -0,0 +1,260 @@
...
+ * Gets the display timestamp for a push (handles both commits and tags)
+ * @param {boolean} isTag - Whether this is a tag push
+ * @param {CommitData | null} commitData - The commit data
+ * @param {TagData} [tagData] - The tag data (optional)
+ * @return {string} Formatted timestamp string or 'N/A'
+ */
+export const getDisplayTimestamp = (
+  isTag: boolean,
+  commitData: CommitData | null,
+  tagData?: TagData,
```
kriswest: Could you use a type guard to reduce this to a single argument thats typed `CommitData | TagData`?

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -105,21 +122,34 @@ async function exec(req: Request, action: Action): Promise<Action> {
...
-    if (action.commitData.length === 0) {
-      step.log('No commit data found when parsing push.');
-    } else {
+    for (const obj of contents) {
+      if (obj.type === GIT_OBJECT_TYPE_COMMIT) ParsedObjects.commits.push(...getCommitData([obj]));
+      else if (obj.type === GIT_OBJECT_TYPE_TAG) ParsedObjects.tags.push(parseTag(obj));
+    }
+
+    action.commitData = ParsedObjects.commits;
+    action.tagData = ParsedObjects.tags;
```
kriswest: This could be simpler, e.g.:
```suggestion
    for (const obj of contents) {
      if (obj.type === GIT_OBJECT_TYPE_COMMIT) {
        action.commitData = getCommitData(contents);
      } else if (obj.type === GIT_OBJECT_TYPE_TAG) {
        action.tagData = [ parseTag(obj) ]
      }  
    }
```

src/proxy/actions/Action.ts
```diff
@@ -17,13 +17,29 @@
...
+
+export enum ActionType {
+  COMMIT = 'commit',
+
+  TAG = 'tag',
+
+  BRANCH = 'branch',
+}
```
kriswest: Please add comments (to explain the difference between COMMIT and BRANCH)

kriswest (commented):
Overall not bad. There are a few small code quality things that could be dealt with (see comments).

Could this PR be improved to handle multiple tags in a single push? Feels like that would be best done when the feature is first introduced to save later rework and potentially breaking changes.

I don't fully understand the difference between ActionType.COMMIT and ActionType.BRANCH, which could do with a comment explaining (both on the tyupe and where each is used).

kriswest:
Resolved comments that have been dealt with already

src/proxy/processors/pre-processor/parsePush.ts
```diff
@@ -61,32 +64,46 @@ async function exec(req: Request, action: Action): Promise<Action> {
...
+      };
+    });
+
+    const allTags = parsedRefs.every((r) => r.isTag);
+
+    if (parsedRefs.length > 1 && !allTags) {
+      step.log(`Received ${parsedRefs.length} ref updates with mixed or multiple branch refs.`);
       throw new Error(
-        'Your push has been blocked. Please make sure you are pushing to a single branch.',
+        'Your push has been blocked. Multi-ref pushes are only supported for tags. Please push one branch at a time.',
```
jescalada, line 95: 👍🏼

src/ui/utils/pushUtils.ts
```diff
@@ -0,0 +1,262 @@
...
+  if (!tagRefs || !Array.isArray(tagRefs) || tagRefs.length === 0) return '';
+  return tagRefs[0].replace('refs/tags/', '');
+};
+
+/**
+ * Gets the appropriate reference to show (tag name or branch name)
+ * @param {PushData} pushData - The push data
+ * @return {string} The reference name to display
+ */
+export const getRefToShow = (pushData: PushData): string => {
```
jescalada, line 78: I'm getting an error here when pushing an empty tag via: `git tag test-tag-1 && git push test-tag-1`:

image

This also creates errors on the dashboard:

image

I tried patching this by modifying the call to `trimPrefixRefsHeads` to default to empty strings like such:

```ts
  return trimPrefixRefsHeads(pushData.branch || '');
``` 

And now the empty tag push is visible on the dashboard, but only on the "All" tab, whereas I expected it to show on the "Pending" tab.

image

---

Notably, the reason for failure is `checkUserPushPermission` failing to find the email associated with the push:

image

jescalada: If non-annotated tag support isn't included, we should perhaps check and throw a more descriptive error as the current one is misleading 👍🏼

src/proxy/chain.ts
```diff
@@ -17,16 +17,15 @@
...
 import { PluginLoader } from '../plugin';
-import { Action } from './actions';
+import { Action, RequestType, ActionType } from './actions';
 import * as proc from './processors';
 import { attemptAutoApproval, attemptAutoRejection } from './actions/autoActions';
 import { handleErrorAndLog } from '../utils/errors';
 
-const pushActionChain: ((req: Request, action: Action) => Promise<Action>)[] = [
-  proc.push.parsePush,
+const branchPushChain: ((req: Request, action: Action) => Promise<Action>)[] = [
```
jescalada: Wondering if we could add a type for this `(req: Request, action: Action) => Promise<Action>` which gets reused in many places

jescalada: We could perhaps have an alias for it based on the `Processor` in `src/proxy/processors/types`:

```ts
export interface Processor {
  exec(req: Request, action: Action): Promise<Action>;
  metadata: ProcessorMetadata;
}
```

Could be used as:

```ts
type ChainStep = Processor['exec'];
```

src/proxy/chain.ts
jescalada: @kriswest I've spent some time trying to understand/refactor this, and I think I see the reasoning: It makes sense for `parsePush` to be separate to the other actions, since it's the only place where we can determine what kind of Git object (tag, branch, etc.) is getting pushed. And it also makes sense to have different chains for different Git objects, since they are often processed differently (we don't check for commit content for tag pushes, etc.).

Since `parsePush` is the "odd one out" (just another pre-processor for pushes), we should probably specify this special case in our docs, and refactor our code accordingly (moving it into `/src/proxy/processors/pre-processor`).

I think the confusion comes from calling `parsePush` an Action/processor in the chain when it has become a pre-processor exclusively for pushes.

---

Wondering if there's a better solution from an architecture point of view. 🤔

jescalada: Pinging @fabiovincenzi and @finos/git-proxy-maintainers for ideas

jescalada:
LGTM so far, the general flow is still working as expected, and the double `parsePush` issue is fixed 👍🏼 

Just a few issues to iron out:
- Empty tag pushes give a missing user error which is misleading
- We should figure out whether to refactor the chains to contain `parsePush` again (which makes intuitive sense)
  - Otherwise, we can make `parsePush` a pre-processor, considering it pre-processes pushes to define what type of Git object is being pushed, and consequentially which push chain must be executed (`branchPushChain`, `tagPushChain` or something else).

src/proxy/chain.ts
kriswest: I'm a fan of using index types like this (I don't think you actually need the alias but can use one if you want) as they are readable and save of keeping extra types in sync.

src/proxy/actions/Action.ts
```diff
@@ -17,13 +17,31 @@
...
+
+export enum RequestType {
+  PUSH = 'push',
+
+  PULL = 'pull',
+
+  DEFAULT = 'default',
+}
+
+export enum ActionType {
```
kriswest: This could perhaps be `PushType` to make it more intuitive.

src/proxy/chain.ts
kriswest: Thansk @jescalada, I see the issue and yes we need to start parsing the push to establish type, makes sense. I agree with moving it over the pre-processors folder and keeping it out of the individual chains..

src/ui/utils/pushUtils.ts
kriswest: Apparently the term is 'lightweight' tag. See https://git-scm.com/book/en/v2/Git-Basics-Tagging#:~:text=Git%20supports%20two%20types%20of,objects%20in%20the%20Git%20database for some reading on them.

It would probably be good to support these or throw a specific error message on them, but we'll need to classify them. We have an issue with them as the awful hack used to cover the fact we don't actually know the pusher (use the last committer email) fails here as theres no commit data. Hence, perhaps we should NOT support these for now and add an error message and a TODO until we fix that architectural flaw.

src/proxy/chain.ts
kriswest: THanks for looking at this @jescalada - yes I missed that it was doing the parsing that detected the push type. I agree with moving this to the pre-processors folder and keeping it out of the individual chains. The comments added help a lot.

Will we ever have to handle mixed pushes of both branches and tags... Doh, yes: https://stackoverflow.com/questions/3745135/push-git-commits-tags-simultaneously

We'll need to handle that case - we could detect and reject it for now?

fabiovincenzi:
All review comments have been addressed. Ready for a final look! @kriswest @jescalada

src/proxy/processors/pre-processor/parsePush.ts
```diff
@@ -95,25 +97,22 @@ async function exec(req: Request, action: Action): Promise<Action> {
...
-      action.actionType = ActionType.TAG;
+      action.actionType = PushType.TAG;
       action.tags = parsedRefs.map((r) => r.refName);
     } else {
-      action.actionType = ActionType.BRANCH;
+      action.actionType = PushType.BRANCH;
       action.branch = parsedRefs[0].refName;
     }
 
-    // Use the first ref's commit range for the action id
```
kriswest: These comments should stay...

jescalada: @fabiovincenzi In case you missed this 👍🏼

src/proxy/chain.ts
jescalada: Resolving this as it's been handled in the latest commits 👍🏼

jescalada (approved):
LGTM! 🚀 Just checked that existing comments have been resolved.

PR #1140, opened 2025-08-04
refactor: migrate configuration system from JSON Schema to QuickType

fabiovincenzi:
Replace JSON-Schema for config validation and default injection
Add typed `config` export, `loadConfig()`, `validate()` and `setConfigFile()`
Update` index.ts` to use the new loader and pass the validated config.

---

## Note: Restored from deleted fork

This PR recreates the **original PR #998**, which was automatically closed due to accidental fork deletion.

** For discussions and reviews:** See the original PR https://github.com/finos/git-proxy/pull/998

*All commits are identical to the original with preserved git history.*


package.json
```diff
@@ -23,7 +23,8 @@
     "lint:fix": "eslint --fix \"src/**/*.{js,jsx,ts,tsx,json}\" \"test/**/*.{js,jsx,ts,tsx,json}\"",
     "format": "prettier --write src/**/*.{js,jsx,ts,tsx,css,md,json,scss} test/**/*.{js,jsx,ts,tsx,json} packages/git-proxy-cli/test/**/*.{js,jsx,ts,tsx,json}  packages/git-proxy-cli/index.js --config ./.prettierrc",
     "gen-schema-doc": "node ./scripts/doc-schema.js",
-    "cypress:run": "cypress run"
+    "cypress:run": "cypress run",
+    "generate-types": "quicktype --src-lang schema --lang typescript --out src/config/generated/config.ts --top-level GitProxyConfig config.schema.json && ts-node scripts/add-banner.ts src/config/generated/config.ts"
```
jescalada: I wonder if this can/should be executed either in the pre-commit hook or the CI? We might have problems if the config was changed but the developer forgot to generate the types for it.

Would be nice to make the script name more clear: since it's generating types for the config, it could be `generate-config-types`.

jescalada (approved):
LGTM!

I wonder if there's a way to fix the CLA issue? We could otherwise re-open the PR with cherry-picked changes.

package.json
fabiovincenzi: Good point! I've changed the name, 
I considered adding it to the pre-commit hook, but that would slow down every commit and automatically modify files during commit (which can be confusing).
Types are regenerated when building (for releases) and CI catches any inconsistencies
Happy to adjust this approach if anyone has a better way to handle it!

jescalada:
@kriswest Would be great if you could take another look at this - I think it fixes a lot of typing issues with the config.

config.schema.json
```diff
@@ -149,6 +149,14 @@
       },
       "required": ["enabled", "key", "cert"]
     },
+    "sslKeyPemPath": {
+      "description": "Legacy: Path to SSL private key file (use tls.key instead)",
```
kriswest: Could use the deprecated annotation here: https://json-schema.org/understanding-json-schema/reference/annotations, which I think generates the jsdocs annotation for the same.

I would use the word Deprecated in the description rather than legacy (but thats a personal preference I guess)

scripts/add-banner.ts
```diff
@@ -0,0 +1,24 @@
+import * as fs from 'fs';
+import * as path from 'path';
+
+const banner = '// THIS FILE IS AUTOMATICALLY GENERATED – DO NOT EDIT MANUALLY.\n\n';
```
kriswest, line 4: ❤️

test/generated-config.test.js
```diff
@@ -0,0 +1,361 @@
+const chai = require('chai');
+const { Convert } = require('../src/config/generated/config');
+
+const { expect } = chai;
+
+describe('Generated Config (QuickType)', () => {
+  describe('Convert class', () => {
+    it('should parse valid configuration JSON', () => {
```
kriswest, line 9: Is there a test that applies the validator to the default config - there should be as that _may_ catch additions to the schema that haven't been generated...

src/config/generated/config.ts
```diff
@@ -0,0 +1,636 @@
...
+/**
+ * Configuration for customizing git-proxy
+ */
+export interface GitProxyConfig {
+  /**
+   * Third party APIs
+   */
+  api?: API;
+  /**
+   * List of authentication sources for API endpoints. May be empty, in which case all
```
kriswest, line 21: Looking forward to having jsDocs for all the config elements! Hopefully, this will drive us to fill in missing descriptions in the schema (that being where they come from).

kriswest (commented):
I think this will be a great improvement, thanks for taking it on. I have one question about validation of the default config to answer, otherwise I think good to go

test/generated-config.test.js
fabiovincenzi: You are right that's a good case!

PR #929, opened 2025-02-28
refactor: Add TypeScript support

fabiovincenzi:
**Overview**
This PR introduces TypeScript to git-proxy and refactors relevant code to support it.

**Changelog**
- Added TypeScript configuration (tsconfig.json) with essential settings for strict type checking, ES6 compatibility, and JSX support.
- Added typescript and ts-node to manage TypeScript code compilation and execution.
- Updated the package.json to include TypeScript dependencies.
- Converted the main entry file from JavaScript (index.js) to TypeScript (index.ts).
- Modified the start script to use ts-node for running TypeScript files.

Thanks to @jescalada for the work on fixing CI issues on https://github.com/G-Research/git-proxy/pull/31, which helped guide the necessary changes in this PR.

Related issue: #927

Issue #927, opened 2025-02-28
Add TypeScript to the Project


fabiovincenzi:
### Goal
Set up TypeScript with an initial loose configuration while ensuring JavaScript files remain executable.

### Tasks

- Install TypeScript dependencies (typescript, @types/node).
- Create a tsconfig.json file with loose settings.
- Ensure the project builds successfully.

### Notes

- Start with a minimal tsconfig.json, then adjust settings incrementally.
- Ensure JavaScript files can still be executed without requiring immediate conversion to TypeScript.

jescalada:
Closing this - should be all done after #1174 is completed!

PR #1065, opened 2025-06-29
refactor: refactor `cli` to TS + ESM

jescalada:
Fixes #1064.

In progress: Test file conversion

PR #960, opened 2025-03-28
refactor(ts): refactor `proxy` to TS

jescalada:
Fixes #965.

This PR refactors most of the proxy logic to TS. It builds on the `db` and `config` refactor PRs.

The tests are passing, but they do not cover the push/pull processor functions (because the logic is stubbed). There's a fix for this in #805, which might be worth looking at!

PR #1142, opened 2025-08-04
refactor: Add TypeScript support

fabiovincenzi:
**Overview**
This PR introduces TypeScript to git-proxy and refactors relevant code to support it.

**Changelog**
- Added TypeScript configuration (tsconfig.json) with essential settings for strict type checking, ES6 compatibility, and JSX support.
- Added typescript and ts-node to manage TypeScript code compilation and execution.
- Updated the package.json to include TypeScript dependencies.
- Converted the main entry file from JavaScript (index.js) to TypeScript (index.ts).
- Modified the start script to use ts-node for running TypeScript files.

Thanks to @jescalada for the work on fixing CI issues on https://github.com/G-Research/git-proxy/pull/31, which helped guide the necessary changes in this PR.

Related issue: #927 

---

## Note: Restored from deleted fork

This PR recreates the **original PR #929**, which was automatically closed due to accidental fork deletion.

** For discussions and reviews:** See the original PR https://github.com/finos/git-proxy/pull/929

*All commits are identical to the original with preserved git history.*

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -60,7 +60,7 @@ jobs:
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@3599b3baa15b485a2e49ef411a7a4bb2452e7f93 # v3
+        uses: github/codeql-action/init@42213152a85ae7569bdb6bec7bcd74cd691bfe41 # v3
         with:
           languages: ${{ matrix.language }}
           # If you wish to specify custom queries, you can do so here or in a config file.
@@ -73,7 +73,7 @@ jobs:
       # Autobuild attempts to build any compiled languages (C/C++, C#, Go, Java, or Swift).
       # If this step fails, then you should remove it and run the build manually (see below)
       - name: Autobuild
-        uses: github/codeql-action/autobuild@3599b3baa15b485a2e49ef411a7a4bb2452e7f93 # v3
+        uses: github/codeql-action/autobuild@42213152a85ae7569bdb6bec7bcd74cd691bfe41 # v3
 
       # ℹ️ Command-line programs to run using the OS shell.
       # 📚 See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun
@@ -86,6 +86,6 @@ jobs:
       #     ./location_of_script_within_repo/buildscript.sh
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@3599b3baa15b485a2e49ef411a7a4bb2452e7f93 # v3
+        uses: github/codeql-action/analyze@42213152a85ae7569bdb6bec7bcd74cd691bfe41 # v3
         with:
           category: '/language:${{matrix.language}}'
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -17,7 +17,7 @@ jobs:
       - name: 'Checkout Repository'
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
       - name: Dependency Review
-        uses: actions/dependency-review-action@56339e523c0409420f6c2c9a2f4292bbb3c07dd3 # v4
+        uses: actions/dependency-review-action@45529485b5eb76184ced07362d2331fd9d26f03f # v4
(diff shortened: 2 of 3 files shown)


jescalada (approved):
Reapproving this - it's a lot easier to see the changes now that merge conflicts have been resolved 👍🏼 

@coopernetes @06kellyjac I suppose this could go into the first release candidate for v2? Or should we deal with the few remaining TS PRs and get it all in a single release? #960 #1063 #1065

.npmignore
```diff
@@ -1,5 +1,12 @@
...
+tsconfig.json
+jest.config.js
+.eslintrc.js
+.prettierrc
+
 website/
-plugins/
-experimental/
-cypress/
+plugins/
```
06kellyjac:
```suggestion
plugins/
```

Missing EOL

package.json
```diff
@@ -2,6 +2,8 @@
   "name": "@finos/git-proxy",
   "version": "1.19.2",
   "description": "Deploy custom push protections and policies on top of Git.",
+  "main": "dist/index.js",
```
06kellyjac, line 5: These files don't live in `dist` due to `./scripts/build-for-publish.sh`

kriswest: @fabiovincenzi I think @06kellyjac is right. Have you tried running `npm run build-lib`? I would assume you can then `npm pack` and you should have module you can test (can you run it/can you import types from it). I _suspect_ both would fail with the _dist/_ in the path.

kriswest (commented):
There is a need to test a against the release publishing scripts before merging this as the @06kellyjac and I are both concerned that the changes to package.json will break import of types and running via the `main` entry.

I am unsure of the correct solution or why we modify the package when publishing...

package.json
fabiovincenzi: This is outdated, package.json now doesn't have dist/ in the path, I just tested it ,and seems to work on my end, maybe a double check would be great.

kriswest (commented):
Still LGTM, will attampt to merge when checks are done

kriswest:
@06kellyjac could you approve this one then merge? Its still got you down as a requesting changes and is asking for a second reviewer:
image

package.json
```diff
@@ -2,6 +2,40 @@
   "name": "@finos/git-proxy",
   "version": "2.0.0-rc.3",
   "description": "Deploy custom push protections and policies on top of Git.",
+  "main": "index.js",
+  "types": "index.d.ts",
+  "exports": {
+    ".": {
```
06kellyjac: If we're emitting to `./dist` does this need adjusting?

06kellyjac: I guess we still have `./prepare.js` IIRC we're going to drop that

package.json
```diff
@@ -2,6 +2,40 @@
...
+    "./db": {
+      "import": "./src/db/index.js",
+      "require": "./src/db/index.js",
+      "types": "./src/db/index.d.ts"
+    },
+    "./proxy": {
+      "import": "./src/proxy/index.js",
+      "require": "./src/proxy/index.js",
+      "types": "./src/proxy/index.d.ts"
+    },
```
06kellyjac, line 27: Similar applies to anything directly referencing `./src`

06kellyjac (changes requested):
There's still some complications from what I can see

Also with 38 commits can we get a squash please, many thanks.

package.json
fabiovincenzi: Mmh `build-for-publish.sh` script moves compiled files to root and `./src/`, then deletes `dist/`, what do you mean for `./prepare.js`?

.github/workflows/sample-publish.yml
```diff
@@ -9,22 +9,32 @@ permissions:
...
+
+      - name: Build TypeScript
+        working-directory: plugins/git-proxy-plugin-samples
+        run: npm run build
+
+      - name: Install peers and publish
         working-directory: plugins/git-proxy-plugin-samples
+        run: |
+          npm install --include=peer
+          npm publish --access=public
```
kriswest:
```suggestion
          npm publish --provenance --access=public
```

06kellyjac: Thanks :slightly_smiling_face: 
It think if we resolve the changes & land this suggestion we can merge.

I have a few potential concerns with the configuration as it stands but it'd be best to merge this and iterate.
Once we have this in main I'll do some local testing, maybe we can have another RC release.

Issue #1019, opened 2025-05-20
Define and publicise our standards for testing and CI gating


JamieSlome:
As part of our community meeting yesterday (19th May 2025), we agreed that a small overhaul of our testing approach is required. This thread came about as a result of PRs going stale that had already met test thresholds but had not been merged. We want to encourage a highly automated environment (where possible) to reduce burden on maintainers. Testing can provide assurance around new code, without a maintainer having to powerlift too much!

### Wishlist ❇ 

- [ ] #1150
- [x] Update docs site to include guidance and specification on how to test and level of coverage required
- [x] Discuss and conclude on minimum code coverage required for PRs and patches
- [x] Discuss and conclude on automatically merging PRs if all checks (including testing) have passed

(cc) @dave-gantenbein - let me know if I've missed anything and can update 🤗

PR #1147, opened 2025-08-10
chore: add testing documentation and coverage checks

jescalada:
Implements the coverage checks mentioned in #1019, although it doesn't add any auto-merge on coverage attained (we still want maintainers to take a look at larger PRs and decide the right time to merge).

I've also added some documentation on how to implement unit, E2E and fuzz tests for new PRs.

### DIFF
--- a/packages/git-proxy-cli/index.js
+++ b/packages/git-proxy-cli/index.js
@@ -333,6 +333,59 @@ async function reloadConfig() {
   }
 }
 
+/**
+ * Create a new user
+ * @param {string} username The username for the new user
+ * @param {string} password The password for the new user
+ * @param {string} email The email for the new user
+ * @param {string} gitAccount The git account for the new user
+ * @param {boolean} [admin=false] Whether the user should be an admin (optional)
+ */
+async function createUser(username, password, email, gitAccount, admin = false) {
+  if (!fs.existsSync(GIT_PROXY_COOKIE_FILE)) {
+    console.error('Error: Create User: Authentication required');
+    process.exitCode = 1;
+    return;
+  }
+
+  try {
+    const cookies = JSON.parse(fs.readFileSync(GIT_PROXY_COOKIE_FILE, 'utf8'));
+
+    await axios.post(
+      `${baseUrl}/api/auth/create-user`,
+      {
+        username,
+        password,
+        email,
+        gitAccount,
+        admin,
+      },
+      {
+        headers: { Cookie: cookies },
+      },
+    );
+
+    console.log(`User '${username}' created successfully`);
+  } catch (error) {
+    let errorMessage = `Error: Create User: '${error.message}'`;
+    process.exitCode = 2;
+
+    if (error.response) {
+      switch (error.response.status) {
+        case 401:
+          errorMessage = 'Error: Create User: Authentication required';
+          process.exitCode = 3;
+          break;
+        case 400:
+          errorMessage = `Error: Create User: ${error.response.data.message}`;
+          process.exitCode = 4;
+          break;
+      }
+    }
+    console.error(errorMessage);
+  }
+}
+
 // Parsing command line arguments
 yargs(hideBin(process.argv)) // eslint-disable-line @typescript-eslint/no-unused-expressions
   .command({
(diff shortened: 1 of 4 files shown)


website/docs/development/contributing.mdx
```diff
@@ -7,7 +7,7 @@ Here's how to get setup for contributing to GitProxy.
 ## Setup
 The GitProxy project relies on the following pre-requisites:
 
-- [Node](https://nodejs.org/en/download) (16+)
+- [Node](https://nodejs.org/en/download) (22+)
```
dcoric: This should be updated in [`ci.yml`](https://github.com/jescalada/git-proxy/blob/02715f083871c04b7f5895fad9763f134c013199/.github/workflows/ci.yml#L22) as well to 22.x

It would be good to add
`"engines": { "node": ">=22.0.0" }` to `package.json`

And optionally, it would be good to add `.nvmrc` as well

dcoric (changes requested):
Very good in general, I just added a small nitpick.

website/docs/development/contributing.mdx
jescalada: I just noticed that the latest 20 works, so I set the Node version in the website and engines to that one instead! As for the `.nvmrc`, I'd rather keep that clean since not everybody uses `nvm`, and it might cause issues for folks that want to use v22 by default.

website/docs/development/testing.mdx
```diff
@@ -0,0 +1,313 @@
+---
+title: Testing
+---
+
+## Testing
+
+As of v1.19.2, GitProxy uses [Mocha](https://mochajs.org/) (`ts-mocha`) as the test runner, and [Chai](https://www.chaijs.com/) for unit test assertions. End-to-end (E2E) tests are written in [Cypress](https://docs.cypress.io), and some fuzz testing is done with [`fast-check`](https://fast-check.dev/).
```
kriswest: Are the Cypress tests E2E? They're not handling the proxy  but the rather just the admin UI and hence only part of application.

```suggestion
As of v1.19.2, GitProxy uses [Mocha](https://mochajs.org/) (`ts-mocha`) as the test runner, and [Chai](https://www.chaijs.com/) for unit test assertions. User interface tests are written in [Cypress](https://docs.cypress.io), and some fuzz testing is done with [`fast-check`](https://fast-check.dev/).
```

website/docs/development/testing.mdx
```diff
@@ -0,0 +1,313 @@
...
+expect(authMethods[0].type).to.equal('local');
+```
+
+#### Unit testing coverage requirement
+
+**All new lines of code introduced in a PR, must have over 80% coverage** (patch coverage). This is enforced by our CI, and generally a PR will not be merged unless this coverage requirement is met. Please make sure to write thorough unit tests to increase GitProxy's code quality!
+
+If test coverage is still insufficient after writing your tests, check out the [CodeCov report](https://app.codecov.io/gh/finos/git-proxy) after making the PR and take a look at which lines are missing coverage.
+
+### E2E testing with Cypress
```
kriswest: E2E or UI? See previous comment

website/docs/development/testing.mdx
```diff
@@ -0,0 +1,313 @@
...
+    expect(authMethods[0].type).to.equal('local');
+    expect(authMethods[1].type).to.equal('ActiveDirectory');
+    expect(authMethods[2].type).to.equal('openidconnect');
+  });
+});
+```
+
+Core concepts to keep in mind when unit testing JS/TS modules with Chai:
+
+#### Stub internal methods to make tests predictable
```
kriswest, line 89: Need to emphasize resetting sinon and a the proxyquire cache after the tests - in #1164 I spent a ton of time trying to debug a weird issue that turned out to be stubs from other tests leaking into subsequent ones (there may be more cases than I fixed).

website/docs/development/testing.mdx
```diff
@@ -0,0 +1,313 @@
...
+
+// Tests go here
+
+after(async function () {
+  await service.httpServer.close();
+
+  await db.deleteRepo('test-repo');
+  await db.deleteUser('u1');
+  await db.deleteUser('u2');
+});
```
kriswest, line 152: Probably worth including a small sinon example and subsequent reset in after - see earlier comment about emphasizing the reset.

kriswest (changes requested):
Lots of good stuff here! I'd like to see resetting stubs emphasized as I lost a couple of hours to debugging confusing issues caused by stubs that were not reset.

website/docs/development/testing.mdx
jescalada: Good point! I thought of them as E2E since they handle the _entire_ flow from UI to API and back again. But you're right that the proxy code never gets tested here, and there is currently no way to test the proxy through the UI...

website/docs/development/testing.mdx
jescalada: Done!

kriswest:
Might be worth updating the template test and including sinon/proxyquire examples in it: https://github.com/finos/git-proxy/blob/main/test/1.test.js

jescalada:
@kriswest Should be ready for a final look!

test/1.test.js
```diff
@@ -1,24 +1,97 @@
-// This test needs to run first
+/*
+  Template test file. Demonstrates how to:
+  - Use chai-http to test the server
+  - Initialize the server
+  - Stub dependencies with sinon sandbox
+  - Reset stubs after each test
+  - Use proxyquire to replace modules
+  - Clear module cache after a test
+*/
```
kriswest, line 9: ❤️

website/docs/development/testing.mdx
```diff
@@ -0,0 +1,344 @@
...
+afterEach(() => {
+  sinon.restore();
+});
+```
+
+Note that `after` will execute once after **all** the tests are complete, whereas `afterEach` will execute at the end of **each** test. 
+
+#### Reset sinon and proxyquire cache
+
+**It's very important to reset Sinon and the Proxyquire/require cache after each test** when necessary. This prevents old stubs from leaking into subsequent tests.
```
kriswest, line 163: ❤️

kriswest (approved):
LGTM

PR #1154, opened 2025-08-17
test: fix Cypress test data dependency

### LINKED ISSUE #978: Cypress tests require data created by mocha tests, but should be self-contained

jescalada:
Fixes #978.

### Changelog
- Added CSRF token command to use API through Cypress
- Refactored `repo.cy.js` to generate and tear down repo data


cypress/support/commands.js
```diff
@@ -39,3 +39,12 @@ Cypress.Commands.add('login', (username, password) => {
     cy.url().should('include', '/dashboard/repo');
   });
 });
+
+Cypress.Commands.add('getCSRFToken', () => {
+  return cy.request('GET', 'http://localhost:8080/api/v1/repo').then((res) => {
+    const cookies = res.headers['set-cookie'];
+    const csrfCookie = cookies.find(c => c.startsWith('csrf='));
+    const token = decodeURIComponent(csrfCookie.split('=')[1].split(';')[0]);
```
dcoric: This assumes `res.headers['set-cookie']` is present and that a cookie starts with `csrf=`. It would be good to add a guard before running `.split('=')[1]` to avoid runtime errors if something is missing.

dcoric (changes requested):
One small remark, otherwise it looks good.

PR #1158, opened 2025-08-18
fix: flaky fuzz test errors

### LINKED ISSUE #1155: Failing `checkCommitMessages` fuzz test
**Describe the bug**
An interesting bug popped up during one of the CI executions:

Image

Apparently, using the commit message `{"toString": ""}` makes the function behave unexpectedly/error out.

jescalada:
Fixes #1155 and other issues with flaky tests.

Adds the `validator` library to check that emails have a valid format.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@finos/git-proxy",
-  "version": "2.0.0-rc.1",
+  "version": "2.0.0-rc.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@finos/git-proxy",
-      "version": "2.0.0-rc.1",
+      "version": "2.0.0-rc.2",
       "license": "Apache-2.0",
       "workspaces": [
         "./packages/git-proxy-cli"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@finos/git-proxy",
-  "version": "2.0.0-rc.1",
+  "version": "2.0.0-rc.2",
   "description": "Deploy custom push protections and policies on top of Git.",
   "scripts": {
     "cli": "node ./packages/git-proxy-cli/index.js",
--- a/test/ConfigLoader.test.js
+++ b/test/ConfigLoader.test.js
@@ -331,7 +331,6 @@ describe('ConfigLoader', () => {
 
       // Verify the loaded config has expected structure
       expect(config).to.be.an('object');
-      expect(config).to.have.property('proxyUrl');
       expect(config).to.have.property('cookieSecret');
     });
 
@@ -381,7 +380,6 @@ describe('ConfigLoader', () => {
 
       // Verify the loaded config has expected structure
       expect(config).to.be.an('object');
-      expect(config).to.have.property('proxyUrl');
       expect(config).to.have.property('cookieSecret');
     });
 


src/proxy/processors/push-action/checkAuthorEmails.ts
```diff
@@ -1,30 +1,27 @@
 import { Action, Step } from '../../actions';
 import { getCommitConfig } from '../../../config';
 import { Commit } from '../../actions/Action';
+import { isEmail } from 'validator';
 
 const commitConfig = getCommitConfig();
 
 const isEmailAllowed = (email: string): boolean => {
-  if (!email) {
+  if (!email || !isEmail(email)) {
```
kriswest, line 9: This is a definite improvement on validating emails are emails.

However, Its not actually going to prevent the flakiness, only reduce it as the random string generator could still produce something that passes this check (e.g. `a@b.cz`)

kriswest (approved):
This improves the situation and git-proxy's validation of emails. Approving on that basis. 

However, I don't think it fully solves the flakiness in `checkAuthorEmails`, but will reduce the incidence rate by an order of magnitude.

To fully resolve you'd need to check if the string is an email (use the validator and flip the expect) or prevent it ever being one (e.g. by removing @ symbols or add something if one exists that guarantees it doesn't validate).

src/proxy/processors/push-action/checkAuthorEmails.ts
jescalada: I've tried generating 100,000+ emails at a time (several times) and the test didn't fail. It seems whatever the fuzz test generates as "valid" is guaranteed to be caught by the `isEmail` function since they go by the same RFC definitions.

I'm not sure if there's an easy way to check if the email is _real_ or not, if that's what you meant (such as checking that the top-level domain actually exists).

kriswest: I don't think it can be 'guaranteed' but the probability of it happening may have reduced significantly - However, I'm not sure what you mean by the same RFC definitions, isEmail references RFC2822 3.4 + additional special case handling, where [`fc.string()`](https://fast-check.dev/docs/core-blocks/arbitraries/primitives/string/) just produces a random string...

I wasn't suggesting that we check emails are real or not, but rather that you might 'guarantee' that the test is not flaky by either checking the generated string with `validate.isEmail(...)` and flipping the condition you expect (as if it passes that, its always going to pass `CheckAuthorEmails` if no white/blacklist is set) or detect cases that may pass and make sure they don't (add an illegal char such as another @).

Finally, it might also be worth adding a log statement on each failure condition (i.e. when isEmailAllowed is false) in to help debug why the check failed on particular emails? (blacklist, whitelist, not a valid email)

kriswest: Anyway, sounds like the probability is so low it doesn't matter. I've just built up an aversion to flaky tests as, even if low probability, they can snowball over time until the whole set is flaky with a not insignificant probability.

jescalada: Oh, I thought I was using `fc.email()`... My guess as to why it doesn't error out yet is that `fc.string()` deliberately doesn't generate valid emails (?).

I should probably write a test using `fc.email()` along with the logging fixes.

kriswest: You are using fc.email() when generating emails. But fc.string() when just throwing random strings at it in ["should not crash on random string in commit email"](https://github.com/jescalada/git-proxy/blob/802ba275de68943486e0415216bdfffd4d706b8a/test/processors/checkAuthorEmails.test.js#L177).

That's the one that failed for me first as every now and then something passes validation.

However, it failed yesterday 3 times when set to 100 tests, I've run this branch several times against 100k tests and not had a failure. I think we can call it done.

jescalada: Got it! Will merge this PR in if things are good to go. I did try stress testing all the fuzz tests with reasonably long repetitions (100k+ for most) just to check they won't be flaky going forwards. Though unlikely, if something does come up I'll patch it up again. 👍🏼

PR #1164, opened 2025-08-21
feat: don't forward requests for unknown repos

### LINKED ISSUE #1163: Ensure that requests are not forwarded for unknown repositories
**Describe the bug**
In #1043 it was observed that requests are forwarded to github.com for unknown repositories in some cases:

E.g.
```
processing request URL: '/gitlab.com/gitlab-community/meta.git/info/refs?service=git-upload-pack'
proxy keys registered:  ["/github.com/"]
        using fallback
Action processed: Allowed
    Request URL: /gitlab.com/gitlab-community/meta.git/info/refs?service=git-upload-pack
    Host:        127.0.0.1:61485
    User-Agent:  git/2.42.0
Request resolved to https://github.com/gitlab.com/gitlab-community/meta.git/info/refs?service=git-upload-pack
```

This is in the proxy route tests at: https://github.com/finos/git-proxy/blob/967961876fb3cdbb18febc4717e40c565303ec96/test/testProxyRoute.test.js#L427

The test currently passes as the resulting URL is not a project at github.com and hence a 404 is returned through the proxy. However, pull requests - likely any requests - should not be forwarded unless the repository is known/configured in git proxy. This appears to be the intent for pull requests as CheckRepoInAuthorisedList is the only default processor in the pull chain: https://github.com/finos/git-proxy/blob/967961876fb3cdbb18febc4717e40c565303ec96/src/proxy/chain.ts#L26-L28. However, on instrumenting the code it appears NOT to be getting run. Why that is needs to be debugged and new test implemented confirming that you can't fetch or pull a repo from github that is NOT in authorized list.

**To Reproduce**
Run the tests and check the output for test 'should be restarted by the api and stop proxying requests for a host (e.g. gitlab.com) when the last project at that host is DELETED via the API'. "Action processed: Blocked" should appear in the log for that test.

Alternatively, manually construct a URL that git-proxy would produce for a gihub repo, but hasn't as its not configured. Then add that as a remote and attempt to fetch/pull from it.

**Expected behavior**
"Action processed: Blocked" should always appear in the log for operations on unknown repositories.

**Additional context**
There exists a test that confirms we run NO actions for anything other than a pull or push request. Whereas I think we should always check that the repository is known before forwarding *any* request. Otherwise, we risk the proxy being abused in unexpected ways.

kriswest:
resolves #1163 by ensuring that requests are never forwarded on for unknown repositories:
- Resolves issues with tests that were allowing requests to be forwarded for unknown repos
  - Stops stubs from the chain tests leaking into other tests 
  - Removes stubbing from Proxy express application tests
  - Removes chain test that checks that no processors are run unless the action type is "push" or "pull"
  - Adds an integration test that checks if pull requests are forwarded to unknown repos
- Improves handling of errors when executing the chain, ensuring that errors always block pushes
- Changes the HTTP response code for blocked pushes to consistently use 403 instead of 200
- Creates a default action chain that ensures that CheckRepoInAuthList is run for all requests

### DIFF
--- a/src/proxy/chain.ts
+++ b/src/proxy/chain.ts
@@ -45,6 +45,10 @@ export const executeChain = async (req: any, res: any): Promise<Action> => {
         return action;
       }
     }
+  } catch (e) {
+    action.error = true;
+    action.errorMessage = `An error occurred when executing the chain: ${e}`;
+    console.error(action.errorMessage);
   } finally {
     await proc.push.audit(req, action);
     if (action.autoApproved) {
--- a/src/proxy/chain.ts
+++ b/src/proxy/chain.ts
@@ -27,22 +27,23 @@ const pullActionChain: ((req: any, action: Action) => Promise<Action>)[] = [
   proc.push.checkRepoInAuthorisedList,
 ];
 
+const defaultActionChain: ((req: any, action: Action) => Promise<Action>)[] = [
+  proc.push.checkRepoInAuthorisedList,
+];
+
 let pluginsInserted = false;
 
 export const executeChain = async (req: any, res: any): Promise<Action> => {
   let action: Action = {} as Action;
+
   try {
     action = await proc.pre.parseAction(req);
     const actionFns = await getChain(action);
 
     for (const fn of actionFns) {
       action = await fn(req, action);
-      if (!action.continue()) {
-        return action;
-      }
-
-      if (action.allowPush) {
-        return action;
+      if (!action.continue() || action.allowPush) {
+        break;
       }
     }
   } catch (e) {
@@ -93,9 +94,13 @@ export const getChain = async (
     // This is set to true so that we don't re-insert the plugins into the chain
     pluginsInserted = true;
   }
-  if (action.type === 'pull') return pullActionChain;
-  if (action.type === 'push') return pushActionChain;
-  return [];
+  if (action.type === 'pull') {
+    return pullActionChain;
+  } else if (action.type === 'push') {
+    return pushActionChain;
+  } else {
+    return defaultActionChain;
+  }
 };
 
 export default {
@@ -113,6 +118,9 @@ export default {
   },
   get pullActionChain() {
     return pullActionChain;
+  },
+  get defaultActionChain() {
+    return defaultActionChain;
   },
   executeChain,
   getChain,
--- a/test/chain.test.js
+++ b/test/chain.test.js
@@ -95,6 +95,8 @@ describe('proxy chain', function () {
 
   afterEach(() => {
     // Clear the module from the cache after each test
+    sandboxSinon.restore();
+    sinon.restore();
     clearCache(sandboxSinon);
   });
 
--- a/test/chain.test.js
+++ b/test/chain.test.js
@@ -277,17 +277,15 @@ describe('proxy chain', function () {
     expect(mockPushProcessors.audit.called).to.be.true;
   });
 
-  it('executeChain should run no actions if not a push or pull', async function () {
+  it('executeChain should always run at least checkRepoInAuthList', async function () {
     const req = {};
     const action = { type: 'foo', continue: () => true, allowPush: true };
 
-    processors.pre.parseAction.resolves(action);
+    mockPreProcessors.parseAction.resolves(action);
+    mockPushProcessors.checkRepoInAuthorisedList.resolves(action);
 
-    const result = await chain.executeChain(req);
-
-    expect(mockPushProcessors.checkRepoInAuthorisedList.called).to.be.false;
-    expect(mockPushProcessors.parsePush.called).to.be.false;
-    expect(result).to.deep.equal(action);
+    await chain.executeChain(req);
+    expect(mockPushProcessors.checkRepoInAuthorisedList.called).to.be.true;
   });
 
   it('should approve push automatically and record in the database', async function () {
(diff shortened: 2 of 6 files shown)


kriswest:
@finos/git-proxy-maintainers I figured out several reasons why the proxy will apparently forward requests for unknown repos:
- stubbing in tests that wasn't needed
- stubbing that was leaking from the chain tests because it wasn't reset - I've fixed one case, there may be others
- no processors are run unless the action type is push or pull
- errors thrown during the execution of the chain weren't being caught

I've tried to address each of these different issues, resulting in a new default action chain that only contains CheckRepoInAuthList (so that git-proxy will not forward requests for repos that it isn't aware of, of any type). I also switched the response code on error/block to 403 (forbidden).

src/proxy/chain.ts
```diff
@@ -27,24 +27,29 @@ const pullActionChain: ((req: any, action: Action) => Promise<Action>)[] = [
...
-        return action;
-      }
-
-      if (action.allowPush) {
-        return action;
+      if (!action.continue() || action.allowPush) {
+        break;
       }
     }
+  } catch (e) {
```
kriswest, line 49: This is critical as, if a processor throws an error while executing we can jump out of the loop into the finally and not run subsequent stages AND not update the action to show the error before writing it into the DB in the finally.

src/proxy/chain.ts
```diff
@@ -27,24 +27,29 @@ const pullActionChain: ((req: any, action: Action) => Promise<Action>)[] = [
   proc.push.checkRepoInAuthorisedList,
 ];
 
+const defaultActionChain: ((req: any, action: Action) => Promise<Action>)[] = [
```
kriswest, line 30: Adding this to ensure that all requests are validated against the allowed repos, regardless of type, reducing the likelihood of the proxy being abused to forward request onto Github for arbitrary/unknown repos.

test/chain.test.js
```diff
@@ -95,6 +95,8 @@ describe('proxy chain', function () {
 
   afterEach(() => {
     // Clear the module from the cache after each test
+    sandboxSinon.restore();
```
jescalada: I think calling `sandbox.restore()` might be redundant when already calling `.reset()`. Some details on the [difference between the two](https://javascript.plainenglish.io/when-should-you-use-sinons-restore-and-reset-functions-bdbd39550a2f):

> The restore functions in sinon only work on spies and stubs that replace an existing method, and our spy was created as an anonymous (standalone) spy.
>
> Therefore, restore does not reset the call count, which will continue increasing as the spy function is called in subsequent tests.
> 
> `reset`: for stubs, resets both behavior and history; for sandboxes, resets the internal states of everything created in the sandbox.

I'm wondering if this means we should go and replace all usages of `stub.restore()` and `sandbox.restore()` with `reset()` to get the right behaviour.

jescalada: On another note, removing the `sandboxSinon.restore()` bit seems to make a few apparently unrelated tests fail...

```
3 failing

  1) proxy express application
       should proxy requests for the default GitHub repository:

      AssertionError: expected 403 to equal 200
      + expected - actual

      -403
      +200
      
      at Context.<anonymous> (test/testProxyRoute.test.js:379:27)
      at fulfilled (test/testProxyRoute.test.js:5:58)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)

  2) proxy express application
       should proxy requests for the default GitHub repository using the fallback URL:

      AssertionError: expected 403 to equal 200
      + expected - actual

      -403
      +200
      
      at Context.<anonymous> (test/testProxyRoute.test.js:392:27)
      at fulfilled (test/testProxyRoute.test.js:5:58)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)

  3) proxy express application
       should be restarted by the api and proxy requests for a new host (e.g. gitlab.com) when a project at that host is ADDED via the API:

      AssertionError: expected Response{ domain: null, …(35), …(2) } to have status code 200 but got 403
      + expected - actual

      -403
      +200
      
      at Context.<anonymous> (test/testProxyRoute.test.js:430:22)
      at fulfilled (test/testProxyRoute.test.js:5:58)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)

```

jescalada (approved):
Looks good to me! I did some experiments with removing the `sinon` restore/reset calls but it seems they're necessary (or our tests in `testProxyRoute` might need some adjustment).

test/chain.test.js
kriswest: Yes this is what I ran into - the stubs remain in place and cause the proxy to reject requests when it should be accepting in these cases. My understanding of the sandbox is that it simply groups the various stubs and spies etc. so we can restore/reset them all together - perhaps thats a pattern we need to make more consistent? Advice on using a sandbox and restore/reset may be a good addition to your test guide #1147.

On reset vs restore I note that the author of the article is a little confused by a similar problem: 

> Bear in mind: adding sandbox.reset() in the afterEach hook will not help, but it works if we add it to the end of the first unit test — still haven’t figured out why this is happening.

jescalada: Interesting, I removed the `sandbox.reset()` line in the `clearCache` function and the tests are still passing. Perhaps calling `reset()` in the `afterEach` is ineffective after all?

kriswest: I've gone ahead and switched to just calling sandbox.restore() in the clearCache function (the `sinon.restore()` is redundant so I removed it). I remain confused as to why `reset()` doesn't work, but then as noted above, I'm not the only one!

@jescalada if this all looks fine to you can you resolve this conversation (it's blocking merge) and feel free to hit merge thereafter!

Issue #1041, opened 2025-06-03
Validate attestation against config


kriswest:
**Describe the bug**

The API currently validates attestations submitted against the definition of the attestation form submitted with the request:
https://github.com/finos/git-proxy/blob/43876844b547dc29092143c420f265ddf6fd3a74/src/service/routes/push.js#L89

To ensure attestations are completed correctly, they should be validated against the requirements set in the configuration, rather than the copy submitted with the request.

**Expected behaviour**
If a request is submitted with an incomplete attestation (i.e. with an element missing, rather than just an element that's not checked).

PR #1166, opened 2025-08-24
refactor: `service` module to TS + ESM (v2)

jescalada:
Draft PR for `service` refactor. Still fixing failing tests!


.eslintrc.json
```diff
@@ -52,7 +52,8 @@
     "@typescript-eslint/no-explicit-any": "off", // temporary until TS refactor is complete
     "@typescript-eslint/no-unused-vars": "off", // temporary until TS refactor is complete
     "@typescript-eslint/no-require-imports": "off", // prevents error on old "require" imports
-    "@typescript-eslint/no-unused-expressions": "off" // prevents error on test "expect" expressions
+    "@typescript-eslint/no-unused-expressions": "off", // prevents error on test "expect" expressions
+    "new-cap": "off" // prevents errors on express.Router()
```
06kellyjac:
```suggestion
    new-cap: ["error", { "capIsNewExceptionPattern": "^express\\.." }]
```

src/db/file/users.ts
```diff
@@ -115,11 +115,10 @@ export const deleteUser = (username: string): Promise<void> => {
...
 };
 
-export const updateUser = (user: User): Promise<void> => {
-  user.username = user.username.toLowerCase();
-  if (user.email) {
-    user.email = user.email.toLowerCase();
-  }
+export const updateUser = (user: Partial<User>): Promise<void> => {
+  if (user.username) user.username = user.username.toLowerCase();
+  if (user.email) user.email = user.email.toLowerCase();
```
06kellyjac:
```suggestion
  if (user.username) {
    user.username = user.username.toLowerCase();
  };
  if (user.email) {
    user.email = user.email.toLowerCase();
  };
```

I'd suggest avoiding shorthand `if` for consistency. Ternary operator is a readable shorthand for where `if else` is necessary.

06kellyjac: Also matches `src/db/mongo/users.ts`

src/service/index.ts
```diff
@@ -1,19 +1,20 @@
...
-const http = require('http');
-const cors = require('cors');
-const app = express();
-const path = require('path');
-const config = require('../config');
-const db = require('../db');
-const rateLimit = require('express-rate-limit');
-const lusca = require('lusca');
-const configLoader = require('../config/ConfigLoader');
+import express, { Express } from 'express';
```
06kellyjac, line 1:
```suggestion
import express, { type Express } from 'express';
```

Explicit type import can avoid using classes and help with treeshaking.
Not super necessary, especially for internal code, but probably easy to keep up with for larger libraries like express.

src/service/index.ts
```diff
@@ -23,57 +24,22 @@ const corsOptions = {
 
 /**
  * Internal function used to bootstrap the Git Proxy API's express application.
- * @param {proxy} proxy A reference to the proxy express application, used to restart it when necessary.
+ * @param {Express} proxy A reference to the proxy express application, used to restart it when necessary.
  * @return {Promise<Express>}
  */
-async function createApp(proxy) {
+async function createApp(proxy: Express) {
```
06kellyjac: Type may need adjusting, please see `start` below.

src/service/index.ts
```diff
@@ -23,57 +24,22 @@ const corsOptions = {
...
-      }
-
-      res.status(500).json({ error: 'Failed to reload configuration' });
-    }
-  });
-
   app.use(
     session({
-      store: config.getDatabase().type === 'mongo' ? db.getSessionStore(session) : null,
+      store: config.getDatabase().type === 'mongo' ? db.getSessionStore() || undefined : undefined,
```
06kellyjac: Thoughts on adjusting `getSessionStore` to return `undefined` rather than `null`, then we'll already have the type.

Also if we adjust `sink` to require `getSessionStore: () => MongoDBStore | undefined;`

https://github.com/finos/git-proxy/blob/b2b1b145432492cee11c998d1db2ad7fc5b7f287/src/db/types.ts#L65

We can then add `getSessionStore = () => undefined;` to the fs implementation and simplify this down to:

```suggestion
      store: db.getSessionStore(),
```

src/service/index.ts
```diff
@@ -113,10 +79,10 @@ async function createApp(proxy) {
 
 /**
  * Starts the proxy service.
- * @param {proxy?} proxy A reference to the proxy express application, used to restart it when necessary.
+ * @param {*} proxy A reference to the proxy express application, used to restart it when necessary.
  * @return {Promise<Express>} the express application (used for testing).
  */
-async function start(proxy) {
+async function start(proxy: any) {
```
06kellyjac:
```suggestion
async function start(proxy?: Express) {
```

`createApp` explicitly wants type `Express`, and `proxy` is what we give it. Currently `proxy` should be `Express` or `createApp` etc needs adjusting.

---

Also not a matter for this pr but line 86 is checking `proxy` is non-truthy, not that it's `null` so we either need to do `proxy != null` or adjust the warning.
I don't believe I know the context for that warning log.
I bring it up because it may relate to some typing we do later on.

src/service/passport/activeDirectory.ts
```diff
@@ -43,8 +58,7 @@ const configure = (passport) => {
               const message = `User it not a member of ${userGroup}`;
               return done(message, null);
             }
-          } catch (err) {
-            console.log('ad test (isUser): e', err);
+          } catch (err: any) {
```
06kellyjac, line 66: I'm in favor of landing this PR with this as-is with `err: any` but for the next PR where we trim back `any` we'll want

```suggestion
          } catch (err: unknown) {
```

+

```js
if (e instanceof Error) {
  // some code using `e.message`
}

File: src/service/passport/index.ts [hunk 8]
```diff
@@ -0,0 +1,39 @@
+import passport, { PassportStatic } from 'passport';
```
06kellyjac:
```suggestion
import passport, { type PassportStatic } from 'passport';
```

src/service/passport/jwtAuthHandler.ts
```diff
@@ -0,0 +1,73 @@
+import { assignRoles, validateJwt } from './jwtUtils';
+import { Request, Response, NextFunction } from 'express';
```
06kellyjac:
```suggestion
import type { Request, Response, NextFunction } from 'express';
```

src/service/passport/jwtAuthHandler.ts
```diff
@@ -0,0 +1,73 @@
...
+      ? [{ type: 'jwt', enabled: true, jwtConfig: overrideConfig }]
+      : getAPIAuthMethods();
+
+    const jwtAuthMethod = apiAuthMethods.find((method) => method.type.toLowerCase() === type);
+
+    if (!overrideConfig && (!jwtAuthMethod || !jwtAuthMethod.enabled)) {
+      return next();
+    }
+
+    if (req.isAuthenticated?.()) {
```
06kellyjac:
```suggestion
    if (req.isAuthenticated && req.isAuthenticated()) {
```

Personally I don't like optional chaining for functions but I defer to the group.

At the least we should check we have [no-unsafe-optional-chaining](https://eslint.org/docs/latest/rules/no-unsafe-optional-chaining) enabled for eslint at some point (separate PR)

src/service/passport/jwtUtils.ts
```diff
@@ -0,0 +1,99 @@
+import axios from 'axios';
+import jwt, { JwtPayload } from 'jsonwebtoken';
```
06kellyjac:
```suggestion
import jwt, { type JwtPayload } from 'jsonwebtoken';
```

src/service/passport/jwtUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+      throw new Error('No matching key found in JWKS');
+    }
+
+    const pubKey = jwkToPem(jwk as any);
+
+    const verifiedPayload = jwt.verify(token, pubKey, {
+      algorithms: ['RS256'],
+      issuer: authorityUrl,
+      audience: expectedAudience,
+    }) as JwtPayload;
```
06kellyjac: We'll want to loop back on the usage of `as` here but best to get most of the changes in

src/service/passport/ldaphelper.ts
```diff
@@ -1,16 +1,33 @@
...
+export const isUserInAdGroup = (
+  req: Request,
+  profile: { username: string },
+  ad: AD,
+  domain: string,
+  name: string,
+): Promise<boolean> => {
   // determine, via config, if we're using HTTP or AD directly
-  if (thirdpartyApiConfig?.ls?.userInADGroup) {
+  if ((thirdpartyApiConfig?.ls as any).userInADGroup) {
```
06kellyjac: Why is an `as any` needed here if we're alreayd doing `?`
Just curious what the specific type error is we're trying to avoid so we can either fix it now or context to loop back on.

src/service/passport/ldaphelper.ts
```diff
@@ -24,8 +41,8 @@ const isUserInAdGroupViaAD = (req, profile, ad, domain, name) => {
   });
 };
 
-const isUserInAdGroupViaHttp = (id, domain, name) => {
-  const url = String(thirdpartyApiConfig.ls.userInADGroup)
+const isUserInAdGroupViaHttp = (id: string, domain: string, name: string): Promise<boolean> => {
+  const url = String((thirdpartyApiConfig?.ls as any).userInADGroup)
```
06kellyjac: same as above

src/service/passport/oidc.ts
```diff
@@ -0,0 +1,130 @@
...
+import { getAuthMethods } from '../../config';
+import { UserInfoResponse } from './types';
+
+export const type = 'openidconnect';
+
+export const configure = async (passport: PassportStatic): Promise<PassportStatic> => {
+  // Use dynamic imports to avoid ESM/CommonJS issues
+  const { discovery, fetchUserInfo } = await import('openid-client');
+  // @ts-expect-error - throws error due to missing type definitions
+  const { Strategy } = await import('openid-client/passport');
```
06kellyjac, line 12: I thought `openid-client` is typescript?
Or are we on a too old version or something?

src/service/passport/types.ts
```diff
@@ -0,0 +1,70 @@
...
+export type AD = {
+  isUserMemberOf: (
+    username: string,
+    groupName: string,
+    callback: (err: Error | null, isMember: boolean) => void,
+  ) => void;
+};
+
+/**
+ * The UserInfoResponse type from openid-client (to fix some type errors)
```
06kellyjac: Some details as to what the errors were would help with review

src/service/routes/push.ts
```diff
@@ -33,15 +35,15 @@ router.get('/:id', async (req, res) => {
...
+router.post('/:id/reject', async (req: Request, res: Response) => {
   if (req.user) {
     const id = req.params.id;
 
     // Get the push request
     const push = await db.getPush(id);
 
     // Get the committer of the push via their email
-    const committerEmail = push.userEmail;
+    const committerEmail = push?.userEmail;
```
06kellyjac: If an email is missing we shouldn't spend time looking for the email in the DB.
A similar 401 to below would be good

src/service/routes/push.ts
```diff
@@ -97,7 +102,7 @@ router.post('/:id/authorise', async (req, res) => {
     console.log({ push });
 
     // Get the committer of the push via their email address
-    const committerEmail = push.userEmail;
+    const committerEmail = push?.userEmail;
```
06kellyjac, line 107: Same as above

src/service/passport/jwtAuthHandler.ts
jescalada: Your version looks cleaner so I'll go with that. 👍🏼

src/service/index.ts
jescalada: Replaced with `Proxy` type on both!

src/service/index.ts
jescalada: Pinging @kriswest for awareness - this is what we're currently doing in `service.start()`:

```ts
/**
 * Starts the proxy service.
 * @param {Proxy} proxy A reference to the proxy, used to restart it when necessary.
 * @return {Promise<Express>} the express application (used for testing).
 */
async function start(proxy: Proxy) {
  if (!proxy) { // Purpose of this check?
    console.warn("WARNING: proxy is null and can't be controlled by the API service");
  }

  const app = await createApp(proxy);

  _httpServer.listen(uiPort);

  console.log(`Service Listening on ${uiPort}`);
  app.emit('ready');

  return app;
}
```

src/service/passport/jwtUtils.ts
jescalada: Fixed this one up - apparently string payloads are exceptional/nonstandard so we can just exclude them with a check.

src/service/passport/ldaphelper.ts
jescalada: If we do `thirdpartyApiConfig?.ls?.userInADGroup` we get:
```
Property 'userInADGroup' does not exist on type '{}'.ts(2339)
```

Meanwhile, doing `thirdpartyApiConfig?.ls.userInADGroup` gives:
```
'thirdpartyApiConfig.ls' is of type 'unknown'.ts(18046)
```

The problem is that `getAPIs` is defined to return `Record<string, unknown>`, since only the following sample is available and it doesn't contain the `thirdPartyApiConfig` or `ls` parameters:

```json
"api": {
  "github": {
    "baseUrl": "https://api.github.com"
  }
},
```

src/service/passport/oidc.ts
jescalada: The problem here is that using `await import('openid-client/passport')` gives the following error:

```
Cannot find module 'openid-client/passport' or its corresponding type declarations.
  There are types at '/home/juan/Desktop/Juan/Projects/git-proxy/node_modules/openid-client/build/passport.d.ts', but this result could not be resolved under your current 'moduleResolution' setting. Consider updating to 'node16', 'nodenext', or 'bundler'.ts(2307)
```

I tried solving this by updating the `tsconfig.json` to `nodenext`, however this makes a bunch of tests fail...

On the other hand, using `await import('openid-client/build/passport')` made the OIDC flow break.

jescalada: I also tried using other `moduleResolution` values without luck...

jescalada: Also, the `openid-client` version is the latest, I think the problem is just the resolution for the types.

src/service/passport/types.ts
jescalada: I found a fix for this one by simply importing the `UserInfoResponse` from `openid-client`. I thought it came from `openid-client/passport` so I didn't try importing it directly.

src/service/routes/push.ts
jescalada: I updated it by adding the following which is hopefully enough:

```ts
if (!push) {
  res.status(404).send({
    message: `Push request not found`,
  });
  return;
}

if (!push.userEmail) {
  res.status(400).send({
    message: `Push request has no user email`,
  });
  return;
}
```

jescalada: I extracted it into a function actually:

```ts

async function getValidPushOrRespond(id: string, res: Response) {
  const push = await db.getPush(id);
  console.log({ push });

  if (!push) {
    res.status(404).send({ message: `Push request not found` });
    return null;
  }

  if (!push.userEmail) {
    res.status(400).send({ message: `Push request has no user email` });
    return null;
  }

  return push;
}
```

And then I'm simply calling this on both cases where we need the `push.userEmail`:

```ts
// Get the push request
if (!push) return;

```

src/service/passport/activeDirectory.ts
jescalada: Made an issue for this: #1174

src/config/types.ts
```diff
@@ -49,6 +51,39 @@ export interface Authentication {
   type: string;
   enabled: boolean;
   options?: Record<string, unknown>;
+  oidcConfig?: OidcConfig;
```
kriswest: At some point we should document these types and include the deprecation annotations where needed

src/db/file/index.ts
```diff
@@ -1,6 +1,9 @@
 import * as users from './users';
 import * as repo from './repo';
 import * as pushes from './pushes';
+import * as helper from './helper';
+
+export const { getSessionStore } = helper;
```
kriswest, line 6: Why do we need this? We only appear to use the sessionStore when running mongo and in one place (to pass to expressSession). Are we making use of sessions in the API properly? If so should we be implementing this properly for the filedb?  See https://www.npmjs.com/package/express-session#session-store-implementation. Otherwise perhaps we should rip it out.

src/db/file/users.ts
```diff
@@ -115,11 +115,14 @@ export const deleteUser = (username: string): Promise<void> => {
   });
 };
 
-export const updateUser = (user: User): Promise<void> => {
-  user.username = user.username.toLowerCase();
+export const updateUser = (user: Partial<User>): Promise<void> => {
+  if (user.username) {
```
kriswest, line 120: good catch

src/db/mongo/pushes.ts
```diff
@@ -10,9 +10,12 @@ const defaultPushQuery: PushQuery = {
   blocked: true,
   allowPush: false,
   authorised: false,
+  type: 'push',
```
kriswest, line 13: I realise on seeing this, that we're storing other types of operation in the pushes table. Thats not necessarily a bad thing, but worth being aware of. I see many `"type"`:"default" but no `"type":"pull"`... on investigation it appears they arrive as POSTs rather than GETS as this code expects:

https://github.com/finos/git-proxy/blob/11ec0cae5dd0f76cb0384dc8568a0c614c0d6330/src/proxy/processors/pre-processor/parseAction.ts#L15

That is probably beyond the scope of this PR and needs a new issue raised.

kriswest: Created issue #1185 for this

src/db/mongo/users.ts
```diff
@@ -50,8 +50,10 @@ export const createUser = async function (user: User): Promise<void> {
   await collection.insertOne(user as OptionalId<Document>);
 };
 
-export const updateUser = async (user: User): Promise<void> => {
-  user.username = user.username.toLowerCase();
+export const updateUser = async (user: Partial<User>): Promise<void> => {
+  if (user.username) {
```
kriswest, line 54: 👍

src/service/index.ts
```diff
@@ -0,0 +1,113 @@
...
+        },
+        hsts: { maxAge: 31536000, includeSubDomains: true, preload: true },
+        nosniff: true,
+        referrerPolicy: 'same-origin',
+        xframe: 'SAMEORIGIN',
+        xssProtection: true,
+      }),
+    );
+  }
+  app.use(passport.initialize());
```
kriswest, line 71: I note you've removed this:

```
// Add new admin-only endpoint to reload config
  app.post('/api/v1/admin/reload-config', async (req, res) => {
```

Just checking that that was intentional and whether there are any other docs etc. changes needed because of that.

src/service/passport/activeDirectory.ts
```diff
@@ -1,18 +1,33 @@
...
   const config = authMethods.find((method) => method.type.toLowerCase() === type);
+
+  if (!config || !config.adConfig) {
+    throw new Error('AD authentication method not enabled');
+  }
+
   const adConfig = config.adConfig;
 
-  const { userGroup, adminGroup, domain } = config;
+  if (!adConfig) {
```
kriswest: Already checking if adConfig is defined on line 13, so this is redundant/won't be hit.

src/service/passport/activeDirectory.ts
```diff
@@ -24,7 +39,7 @@ const configure = (passport) => {
         integrated: false,
         ldap: adConfig,
       },
-      async function (req, profile, ad, done) {
+      async function (req: any, profile: any, ad: any, done: (err: any, user: any) => void) {
```
kriswest: Limited info to type this more tightly at: https://www.passportjs.org/packages/[passport-activedirectory](https://www.passportjs.org/packages/passport-activedirectory/)/ and https://github.com/bhoriuchi/passport-activedirectory/blob/master/src/strategy.js

probably not worth the trouble as its an untyped js lib.

src/service/passport/ldaphelper.ts
kriswest: I think the typing for getAPIs needs tightening in that case, with optional properties for ls and github. The type details should come from the schema: https://github.com/finos/git-proxy/blob/11ec0cae5dd0f76cb0384dc8568a0c614c0d6330/config.schema.json#L15-L43

I still haven't had time to engage with @fabiovincenzi's work on using quicktype - but it could generate the type objects from the schema...

src/service/passport/ldaphelper.ts
kriswest: typing thirdpartyApiConfig would fix this

src/service/passport/local.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+  });
+
+  return passport;
+};
+
+/**
+ * Create the default admin user if it doesn't exist
+ */
+export const createDefaultAdmin = async () => {
```
kriswest, line 54: we ought to drive this from config at some point...

src/service/passport/types.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * {
+ *   "admin": {
+ *     "name": "John Doe"
+ *   }
+ * }
+ */
+export type RoleMapping = Record<string, Record<string, string>>;
+
+export type AD = {
```
kriswest: Apparently these are the types ActiveDirectory: https://www.npmjs.com/package/@types/activedirectory2

They are linked to: https://github.com/jsumners/node-activedirectory#readme which is no longer maintained, and that was a fork of https://github.com/gheeres/node-activedirectory

all very dated and not well maintained.

src/service/routes/auth.ts
```diff
@@ -47,10 +54,10 @@ const getLoginStrategy = () => {
   return enabledAppropriateLoginStrategies[0].type.toLowerCase();
 };
 
-const loginSuccessHandler = () => async (req, res) => {
+const loginSuccessHandler = () => async (req: Request, res: Response) => {
   try {
-    const currentUser = { ...req.user };
-    delete currentUser.password;
+    const currentUser = { ...req.user } as User;
+    delete (currentUser as any).password;
```
kriswest: why the `any`, `User` has the password field...

src/service/routes/push.ts
```diff
@@ -13,15 +15,15 @@ router.get('/', async (req, res) => {
     if (k === 'limit') continue;
     if (k === 'skip') continue;
     let v = req.query[k];
-    if (v === 'false') v = false;
-    if (v === 'true') v = true;
-    query[k] = v;
+    if (v === 'false') v = false as any;
+    if (v === 'true') v = true as any;
+    query[k as keyof PushQuery] = v as any;
```
kriswest: Can `v` be typed better to avoid the `any`s?

src/service/routes/push.ts
```diff
@@ -33,12 +35,13 @@ router.get('/:id', async (req, res) => {
...
 
-router.post('/:id/reject', async (req, res) => {
+router.post('/:id/reject', async (req: Request, res: Response) => {
   if (req.user) {
     const id = req.params.id;
 
     // Get the push request
-    const push = await db.getPush(id);
+    const push = await getValidPushOrRespond(id, res);
+    if (!push) return;
```
kriswest: Should we respond here with a 401 or 404? Might be necessary for an error to display in the UI. Otherwise this post request is just going to hang idle

src/service/routes/push.ts
```diff
@@ -51,19 +54,22 @@ router.post('/:id/reject', async (req, res) => {
       return;
     }
 
-    if (list[0].username.toLowerCase() === req.user.username.toLowerCase() && !list[0].admin) {
+    if (
+      list[0].username.toLowerCase() === (req.user as any).username.toLowerCase() &&
```
kriswest: should be able to eliminate the `any` here and several more below by pulling req.user.username into typed (string) variable - can happily check its type in the if statement up on line 39 as well.

src/service/routes/push.ts
```diff
@@ -77,24 +83,23 @@ router.post('/:id/reject', async (req, res) => {
...
+router.post('/:id/authorise', async (req: Request, res: Response) => {
   console.log({ req });
 
   const questions = req.body.params?.attestation;
   console.log({ questions });
 
   // TODO: compare attestation to configuration and ensure all questions are answered
   // - we shouldn't go on the definition in the request!
-  const attestationComplete = questions?.every((question) => !!question.checked);
+  const attestationComplete = questions?.every((question: any) => !!question.checked);
```
kriswest: Removing the `any` here would be good - but I think we'd need to do whats specified in the TODO - the attestation applied should come from the config and not the request.

src/service/routes/push.ts
```diff
@@ -156,18 +164,20 @@ router.post('/:id/authorise', async (req, res) => {
   }
 });
 
-router.post('/:id/cancel', async (req, res) => {
+router.post('/:id/cancel', async (req: Request, res: Response) => {
   if (req.user) {
     const id = req.params.id;
 
-    const isAllowed = await db.canUserCancelPush(id, req.user.username);
+    const isAllowed = await db.canUserCancelPush(id, (req.user as any).username);
```
kriswest: and again, replace the any by checking req.user.username and pull into a variable

src/service/routes/repo.ts
```diff
@@ -1,43 +1,43 @@
...
     const proxyURL = getProxyURL(req);
     const _id = req.params.id;
     const qd = await db.getRepoById(_id);
     res.send({ ...qd, proxyURL });
   });
 
-  router.patch('/:id/user/push', async (req, res) => {
-    if (req.user && req.user.admin) {
+  router.patch('/:id/user/push', async (req: Request, res: Response) => {
+    if (req.user && (req.user as any).admin) {
```
kriswest: creating some typeguards might be a good way to eliminate these `any`s

test/testPush.test.js
```diff
@@ -314,6 +314,34 @@ describe('auth', async () => {
         .set('Cookie', `${cookie}`);
       res.should.have.status(401);
     });
+
+    it('should fetch all pushes', async function () {
+      await db.writeAudit(TEST_PUSH);
+      await loginAsApprover();
+      const res = await chai.request(app).get('/api/v1/push').set('Cookie', `${cookie}`);
+      res.should.have.status(200);
+      res.body.should.be.an('array');
```
kriswest, line 323: should you check that there is content in the array? particularly the element you write to the DB at the beginning of the test

test/testPush.test.js
```diff
@@ -314,6 +314,34 @@ describe('auth', async () => {
...
+    });
+
+    it('should allow a committer to cancel a push', async function () {
+      await db.writeAudit(TEST_PUSH);
+      await loginAsCommitter();
+      const res = await chai
+        .request(app)
+        .post(`/api/v1/push/${TEST_PUSH.id}/cancel`)
+        .set('Cookie', `${cookie}`);
+      res.should.have.status(200);
```
kriswest, line 338: should you check that the status of the TEST_PUSH changed to cancelled?

test/testPush.test.js
```diff
@@ -314,6 +314,34 @@ describe('auth', async () => {
...
+    });
+
+    it('should not allow a non-committer to cancel a push (even if admin)', async function () {
+      await db.writeAudit(TEST_PUSH);
+      await loginAsAdmin();
+      const res = await chai
+        .request(app)
+        .post(`/api/v1/push/${TEST_PUSH.id}/cancel`)
+        .set('Cookie', `${cookie}`);
+      res.should.have.status(401);
```
kriswest, line 354: should you check status did not change?

kriswest (commented):
You're doing a great job with this conversion. Added some suggestions to tighten further.

src/service/passport/ldaphelper.ts
jescalada: Added a temporary typing fix for this! I believe Fabio's PR might fix all the config typing issues (`Record<string, unknown>`  and such).

src/config/types.ts
jescalada: Right now the `oidcConfig` type is documented in the `config.schema.json`, however it'd be better if we could have a single `options` value for all possible authentication methods.

I like how the `config.schema.json` allows setting `"oneOf"` to allow only one of the defined options. If the config typing PR can interpret that correctly, we could reuse the generated types instead.

Otherwise we could have something like `config: AdConfig | OidcConfig | JwtConfig` but I feel this isn't very clean.

src/db/file/index.ts
jescalada: I did some investigating here, and it seems that we're actually using the obtained `MongoDBStore` when calling passport.use(session). What's interesting is that when using the file DB (so `store: null`) we didn't get any errors. This might be because the `express-session` is defaulting to a non-production `MemoryStore` and putting the sessions there, with the caveat that restarting the server would make users lose their active sessions.

Made an issue for this to push this PR forward: #1186

src/service/index.ts
jescalada: This was part of an unfinished feature for SSH which was missing the `ConfigLoader.loadConfiguration()` function (so I decided to remove it for now until a fully usable and TS-compliant SSH flow is done). The completed flow is implemented in one of the SSH PRs which will need some extra refactoring for the TS improvements here.

jescalada: I think we should also remove it from the `createApp` function 😅

src/service/passport/activeDirectory.ts
jescalada: Oops!

src/service/passport/activeDirectory.ts
jescalada: Added some custom types for AD passport since the previous ones were a bit haphazard. 🙂

src/service/passport/local.ts
jescalada: Issue available: #1022. We can prioritize this for an eventual release after v2.0.0.

src/service/passport/types.ts
jescalada: I was able to improve the `AD` typing, but it seems that `activedirectory2` doesn't have the other types for profile, callbacks and such...

src/service/routes/auth.ts
jescalada: This is because `delete` can't be called unless the field is optional. Looking into it though, I noticed that deleting the password is pointless if we're already using `toPublicUser`. I refactored it so that `toPublicUser` gets called first, then the clean user is logged and sent in the response.

```ts
const loginSuccessHandler = () => async (req: Request, res: Response) => {
  try {
    const currentUser = toPublicUser({ ...req.user });
    console.log(
      `serivce.routes.auth.login: user logged in, username=${
        currentUser.username
      } profile=${JSON.stringify(currentUser)}`,
    );
    res.send({
      message: 'success',
      user: currentUser,
    });
  } catch (e) {
    console.log(`service.routes.auth.login: Error logging user in ${JSON.stringify(e)}`);
    res.status(500).send('Failed to login').end();
  }
};
```

src/service/routes/push.ts
jescalada: This took me a while, but I found a solution (although we will likely have to improve the whole "filtering" system down the line to prevent unnecessary conversions and further narrow down the "allowed" filter keywords):

```ts
router.get('/', async (req: Request, res: Response) => {
  const query: Partial<UserQuery> = {};

  console.log(`fetching users = query path =${JSON.stringify(req.query)}`);
  for (const k in req.query) {
    if (!k) continue;
    if (k === 'limit' || k === 'skip') continue;

    const rawValue = req.query[k];
    let parsedValue: boolean | undefined;
    if (rawValue === 'false') parsedValue = false;
    if (rawValue === 'true') parsedValue = true;
    query[k] = parsedValue ?? rawValue?.toString();
  }

  const users = await db.getUsers(query);
  res.send(users.map(toPublicUser));
});
```

jescalada: I also fixed up the `users.ts` and `repo.ts` routes!

src/service/routes/push.ts
jescalada: So this is how the frontend handles the `/push/:id/reject` endpoint:

```ts
const rejectPush = async (id, setMessage, setUserAllowedToReject) => {
  const url = `${baseUrl}/push/${id}/reject`;
  let errorMsg = '';
  let isUserAllowedToReject = true;
  await axios.post(url, {}, getAxiosConfig()).catch((error) => {
    if (error.response && error.response.status === 401) {
      errorMsg = 'You are not authorised to reject...';
      isUserAllowedToReject = false;
    }
  });
  await setMessage(errorMsg);
  await setUserAllowedToReject(isUserAllowedToReject);
};
```

I think this is going to require some deeper refactoring, so I've made a separate issue for it: #1188

src/service/routes/push.ts
jescalada: Removed the `any` here! I'd like to leave the TODO for another PR (we already have this issue to track it, and hopefully we'll get around to it for an eventual v2 release #1041).

src/service/passport/types.ts
kriswest: Doesn't surprise ms

jescalada:
Finally answered/fixed up all the comments here! Would love to get one last look 😃 @kriswest @06kellyjac

src/config/types.ts
kriswest: > I like how the `config.schema.json` allows setting `"oneOf"` to allow only one of the defined options. If the config typing PR can interpret that correctly, we could reuse the generated types instead.

You have to be careful with `oneOf` as it means 'only one of', i.e. all the types must be mutually exclusive - if the element could map to/validate again more than one of the types then it does not validate. In that situation `anyOf` is must be used. When using quicktype to translate from jsonSchema to typescript some of the nuance may be lost and a union type as above is the best you can hope for (the worst being that they all get merged into a single composite type with optional fields for everything - still a lot better than nothing, but allows a few edge cases through!). This is all possible to sidestep by designing the config elements to be unique of course.

I think generating config types from quick type will be super valuable however, as we should be able to just apply it at the top-level of the config and let it flow down through all the functions and elements.

config.schema.json
```diff
@@ -173,6 +173,14 @@
           }
         }
       }
+    },
+    "smtpHost": {
```
kriswest: should this go if nodemailer is removed? If it will stop existing configs with it from validating might just be a case of saying its not currently used in the description

proxy.config.json
```diff
@@ -182,5 +182,7 @@
         "loginRequired": true
       }
     ]
-  }
+  },
+  "smtpHost": "",
```
kriswest: See previous comment: Should this be removed if nodemailer is gone?

kriswest (commented):
A couple of quick comments - will try to complete a pass later on

config.schema.json
jescalada: I added all references to SMTP in this PR, so I guess it wouldn't hurt to remove them entirely. I just don't know if the `sendEmail.js` was ever being used previously since it used the config but the SMTP values were never present in `proxy.config.json`.

src/service/routes/publicApi.ts
```diff
@@ -1,4 +1,4 @@
-export const toPublicUser = (user) => {
+export const toPublicUser = (user: any) => {
```
kriswest: Can you tighten the user object type like so?
```suggestion
export const toPublicUser = (user: Record<string, boolean | string | null>) => {
```

Or perhaps use the DB's User type?

jescalada: Yes, I used the DB type and I extended it to accept optional fields for `displayName` and `title` which might be AD related?

jescalada:
Hi @06kellyjac, just got the merge conflicts fixed so this is ready for a final look! 😃

src/config/types.ts
```diff
@@ -8,7 +8,7 @@ export interface UserSettings {
   apiAuthentication: Authentication[];
   tempPassword?: TempPasswordConfig;
   proxyUrl: string;
-  api: Record<string, any>;
+  api: ThirdPartyApiConfig;
```
kriswest: I think this whole type is defunct and many of those it references  - replaced by the generated config types

src/config/types.ts
```diff
@@ -49,6 +49,39 @@ export interface Authentication {
...
+  jwtConfig?: JwtConfig;
+
+  // Deprecated fields for backwards compatibility
+  // TODO: remove in future release and keep the ones in adConfig
+  userGroup?: string;
+  adminGroup?: string;
+  domain?: string;
+}
+
+export interface OidcConfig {
```
kriswest: I think these new types should not be created, but rather we should use the types generated from the config schema

src/service/index.ts
kriswest: Not applying the suggestion @jescalada?

src/service/passport/jwtAuthHandler.ts
```diff
@@ -0,0 +1,73 @@
...
+      return next();
+    }
+
+    const token = req.header('Authorization');
+    if (!token) {
+      res.status(401).send('No token provided\n');
+      return;
+    }
+
+    const config = jwtAuthMethod!.jwtConfig!;
```
kriswest: There is no earlier assertion that `jwtAuthMethod.jwtConfig` exists - it should be checked and a specific error returned if jwtAuthMethod exists but the config is invalid

src/service/passport/ldaphelper.ts
```diff
@@ -0,0 +1,65 @@
+import axios from 'axios';
+import type { Request } from 'express';
+import ActiveDirectory from 'activedirectory2';
+import { getAPIs } from '../../config';
+
```
kriswest:
```suggestion
```

src/service/passport/ldaphelper.ts
```diff
@@ -0,0 +1,65 @@
...
+    responseType: 'json',
+    headers: {
+      'content-type': 'application/json',
+    },
+  });
+
+  console.log(`checking if user is in group ${url}`);
+  return client
+    .get(url)
+    .then((res) => res.data)
```
kriswest: Although not introduced in this PR, I think we could be checking the output here and explicitly returning a boolean, rather than `res.data` which is not necessarily boolean and not governed by this repo (as its a third party api call)

src/service/routes/auth.ts
```diff
@@ -171,27 +188,32 @@ router.post('/gitAccount', async (req, res) => {
...
     });
+    return;
   }
 
   try {
     const { username, password, email, gitAccount, admin: isAdmin = false } = req.body;
 
     if (!username || !password || !email || !gitAccount) {
-      return res.status(400).send({
+      res.status(400).send({
```
kriswest, line 216: losing the return here means that you'll proceed on to try and create the user in the DB when you should not

src/service/routes/push.ts
```diff
@@ -0,0 +1,208 @@
...
+      });
+      return;
+    }
+
+    const attestation = {
+      questions,
+      timestamp: new Date(),
+      reviewer: {
+        username,
+        gitAccount: reviewerGitAccount,
```
kriswest: we should stop including this as we don't use it as a key anymore

src/service/urls.ts
```diff
@@ -0,0 +1,22 @@
+import { Request } from 'express';
+
+import { serverConfig } from '../config/env';
+import * as config from '../config';
+
+const { GIT_PROXY_SERVER_PORT: PROXY_HTTP_PORT, GIT_PROXY_UI_PORT: UI_PORT } = serverConfig;
+
+export const getProxyURL = (req: Request): string => {
+  const defaultURL = `${req.protocol}://${req.headers.host}`.replace(
```
kriswest: very minor nit: could prepare the url only if using it. Feel free to ignore

src/config/types.ts
kriswest: (that may have arrived after this branch was started off of course)

In theory we generate a complete type for the config, then it should flow down through anything touching it so we don't need any additional mand maintained types.

kriswest: (that may have arrived after this branch was started off of course)

In theory we generate a complete type for the config, then it should flow down through anything touching it so we don't need any additional hand maintained types.

src/config/types.ts
jescalada: You're right, I just verified that the whole file can be deleted - just one of them was actually being used 👍🏼

src/service/passport/ldaphelper.ts
jescalada: Do you think casting `res.data` as a `Boolean` is enough? Otherwise an explicit `res.data === true` could work for strictness.

src/service/routes/push.ts
jescalada: Since the `reviewerGitAccount` isn't being used anywhere else, I decided to remove the check we did there for the reviewer's git account being present.

I wonder if this ought to be replaced with a check for the reviewer's email instead?

kriswest: Putting the reviewer email into the audit log is not a bad idea at all.

kriswest: @jescalada Perhaps check the user is retrievable from the DB, which I don't think we are doing. If they have been deleted, they might still persist in a repo's `users.canAuthorise` field (I don't think we do anything about that) and hence might still be able to authorise when they shouldn't be able to. If thats the check, you can then just include the author's email address from the DB record in the attestation record.

jescalada: Done! I reused the old logic and replaced it to use `reviewerEmail` instead. The `canUserApproveRejectPush` checks the `canAuthorise` field under the hood, but it returns a boolean rather than the user per se.

src/service/index.ts
kriswest: somehow only just saw this comment. ``async function start(proxy: Proxy)` is how it should stay as it needs the reference to the proxy to update it when there are new domains to proxy

PR #1167, opened 2025-08-27
fix: mongoDB client impl issues

kriswest:
@andypols noted a couple of implementation errors in the mongoDB client (not awaiting promises, searching for repos against the wrong field), which this PR corrects. 

The error is not picked up by tests as our testing of the mongo client is inadequate. I'll open a separate issue to resolve that. In the meantime, this PR will need to be tested manually.


kriswest:
@andypols thanks, fixed. Had a brief poke at why TypeScript isn't helping catch these, a tweak to the `toClass` helper function has improved, but not resolved that situation. It probably needs removing and replacing with use of constructors in the individual classes.

jescalada (approved):
Tested most of the flows with MongoDB Atlas:
- Creating new repo
- Fetching all repos
- Logging in through OIDC (user creation)
- Adding user to push/approval list
- Pushing to repo
- Fetching all pushes
- Approving a push

Everything seems to be working fine! 🎉 

I did find a minor issue on the pushes table:

```ts
<TableCell>
  {isGitHub && (
    <a
      href={`https://github.com/${c.committer}`}
      rel='noreferrer'
      target='_blank'
    >
      {c.committer}
    </a>
  )}
  {!isGitHub && <span>{c.committer}</span>}
</TableCell>
```

Image

The `committer` is `Juan Escalada` with a space in the middle which makes the `a` tag rendering fail. This _might_ be due an issue with the OIDC implementation, but I think that `commitData.committer` is set to be the Git `user.name` rather than the GitHub username.

Any thoughts on this?

kriswest:
There are two places with this issue I think, ui/views/OpenpushRequests/components/PushesTable.tsx and ui/views/PushDetails/PushDetails.tsx. These should be using the same util function to render, but are not. Both are assuming the committer is set the the username at that platform - although only one of them supports that for Gitlab. In neither case is it a safe assumption. There is no attempt to resolve to the username through the gitAccount fields from our users tables (which can only support one user name and doesn't link it to a platform...). 

I think changing this up to render the name and a link for the email address is the easiest and safest thing to do - the email address is really the only thing we can rely on. We can then work from there to reintroduce a way of resolving the user to either a github or gitlab user profile, with handling for failures to resolve. The resolution could either be through an API lookup of some sort (if a reliable one can be found) OR resolution through out own user table, after updating it to support usernames for multiple SCM providers.

I'll raise two issues covering that and will have a quick crack to the first (to fix the immediate rendering issues and give us email links). Let em know if you'd prefer we went another way.

PR #1179, opened 2025-08-29
fix: render committer and author email links instead of estimated profile links

### LINKED ISSUE #1175: UI Push tables assume the git user.name is set to the github username
**Describe the bug**
The UI components that render details of pushes in the admin UI (/src/ui/views/OpenpushRequests/components/PushesTable.tsx and /src/ui/views/PushDetails/PushDetails.tsx) both assume that the `committer` and `author` fields (which come from the git client's `user.name` config variable) will be GitHub user ids. However, they are generally set to a human-readable name for the user (see: https://docs.github.com/en/get-started/git-basics/setting-your-username-in-git#setting-your-git-username-for-every-repository-on-your-computer).

This causes rendering issues and broken links (to github user profiles) in those pages, e.g.:

Image

Further, these pages should be using the same util function to render the links, but are not. Both are assuming the committer is set the the username at that platform - although only one of them supports that for Gitlab. In neither case is it a safe assumption. There is no attempt to resolve to the username through the gitAccount fields from our users tables (which can only support one user name and doesn't link it to a platform...).

I think changing this up to render the name and a link for the email address is the easiest and safest thing to do - the email address is really the only thing we can rely on. 

A later issue reintroduce a way of resolving the user to either a github or gitlab user profile, with handling for failures to resolve. The resolution could either be through an API lookup of some sort (if a reliable one can be found) OR resolution through out own user table, after updating it to support usernames for multiple SCM providers.

**To Reproduce**
1. Set your git `user.name`config: `git config --global user.name "Mona Lisa"`
2. Push a commit into review
3. Observe how it is rendered in the OpenPushRequests and PushDetails views, including the profile links generated.

**Expected behavior**
The user.name and email address are rendered, without a link to the user profile (for the time being).

kriswest:
resolves #1175 

Replaces estimated links to git profile URLs on the OpenPushRequests and PushDetails views with mailto: links based on the email addresses used in commits (simple and will be supportable for all SCM providers/bare git servers).

OpenPushRequests was previously only rendering a link for the first author in a push. It will now extract a list of unique authors appearing in the push and render a link for each (one per line, in order of appearance). 

image

image

### DIFF
--- a/src/types/models.ts
+++ b/src/types/models.ts
@@ -15,6 +15,7 @@ export interface CommitData {
   commitTs?: number;
   message: string;
   committer: string;
+  committerEmail: string;
   tree?: string;
   parent?: string;
   author: string;
--- a/src/ui/utils.tsx
+++ b/src/ui/utils.tsx
@@ -1,8 +1,10 @@
 import axios from 'axios';
+import React from 'react';
 import {
   SCMRepositoryMetadata,
   GitHubRepositoryMetadata,
   GitLabRepositoryMetadata,
+  CommitData,
 } from '../types/models';
 import moment from 'moment';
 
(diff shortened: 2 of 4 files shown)


jescalada (approved):
LGTM! Tested with the previous pushes and is displaying correctly now 🚀

Issue #1174, opened 2025-08-29
Clean up `any` and `as` in TS files


jescalada:
**Is your feature request related to a problem? Please describe.**
After our TS refactors, a few places have `any` for wildcard types and `as` for casting, which we want to remove for a truly clean TS refactor.

One of the quintessential examples of this is on try-catch blocks (source: `checkHiddenCommits.ts`):

```ts
try {
  //...
} catch (e: any) {
  step.setError(e.message);
  throw e;
} finally {
  action.addStep(step);
}
```

We can replace the `catch` bit with the following:

```ts
catch (e: unknown) {
  if (e instanceof Error) {
    step.setError(e.message);
  }
  throw e;
}
```

**Describe the solution you'd like**
Replace all (or as many as possible) usages of `any` and `as` with properly-typed equivalents.

**Additional context**
Requested by @06kellyjac in https://github.com/finos/git-proxy/pull/1166#discussion_r2307003698.

PR #1180, opened 2025-08-30
refactor: `git-proxy-cli` package to TS + ESM (v2)

### LINKED ISSUE #1064: Refactor CLI script to TS + ESM
The `cli` is one of the last few pieces that still needs to be converted to TS + ESM.

Convert `/packages/git-proxy-cli` to TS + ESM. This could include the test file since it's not particularly long.

None

**Additional context**
None

jescalada:
Fixes #1064. Replaces #1065 due to merge conflicts.

I modified some things such as the CLI execution script because the original one (`npx -- @finos/git-proxy-cli`), oddly, no longer works.

I'd love to know better ways to handle the build/testing step and if the CLI flow works as it used to!

### DIFF
--- a/proxy.config.json
+++ b/proxy.config.json
@@ -67,11 +67,7 @@
       }
     }
   ],
-  "api": {
-    "github": {
-      "baseUrl": "https://api.github.com"
-    }
-  },
+  "api": {},
   "commitConfig": {
     "author": {
       "email": {
--- a/src/proxy/processors/push-action/gitleaks.ts
+++ b/src/proxy/processors/push-action/gitleaks.ts
@@ -171,7 +171,7 @@ const exec = async (req: any, action: Action): Promise<Action> => {
         step.setError('\n' + gitleaks.stdout + gitleaks.stderr);
       }
     } else {
-      console.log('succeded');
+      console.log('succeeded');
       console.log(gitleaks.stderr);
     }
   } catch (e) {
--- a/test/fixtures/test-package/package-lock.json
+++ b/test/fixtures/test-package/package-lock.json
@@ -1,0 +1,129 @@
+{
+  "name": "test-package",
+  "version": "0.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "test-package",
+      "version": "0.0.0",
+      "dependencies": {
+        "@finos/git-proxy": "file:../../.."
+      }
+    },
+    "../../..": {
+      "name": "@finos/git-proxy",
+      "version": "2.0.0-rc.3",
+      "license": "Apache-2.0",
+      "workspaces": [
+        "./packages/git-proxy-cli"
+      ],
+      "dependencies": {
+        "@material-ui/core": "^4.12.4",
+        "@material-ui/icons": "4.11.3",
+        "@primer/octicons-react": "^19.18.0",
+        "@seald-io/nedb": "^4.1.2",
+        "axios": "^1.12.2",
+        "bcryptjs": "^3.0.2",
+        "clsx": "^2.1.1",
+        "concurrently": "^9.2.1",
+        "connect-mongo": "^5.1.0",
+        "cors": "^2.8.5",
+        "diff2html": "^3.4.52",
+        "env-paths": "^3.0.0",
+        "express": "^4.21.2",
+        "express-http-proxy": "^2.1.2",
+        "express-rate-limit": "^8.1.0",
+        "express-session": "^1.18.2",
+        "history": "5.3.0",
+        "isomorphic-git": "^1.33.1",
+        "jsonwebtoken": "^9.0.2",
+        "jwk-to-pem": "^2.0.7",
+        "load-plugin": "^6.0.3",
+        "lodash": "^4.17.21",
+        "lusca": "^1.7.0",
+        "moment": "^2.30.1",
+        "mongodb": "^5.9.2",
+        "nodemailer": "^6.10.1",
+        "openid-client": "^6.8.0",
+        "parse-diff": "^0.11.1",
+        "passport": "^0.7.0",
+        "passport-activedirectory": "^1.4.0",
+        "passport-local": "^1.0.0",
+        "perfect-scrollbar": "^1.5.6",
+        "prop-types": "15.8.1",
+        "react": "^16.14.0",
+        "react-dom": "^16.14.0",
+        "react-html-parser": "^2.0.2",
+        "react-router-dom": "6.30.1",
+        "simple-git": "^3.28.0",
+        "uuid": "^11.1.0",
+        "validator": "^13.15.15",
+        "yargs": "^17.7.2"
+      },
+      "bin": {
+        "git-proxy": "index.js",
+        "git-proxy-all": "concurrently 'npm run server' 'npm run client'"
+      },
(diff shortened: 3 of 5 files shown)


packages/git-proxy-cli/package.json
```diff
@@ -2,7 +2,9 @@
   "name": "@finos/git-proxy-cli",
```
fabiovincenzi, line 2: The TypeScript migration looks good! To fully enable ES modules, a couple small config changes are needed:
Add to package.json:
```
{
  "type": "module"
}
```
Update tsconfig.json:
```
json{
  "module": "ES2022",
  "moduleResolution": "bundler",
  "target": "ES2020"
}
```
Without these changes, TypeScript compiles ES module syntax back to CommonJS.

packages/git-proxy-cli/index.ts
```diff
@@ -44,7 +46,7 @@ async function login(username, password) {
     const user = `"${response.data.username}" <${response.data.email}>`;
     const isAdmin = response.data.admin ? ' (admin)' : '';
     console.log(`Login ${user}${isAdmin}: OK`);
-  } catch (error) {
+  } catch (error: any) {
     if (error.response) {
```
fabiovincenzi, line 51: this could be something like this:
` 
catch (error: unknown) {
if (error instanceof Error && 'response' in error && error.response) 
  `
  
  but we can handle all the many cases in a different PR

packages/git-proxy-cli/package.json
jescalada: I've tried setting these, and I'm having several interop issues with the imports from the core GitProxy `/src`, when running tests. `ts-mocha` no longer seems to recognize some of the files.

Tried a few things without luck:
- Adding tests to build for execution (since ts-mocha expects JS files for some reason)
- Replacing `ts-mocha` with `tsx`, which does run but still has issues with not recognizing imports
- Tweaking `moduleResolution`

I think setting `"type": "module"` means that the parent package must also be an ESM module, which won't be possible until we do a complete ESM refactor including the tests. However, I don't know if I'm just missing something and it's possible to set it up right now.

I'm also wondering if the benefits of a "full" ESM conversion are actually worth the trouble. Having TS + ESM syntax already makes things a lot cleaner, while keeping the CommonJS interop spares us headaches when testing or using the plugin module.

Pinging @06kellyjac for thoughts about this 😃

packages/git-proxy-cli/index.ts
jescalada: I made #1174 to deal with this one, would be nice to have that in a separate PR!

packages/git-proxy-cli/types.ts
```diff
@@ -0,0 +1,35 @@
+export interface PushFilters {
+  allowPush?: boolean;
+  authorised?: boolean;
+  blocked?: boolean;
+  canceled?: boolean;
+  error?: boolean;
+  rejected?: boolean;
+}
+
+export interface PushData {
```
kriswest: Why not import the PushData type from the service (/src/types/models.ts/PushData) instead of duplicating it (although there are some minor differences). 

Using the same type across the CLI and the API that it uses should help us with future maintenance (change it in the API and the build could identify necessary changes in the CLI ).  The more separate copies of types we have the less useful they are. If we use the same one across multiple projects, changes in one can highlight compatibility issues we are creating, or can reveal properties we could use over there, but might not notice if using a type that just defined what we use, rather than what's available, which is I think what's happening here (CLI uses, API makes available).

packages/git-proxy-cli/types.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export interface PushStep {
+  stepName: string;
+  error: boolean;
+  errorMessage: string;
+  blocked: boolean;
+  blockedMessage: string;
+}
+
+export interface CommitData {
```
kriswest: /src/types/models.ts/CommitData ?

packages/git-proxy-cli/types.ts
```diff
@@ -0,0 +1,35 @@
+export interface PushFilters {
```
kriswest: @jescalada you are introducing a type for this in #1166 (PushQuery) at https://github.com/finos/git-proxy/pull/1166/files#diff-696f37e2ac7af62c241cb8e81c4a2773957fd38f07cd6241709126382adec0e2R9

Can we use the same one?

packages/git-proxy-cli/types.ts
jescalada: I did some refactoring with here for the `Step` class which was causing issues (extracted a `StepData` interface only with the data fields). We can handle the other issues in #1193!

packages/git-proxy-cli/package.json
dcoric: I would proceed with the current approach as we get immediate benefits: 
- TypeScript + modern syntax, 
- no breaking changes: CLI works identically, 

I got the chance to play a bit with CLI, and to end users, it won't make any difference. We would benefit of this "module" change with additional performance + tooling benefits, but I think it is safe to proceed as it is and migrate to true ES modules later.

jescalada:
@kriswest This should be ready for a final check - looking to merge this ASAP as it's blocking some of our team's PRs 🙏🏼

packages/git-proxy-cli/test/testCli.test.ts
```diff
@@ -129,25 +130,25 @@ describe('test git-proxy-cli', function () {
...
       const username = 'admin';
       const password = 'admin';
-      const cli = `npx -- @finos/git-proxy-cli login --username ${username} --password ${password}`;
+      const cli = `${CLI_PATH} login --username ${username} --password ${password}`;
       const expectedExitCode = 0;
       const expectedMessages = [`Login "${username}" <admin@place.com> (admin): OK`];
       const expectedErrorMessages = null;
       try {
-        await helper.startServer(service);
+        await helper.startServer(service as unknown as Proxy);
```
kriswest: This isn't right surely - see comment on testCliUtils.ts, I think the error is there.

packages/git-proxy-cli/test/testCliUtils.ts
```diff
@@ -81,11 +85,11 @@ async function runCli(
 
 /**
  * Starts the server.
- * @param {Object} service - The GitProxy API service to be started.
+ * @param {Proxy} service - The GitProxy API service to be started.
  * @return {Promise<void>} A promise that resolves when the service has
  * successfully started. Does not return any value upon resolution.
  */
-async function startServer(service) {
+async function startServer(service: Proxy) {
```
kriswest: I think this type is wrong which is why you keep having to cast it `as unknown as Proxy` in testCliUtils.ts

kriswest: There isn't a type for the service I imagine until #1166 merges. I'll leave it upto you on how to resolve. Not against combining them or getting the other one merged first then update this one to use the type.

jescalada: My temporary solution for this was to replace the type with this to remove all the `unknown`s:

```ts
startServer(service: typeof import('../../../src/service/index'))
```

I'll fix the typing properly in the `service` PR after this one gets merged!

kriswest: okey dokey - merging in that case

PR #1182, opened 2025-09-01
fix: allowing empty diffs in scanDiff

### LINKED ISSUE #1008: (bug) scanDiff fails on force pushes

fabiovincenzi:
Fixes #1008 - scanDiff now properly handles force pushes after rebases by allowing legitimate empty diffs instead of blocking them.


test/integration/forcePush.integration.test.js
```diff
@@ -0,0 +1,134 @@
...
+      const afterGetDiff = await getDiff({}, action);
+      expect(afterGetDiff.steps).to.have.length.greaterThan(0);
+
+      const diffStep = afterGetDiff.steps.find((s) => s.stepName === 'diff');
+      expect(diffStep).to.exist;
+
+      // May produce empty diff or error - both should be handled
+      if (diffStep.error) {
+        console.log('getDiff failed as expected with unreachable commitFrom');
+        expect(diffStep.errorMessage).to.be.a('string');
```
kriswest: Is there a particular error message component that should be checked?

test/integration/forcePush.integration.test.js
```diff
@@ -0,0 +1,134 @@
...
+      ];
+
+      const afterGetDiff = await getDiff({}, action);
+      expect(afterGetDiff.steps).to.have.length.greaterThan(0);
+
+      const diffStep = afterGetDiff.steps.find((s) => s.stepName === 'diff');
+      expect(diffStep).to.exist;
+
+      // May produce empty diff or error - both should be handled
+      if (diffStep.error) {
```
kriswest: This concerns me - do the test steps genuinely produce either output?

test/scanDiff.emptyDiff.test.js
```diff
@@ -0,0 +1,91 @@
+const { Action } = require('../src/proxy/actions');
```
kriswest: This file be in the /test/processors folder

test/scanDiff.test.js
```diff
@@ -249,7 +249,7 @@ describe('Scan commit diff...', async () => {
       expect(errorMessage).to.contains('Your push has been blocked');
     }
   });
-  it('When no diff is present, the proxy is blocked...', async () => {
+  it('When no diff is present, the proxy allows the push (legitimate empty diff)...', async () => {
```
kriswest, line 252: This file should be moved into the /test/processors folder

test/integration/forcePush.integration.test.js
fabiovincenzi: @kriswest  You're right - I've split this into two tests:
  1. Valid SHA → always succeeds with real diff content
  2. Invalid SHA → always fails with "fatal: Invalid revision range"

kriswest (approved):
LGTM

PR #1187, opened 2025-09-04
fix: reimplement push parsing to prevent z data errors

### LINKED ISSUE #1040: Z_DATA_ERROR during push parsing
The following stack trace is observed when GitProxy is receiving a push. It's not clear what the cause of it is but it seems related to the parsing logic in parsePush step. The same error is hit multiple times in the process of a single push.

```
Error: incorrect header check
    at genericNodeError (node:internal/errors:984:15)
    at wrappedFn (node:internal/errors:538:14)
    at Zlib.zlibOnError [as onerror] (node:zlib:189:17)
    at Zlib.callbackTrampoline (node:internal/async_hooks:130:17)
    at processChunkSync (node:zlib:415:12)
    at zlibBufferSync (node:zlib:178:12)
    at Object.syncBufferWrapper [as inflateSync] (node:zlib:767:14)
    at unpack (/app/src/proxy/processors/push-action/parsePush.js:234:25)
    at getContent (/app/src/proxy/processors/push-action/parsePush.js:212:35)
    at getContents (/app/src/proxy/processors/push-action/parsePush.js:145:37) {
  errno: -3,
  code: 'Z_DATA_ERROR'
}
```

It seems to be mostly cosmetic as the remaining steps still proceed as expected. Parsing of the pushed data is only used for extracting certain metadata out of the payload for the purposes of checking out the remote repo, performing diffs against HEAD and base branches and other operations. 

The receive pack from git usually contains "delta-fied"[^1] chunks of data and I suspect it is not actually zlib compressed hence this error message. We can likely rely solely on the first part of the payload which contains the following info:

- tree + commit sha
- parent + commit sha
- author & committer names & email addresses
- commit message
- GPG signature

This is the expected payload which is successfully parsed even with these errors raised:

```
{
  x: {
    item: 0,
    value: 159,
    type: 1,
    size: 543,
    deflatedSize: 397,
    objectRef: null,
    content: 'tree {commit sha redacted}\n' +
      'parent {commit sha redacted}\n' +
      'author {redacted} 1747147000 -0400\n' +
      'committer Thomas Cooper <{redacted}> 1748440293 -0400\n' +
      'gpgsig -----BEGIN PGP SIGNATURE-----\n' +
      ' \n' +
      ' {redacted}\n' +
      ' {redacted}\n' +
      ' {redacted}\n' +
      ' {redacted}\n' +
      ' -----END PGP SIGNATURE-----\n' +
      '\n' +
      "{commit message}\n" +
      '\n' +
      'Signed-off-by: {commit sha redacted}\n'
  }
}
```

[^1]: https://github.com/git/git/blob/master/Documentation/gitprotocol-capabilities.adoc#ofs-delta

kriswest:
resolves #1040

Reimplements parsing of PACK file contents to resolve a number of issues that were causing a loss of sync with the headers and compressed data streams that make up the pack file, resulting in Z_DATA_ERROR being thrown and the rest of the pack file failing to decode. This results in partial content being extracted from push pack files and commits being missed/not processed by later steps. The missing commits can also interact with the checkHiddenCommits task to block a subset of pushes, while others proceed with missing data.

After this PR, Git Proxy will correctly process a handful of pushes I have queued up that either parse with missing data or are blocked by checkHiddenCommits.

This PR:
- Corrects issues with the reading of the PACK file metadata (correct number of entries)
- Corrects parsing of the git object headers, including the variable length size field and variable length offset value for ofs_daelta (type 6) object headers
- Reimplements the decompression such that the input is fed into the inflator a byte at a time, until an end event fires. This appears to happen one byte after the end of the compressed data (and any trailing bytes in its stream). This replaces the current approach of inflating the object and then deflating it again (possibly with different settings) to estimate the length of the compressed data and better replicates how git handles the operation.
- Implements two tests for decompressing multi-object pack files that confirm correct handling of the variable length headers and implements a number of utility functions that accurately mock git pack file encoding.
- Removes the previous implementation and tests for it - the updated tests don't mock zlib anymore, instead producing realistic mock packfile data encoded and decoded with zlib.


src/proxy/processors/push-action/parsePush.ts
```diff
@@ -263,146 +261,267 @@ const getCommitData = (contents: CommitContent[]): CommitData[] => {
...
-    more = nextM.getBit(3);
+const parseOfsDeltaOffset = (
+  buffer: Buffer,
+  offset: number,
+): { baseOffset: number; length: number } => {
+  let i = 0;
+  let byte = buffer[offset];
+  let value = byte & 0x7f;
+
+  while (byte & 0x80) {
```
jescalada: Wonder if it makes sense to make constants for these `0x7f`, `0x80` for readability?

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -263,146 +261,267 @@ const getCommitData = (contents: CommitContent[]): CommitData[] => {
...
+    while (offset < buffer.length && !(done || error)) {
+      try {
+        await new Promise<void>((resolve, reject) => {
+          if (!done) {
+            const byte = buffer.subarray(offset, offset + 1);
+            offset++;
+            // store the resolve function in case an error occurs as callback will never be called
+            currentWriteResolve = resolve;
+            // use the callback to throttle input such that each byte is processed before we insert the next
+            inflater.write(byte, () => {
```
jescalada: I'm not too sure if this affects push parse times, but perhaps there's a way to improve the efficiency of the writing here?

jescalada (commented):
I'm not a low-level code expert, so I'll just leave some general suggestions!

I've experimented a bit to verify whether the current code slows down the push process somewhat, and it seems that it's slightly slower (10~20%) for very large pushes:

### Basic timing method
Image

## Timing scripts
### Regular commit
#### `main`
Image

#### This PR
Image

### Massive commit (~12 MB of text changes)
#### `main`
Image

#### This PR
Image

This might not be something we want to optimize right now, though.

src/proxy/processors/push-action/parsePush.ts
kriswest: Probably quite readable to a C programmer - but fair enough for us humans - I'll pop a couple in for those we reuse..

src/proxy/processors/push-action/parsePush.ts
kriswest: I'm not surprised at a slight slow-down - although unless you time the decompression step directly you may see a slow down due to more of the data being extracted! Having to process the data a byte at a time, and allocate and resolve a promise at each byte, feels very inefficient - but its necessary to catch the end of each object. Its no doubt far more efficiently implemented in C, but other than cracking open zlib I don't see a better way of doing this.

We could possibly switch to [fast-zlib](https://github.com/timotejroiko/fast-zlib) or zlib-sync for a speedup - but I don't immediately see how we find the end of the stream with those libs and it'll take further experimentation to figure that out.

kriswest:
I think there's not a lot more I can do on this PR, without heading into a port to fast-zlib or zlib-sync, which I think is a separate challenge - we probably need to do something similar to fast-zlib (using the internals of zlib, such as `_processChunk` directly) but will need to figure out how to detect that we are done. I think that's best left to subsequent work as the push parsing time is currently dominated by the repo checkout - decompression is just a small part last milliseconds.

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -263,146 +269,266 @@ const getCommitData = (contents: CommitContent[]): CommitData[] => {
...
+  baseSha?: Buffer;
+}
 
-  return parseInt(strBits, 2);
+/**
+ * Maps Git object type codes to human-readable names.
+ * @param {number} typeCode  Numeric type code from PACK file.
+ * @return {string} Git object type as string.
+ */
+const gitObjectType = (typeCode: number): string => {
```
coopernetes: Consider making the return a sum type.

```suggestion
type GitObjectType =
  | 'commit'
  | 'tree'
  | 'blob'
  | 'tag'
  | 'ofs_delta'
  | 'ref_delta'
  | unknown;

/**
 * Maps Git object type codes to human-readable names.
 * @param {number} typeCode  Numeric type code from PACK file.
 * @return {GitObjectType} Git object type
 */
const gitObjectType = (typeCode: number): GitObjectType => {
```

coopernetes (approved):
One very minor nit but otherwise LGTM. I think @kriswest knows the packet format best at this point now that this refactoring has been done lol. 

My only suggestion would be to add some real (binary) receive-pack files to use as input for the tests.

```shell
    # Create some new content to generate a pack
    echo "New content for pack testing $(date)" > new-file.txt
    echo "# Updated README $(date)" >> README.md
    git add .
    git commit -m "Test commit for pack capture $(date)"
    
    echo "Capturing receive-pack output to $output_file..."
    
    # Method 1: Capture the actual pack data being sent
    # This captures the raw pack file that would be sent to receive-pack
    git format-patch --stdout HEAD~1..HEAD > "$output_file.patch"
    
    # Method 2: Capture pack file from push operation with verbose output
    GIT_TRACE_PACKET=1 git push origin main 2> "$output_file.trace" || true
    
    # Method 3: Create a pack file directly
    git pack-objects "$output_file" < <(git rev-list --objects HEAD)
```

PR #1189, opened 2025-09-07
feat: push speed optimizations

### LINKED ISSUE #985: optimize pullRemote for large repos
On particularly large repositories as well as when GitProxy is deployed in a production-like environment (constrained or at least defined upper limits on CPU/memory), `pullRemote` step will sometimes cause the `git` client to time out and return a server error if it takes too long to complete.

We should consider whether we need to pull down the whole history every time there is a new push. Using `depth: 1, singleBranch: true` on the git clone call seems to reduce the amount of time (in our particular instance) that the push takes and allows it to proceed through proxy processing normally.

- Tweak the isomorphic-git call during cloning to the minimum required to produce a working diff. 
- Pull down selective history such as default branches only?
- Make it configurable?

Increasing the timeout for the server (we have it deployed behind a load balancer) helps marginally but the full clone can be very expensive with large repos.

**Additional context**

This is likely a problem that affects outdated branches more than anything. When a commit drifts (is left stale for days, weeks or months), the combination of the long clone time + large diff usually results in server errors.

jescalada:
Fixes #985. The profiling screenshots are in the issue discussion.

This PR reduces _push_ times by around 50% by using the `depth: 1` and `singleBranch: true` options **in `pullRemote`** when cloning. It also adds the `stream: true` option to the `express-http-proxy` configuration, which should help with larger pushes.

I've experimented with other optimizations to the chain actions but it seems that the majority of time is spent cloning the target repository (and therefore limited by download speeds).

I'd say storing the target repository so we only clone the entire repo once is the best option, the tradeoff being that we might need to store several GB of data for larger repos. Perhaps we could have a config option to toggle this so admins can do as they see fit?


src/proxy/processors/push-action/pullRemote.ts
```diff
@@ -33,15 +32,12 @@ const exec = async (req: any, action: Action): Promise<Action> => {
...
       http: gitHttpClient,
       url: action.url,
-      onAuth: () => ({
-        username,
-        password,
-      }),
       dir: `${action.proxyGitPath}/${action.repoName}`,
+      onAuth: () => ({ username, password }),
+      singleBranch: true,
+      depth: 1,
```
coopernetes, line 38: Note that this effectively means that only the default branch is supported as a diff target. It can be deferred as this speeds up the checkout pretty significantly. However, this does nothing to address overall scalability with giant repos being proxied, pull time and ensuring that the proper diffs are captured by the proxy for presentation in the UI.

As a follow-up to these optimizations, it would be useful to capture the list of tips from a remote repository, persist it in the database and allow for diff's to be "selectable" to a given target branch for proper review. Otherwise, this effectively locks in Git Proxy to supporting one particular branching strategy. Probably fine for most projects but just thought I would raise it here. It can be tracked in an enhancement issue.

Were the proxy to clone all branches, then you can make `singleBranch: false` while still performing a shallow clone with `depth: 1`.

jescalada: Understood! I added #1207 to track this.

kriswest: Hmm you make a good point @coopernetes. I note that when we `getDiff` we use a revision range (`[commitFrom, commitTo]`. We'll need to check that, if the `commitFrom` doesn't exist on the default branch, that we can still get the diff (i.e. does the simpleGit retrieve the necessary data if it doesn't have it - that would be most convenient). If not, we may need to play around a bit... for example cloning the branch that that the commits are being pushed to (if it exists already) vs. the default branch if it doesn't?

@jescalada a quick smoke test might help:
- Create a branch
- Push a couple of commits to it on the remote
- add a couple more commits and push through the proxy
- Check if the diff is correct/no errors are thrown

If theres an issue, then figuring out the base might help... I'll try and find time to poke at what we're doing at the moment and git protocol later on today if I can

jescalada: Related issue: #1291

PR #1197, opened 2025-09-10
chore: add /healthcheck endpoint to the proxy

andypols:
# Summary

We’re deploying GitProxy on AWS behind an [Application Load Balancer (ALB) listener](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-listeners.html) that routes traffic to either the web app or the proxy. ALB target groups require a health-check path. The web app already exposes `/healthcheck`, but the proxy does not.

This PR adds a `/healthcheck` endpoint to the proxy so the load balancer can verify it’s running, and includes a test.

## Changes

- Add `/healthcheck` route to the proxy returning 200 OK.
- Add corresponding test in `test/testProxyRoute.test.js`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -81,7 +81,7 @@
         "@vitejs/plugin-react": "^4.7.0",
         "chai": "^4.5.0",
         "chai-http": "^4.4.0",
-        "cypress": "^14.5.4",
+        "cypress": "^15.2.0",
         "eslint": "^8.57.1",
         "eslint-config-google": "^0.14.0",
         "eslint-config-prettier": "^10.1.8",
@@ -107,6 +107,9 @@
         "typescript": "^5.9.2",
         "vite": "^4.5.14",
         "vite-tsconfig-paths": "^5.1.4"
+      },
+      "engines": {
+        "node": ">=20.19.2"
       },
       "optionalDependencies": {
         "@esbuild/darwin-arm64": "^0.25.9",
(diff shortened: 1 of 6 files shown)


src/proxy/routes/index.ts
```diff
@@ -234,6 +234,14 @@ const getRouter = async () => {
   console.log('proxy keys registered: ', JSON.stringify(proxyKeys));
 
   router.use('/', (req, res, next) => {
+    if (req.path === '/healthcheck') {
+      res.set('Cache-Control', 'no-cache, no-store, must-revalidate, proxy-revalidate');
+      res.set('Pragma', 'no-cache');
+      res.set('Expires', '0');
+      res.set('Surrigate-Control', 'no-store');
```
kriswest: typo?
```suggestion
      res.set('Surrogate-Control', 'no-store');
```

test/testProxyRoute.test.js
```diff
@@ -200,6 +200,30 @@ describe('proxy route helpers', () => {
...
+    app.use('/', await getRouter());
+  });
+
+  it('returns 200 OK with no-cache headers', async () => {
+    const res = await chai.request(app).get('/healthcheck');
+
+    expect(res).to.have.status(200);
+    expect(res.text).to.equal('OK');
+
+    // Basic header checks (values defined in route)
```
kriswest, line 217: perhaps check for the surrogate-control header as you are setting it

kriswest (changes requested):
Seems fine - I think there's a typo to fix.

src/proxy/routes/index.ts
andypols: Doh! Thanks.  Will fix that.

test/testProxyRoute.test.js
andypols: Will extend test.

kriswest (approved):
LGTM

PR #1201, opened 2025-09-12
fix: Logout calls localhost in prod; standardise API base resolution

andypols:
# Summary

This PR fixes a production logout failure caused by the client defaulting to http://localhost:3000 when VITE_API_URI is unset.

## Root Cause

In `src/ui/components/Navbars/DashboardNavbarLinks.tsx` we have:

```
      const { data } = await axios.post(
        `${process.env.VITE_API_URI || 'http://localhost:3000'}/api/auth/logout`,
        {},
        getAxiosConfig(),
      );
```
Because `VITE_API_URI` isn’t set in prod, the code hit the `localhost` fallback.

Different areas of the code use different patterns, leading to inconsistent behaviour:

- `src/ui/views/Login/Login.tsx`:
```
const loginUrl = `${process.env.VITE_API_URI}/api/auth/login`;
```

- `src/ui/services/user.ts`:
```
const baseUrl = process.env.VITE_API_URI || location.origin;
```

Note: in Vite client code we should prefer `import.meta.env.VITE_API_URI` over `process.env.*` but that does not work until we upgrade typescript config to use ES Modules (and not "CommonJS").

## Fix

What this PR changes

- Introduces a single source of truth for the API base.
- Switches all API calls to use `relative URLs by default`, with an **optional env override**.
- Removes hardcoded `localhost` and `location.origin` fallbacks from call sites.

### DIFF
--- a/src/proxy/routes/index.ts
+++ b/src/proxy/routes/index.ts
@@ -6,23 +6,31 @@ import { executeChain } from '../chain';
 import { processUrlPath, validGitRequest, getAllProxiedHosts } from './helper';
 import { ProxyOptions } from 'express-http-proxy';
 
+enum ActionType {
+  // eslint-disable-next-line no-unused-vars
+  ALLOWED = 'Allowed',
+  // eslint-disable-next-line no-unused-vars
+  ERROR = 'Error',
+  // eslint-disable-next-line no-unused-vars
+  BLOCKED = 'Blocked',
+}
+
 const logAction = (
   url: string,
   host: string | null | undefined,
   userAgent: string | null | undefined,
-  errMsg: string | null | undefined,
-  blockMsg?: string | null | undefined,
+  type: ActionType,
+  message?: string,
 ) => {
-  let msg = `Action processed: ${!(errMsg || blockMsg) ? 'Allowed' : 'Blocked'}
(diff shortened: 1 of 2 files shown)


src/ui/apiBase.ts
```diff
@@ -0,0 +1,8 @@
+const strip = (s: string) => s.replace(/\/+$/, '');
```
JamieSlome: Can we write a test? @andypols

Feels like the sort of functionality that would be nice to prevent future regression on ❤️ Nice work!

andypols:
Please can a @finos/git-proxy-maintainers review this for me - thanks

jescalada (approved):
LGTM! Just wondering if we can use the `API_BASE` instead of envs in `git-proxy-cli/index.js`:

```js
const { GIT_PROXY_UI_HOST: uiHost = 'http://localhost', GIT_PROXY_UI_PORT: uiPort = 8080 } =
  process.env;

const baseUrl = `${uiHost}:${uiPort}`;
```

kriswest (approved):
As we can't (shouldn't?) use `API_BASE` in the JS UI services until converted to TS, I think this is good to go.

PR #1165, opened 2025-08-21
feat: e2e tests using docker compose

coopernetes:
Closes #1302 

This PR adds a new end-to-end test setup using docker compose. It includes a few basic tests such as git clone (fetch) and a single push test with an expected failure. By moving end-to-end testing to a container-based setup, we can start to remove the state coupling that exists between various test suites at the moment. For example, attempting to run a single test can result in unexpected failures because it relies on setups from another test that starts a real GitProxy server, inserts some data into the database, etc. By refactoring those tests into these new e2e style tests, we can remove those dependencies and start to break apart the existing tests to focus more on the system-under-test.

In addition to this setup, a few other changes were made to support it as well as to cleanup some miscellaneous issues. **If preferred, these can be moved out into their own PRs.**.

1. ~Remove some database initialization that was triggered for local files even though Mongo is in use.~ fixed via #1316 
2. Added a setup for the Vite frontend build to be able to fetch its base URL from the server to allow for dynamic URLs to be used to host the app. 
3. Added a Dockerfile for the main application. 

Related issues:

- #1143 
- #1302
- #1303 

See below compose logs:

<summary>
<details>

```
finos-git-proxy-git-proxy-1  | Created runtime configuration with:
finos-git-proxy-git-proxy-1  |   API URL: auto-detect
finos-git-proxy-git-proxy-1  |   Allowed Origins: *
finos-git-proxy-git-proxy-1  |   Environment: test
finos-git-proxy-git-proxy-1  | Configuration sources are disabled
finos-git-proxy-git-proxy-1  | Configuration loaded successfully
finos-git-proxy-git-proxy-1  | Setting config file to: /app/test-e2e.proxy.config.json
finos-git-proxy-git-proxy-1  | validating config
finos-git-proxy-git-proxy-1  | Setting up the proxy and Service
finos-git-proxy-git-proxy-1  | No plugins configured
finos-git-proxy-git-proxy-1  | Found 0 plugin modules
finos-git-proxy-git-proxy-1  | Loading MongoDB database adaptor
finos-git-proxy-git-proxy-1  | creating user
finos-git-proxy-git-proxy-1  |         user=admin,
finos-git-proxy-git-proxy-1  |         gitAccount=none
finos-git-proxy-git-proxy-1  |         email=admin@place.com,
finos-git-proxy-git-proxy-1  |         admin=true
finos-git-proxy-git-proxy-1  |         oidcId=
[...]

Issue #1143, opened 2025-08-05
E2E test improvements


jescalada:
**Is your feature request related to a problem? Please describe.**
We could improve our Cypress tests in various ways, including:
- [ ] Aliasing long component IDs
- [ ] Adding functions for interacting with API to seed database and cleanup
- [ ] Removing dependencies with regular unit tests (#978)

**Describe the solution you'd like**
### Aliasing long component IDs

Some of the latest tests added are comprehensive, but a bit hard to read and brittle in the sense that changes on the components themselves may require lots of changes to fix the tests. An initial solution to this would be aliases on the component IDs, so that we only need to change the Cypress check in one place:

```js
cy.get('[data-testid="repo-list-view"]').as('repoList');
cy.get('@repoList').find('[data-testid="add-repo-button"]');  
```

### Database seed/cleanup functions

We could have `cy.seedDb()`, `cy.cleanUp()` functions to make sure all E2E tests have the required data before running (via API calls). A good idea is to seed with data that we know won't be present normally and then delete it after the E2E tests.

### Removing dependencies with regular unit tests
See #978 for context.

**Describe alternatives you've considered**
None

**Additional context**
Related #978

kriswest:
Closing to rethink once the refreshed UI is in: 
- #1507 

We're expecting to have to re-vamp the cyprus tests as part of #1507 anyway... and teh new E2E tests are better at a full integration test

PR #1202, opened 2025-09-14
refactor: migrate tests to Vitest + TS

### LINKED ISSUE #1150: Migrate tests from Mocha + JS to Vitest + TS
Although not critical, migrating the tests from Mocha (and JS) to ~~Jest~~ **Vitest*** (and TS) would be great to truly complete the JS -> TS refactor, as well as help increase coverage for hard-to-test modules.

jescalada:
Fixes #1150, built on top of #1166. 

Not all tests were perfectly refactored from Chai/Mocha -> Vitest. Some required partial or total rewriting to get them to play nice.

Some tests, specifically the proxy-related tests (`proxy.test.ts`, `testProxy.test.ts` and most importantly `testProxyRoute.test.ts`) have been very problematic for some reason - the tests in `testPush.test.ts` seem to fail in the CI when some of the proxy tests execute prior. They fail on the first execution locally, and then pass without issue (maybe a caching/cleanup problem).

Coverage has dropped somewhat to 80.5%, but we should see an increase after removing unused/untested code such as plugin stuff.

Note that the CI is failing due to the new Blue Oak license.

### DIFF
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -3,7 +3,6 @@ import Layout from '@theme/Layout';
 import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 import Avatar from '../components/avatar';
 import Testimonials from './testimonials';
-import ReactPlayer from 'react-player';
 import axios from 'axios';
 
 /**
@@ -60,14 +59,17 @@ function Home() {
           </header>
           {showDemo ? (
             <section style={{ background: 'black' }}>
-              <ReactPlayer
-                url='https://www.finos.org/hubfs/Projects%20%2B%20SIGs/Open%20Source%20Readiness%20OSR/OSR%20Meeting_%20GitProxy%20Jamie%20Slome%20Citi%20Presentation.mp4#t=496'
-                controls={true}
-                width='100%'
-                pip={true}
-                stopOnUnmount={false}
-                playing={true}
-              />
+              <video
+                controls
+                style={{ width: '100%', display: 'block', aspectRatio: '16/9' }}
+                preload='metadata'
+              >
+                <source
+                  src='https://www.finos.org/hubfs/Projects%20%2B%20SIGs/Open%20Source%20Readiness%20OSR/OSR%20Meeting_%20GitProxy%20Jamie%20Slome%20Citi%20Presentation.mp4#t=496'
+                  type='video/mp4'
+                />
+                Your browser does not support the video tag.
+              </video>
             </section>
           ) : (
             <section className='terminal'>


coopernetes:
@jescalada related to #1165 can we use this PR as an opportunity to remove "functional" tests from our unit test suite? IME it is difficult to run all the tests inside corporate environments behind proxies & firewalls since some tests reach out to github.com repositories. 

If we can simplify unit & integration tests as part of this rewrite, we can start writing true E2E tests using the isolated Git server incorporated into that PR & remove the coupling to public repos being accessible.

src/service/index.ts
```diff
@@ -0,0 +1,114 @@
...
+const { GIT_PROXY_UI_PORT: uiPort } = serverConfig;
+
+const app: Express = express();
+const _httpServer = http.createServer(app);
+
+const corsOptions = {
+  credentials: true,
+  origin: true,
+};
+
```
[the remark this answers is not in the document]
06kellyjac: didn't we have configuration for CORS before? :man_shrugging:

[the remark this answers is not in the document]
jescalada: Yes, these were present before the refactor... 🤷🏼

.github/workflows/dependency-review.yml
```diff
@@ -21,6 +21,6 @@ jobs:
         with:
           comment-summary-in-pr: always
           fail-on-severity: high
-          allow-licenses: MIT, MIT-0, Apache-2.0, BSD-3-Clause, BSD-3-Clause-Clear, ISC, BSD-2-Clause, Unlicense, CC0-1.0, 0BSD, X11, MPL-2.0, MPL-1.0, MPL-1.1, MPL-2.0, OFL-1.1, Zlib
+          allow-licenses: MIT, MIT-0, Apache-2.0, BSD-3-Clause, BSD-3-Clause-Clear, ISC, BSD-2-Clause, Unlicense, CC0-1.0, 0BSD, X11, MPL-2.0, MPL-1.0, MPL-1.1, MPL-2.0, OFL-1.1, Zlib, BlueOak-1.0.0
```
kriswest, line 24: 👍

config.schema.json
```diff
@@ -427,6 +427,10 @@
                 "password": {
                   "type": "string",
                   "description": "Password for the given `username`."
+                },
```
kriswest, line 437: Are you able to run the schema docs generator (gen-schema-doc) to add these properties to the config reference? Give me a shout if not (i.e. if you haven't got the python env setup needed to do so).

You have run the quicktype generator (generate-config-types).

src/config/index.ts
```diff
@@ -203,14 +203,19 @@ export const getAPIs = () => {
   return config.api || {};
 };
 
-export const getCookieSecret = (): string | undefined => {
+export const getCookieSecret = (): string => {
   const config = loadFullConfiguration();
+
+  if (!config.cookieSecret) {
+    throw new Error('cookieSecret is not set!');
```
kriswest, line 211: We should possibly make `cookieSecret` required in the config schema. Currently none of the top-level properties are required... but then that lets you validate a file full of overrides (as long as the defaults cover everything that is actually required). However, its leading to our generated type for the config having these properties as optional.

This may be worth some further thought - we could have two top-level schemas (for the defaults and for override files) that reference the same defs for top-level properties I suppose. That would give us two different typescript types to use, as appropriate.

src/config/index.ts
```diff
@@ -203,14 +203,19 @@ export const getAPIs = () => {
...
+  }
+
   return config.cookieSecret;
 };
 
-export const getSessionMaxAgeHours = (): number | undefined => {
+export const getSessionMaxAgeHours = (): number => {
   const config = loadFullConfiguration();
-  return config.sessionMaxAgeHours;
+  return config.sessionMaxAgeHours || 24;
```
kriswest, line 219: As above - no default needed if it has to be in the defaults and we have a type expressing that.

src/proxy/processors/push-action/checkCommitMessages.ts
```diff
@@ -1,48 +1,53 @@
...
-  console.log(`isMessageAllowed(${commitMessage})`);
+  try {
+    const commitConfig = getCommitConfig();
 
-  // Commit message is empty, i.e. '', null or undefined
-  if (!commitMessage) {
-    console.log('No commit message included...');
-    return false;
-  }
+    console.log(`isMessageAllowed(${commitMessage})`);
```
kriswest: this is (existing) debug output, we should drop it or start doing a better job with logging!

Not a blocker for this PR, but a note to self/reminder

kriswest: There are several other places that info on why the message has failed the check is just being logged. We should refactor this to return an error message or null and then carry that forward into step.log messages, as that'll make illegal messages way less annoying to debug and we can drop all the debug logging here.

src/proxy/processors/push-action/checkCommitMessages.ts
```diff
@@ -1,48 +1,53 @@
...
-  const patternMatches = positivePatterns.flat().filter((result) => !!result);
+    // Flatten any positive pattern results into a 1D array...
+    const patternMatches = positivePatterns.flat().filter((result) => !!result);
 
-  // Commit message matches configured block pattern(s)
-  if (literalMatches.length || patternMatches.length) {
-    console.log('Commit message is blocked via configured literals/patterns...');
+    // Commit message matches configured block pattern(s)
+    if (literalMatches.length || patternMatches.length) {
+      console.log('Commit message is blocked via configured literals/patterns...');
```
kriswest, line 39: Another debug message - although here it'd also be nice to have names for the patterns (not in our config currently), or the pattern itself

src/service/routes/push.ts
```diff
@@ -103,8 +103,16 @@ router.post('/:id/authorise', async (req: Request, res: Response) => {
     const push = await db.getPush(id);
     console.log({ push });
```
kriswest: Debug output again through this file. Still not a blocker for this PR, but obvious we need to improve.

test/1.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  - Reset module cache after a test
+*/
+
+import { describe, it, beforeAll, afterAll, beforeEach, afterEach, expect, vi } from 'vitest';
+import request from 'supertest';
+import service from '../src/service';
+import * as db from '../src/db';
+import Proxy from '../src/proxy';
+
+const TEST_REPO = {
```
kriswest, line 18:
```suggestion
// Create constants for values used in multiple tests
const TEST_REPO = {
```

test/1.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    const authMethods = config.getAuthMethods();
+    expect(authMethods).toHaveLength(3);
+    expect(authMethods[0].type).toBe('local');
+  });
+
+  // Runs after each test
+  afterEach(function () {
+    // Restore all stubs
+    vi.restoreAllMocks();
```
kriswest: what about moving the `vi.resetModules();` on line 54 here so it doesn't have to appear in multiple tests?

test/ConfigLoader.test.ts
```diff
@@ -129,14 +128,14 @@ describe('ConfigLoader', () => {
 
       fs.writeFileSync(tempConfigFile, JSON.stringify(newConfig));
 
-      configLoader = new ConfigLoader(initialConfig);
-      const spy = sinon.spy();
+      configLoader = new ConfigLoader(initialConfig as Configuration);
```
kriswest: Can you not declare this type with the variable on line 111? Or is there a reason it needs to be an override (as)

test/ConfigLoader.test.ts
```diff
@@ -160,14 +159,14 @@ describe('ConfigLoader', () => {
 
       fs.writeFileSync(tempConfigFile, JSON.stringify(testConfig));
 
-      configLoader = new ConfigLoader(config);
-      const spy = sinon.spy();
+      configLoader = new ConfigLoader(config as Configuration);
```
kriswest: As above, can you declare this on line 146 instead? Would ensure type safety in the test definition

test/ConfigLoader.test.ts
```diff
@@ -177,13 +176,13 @@ describe('ConfigLoader', () => {
         },
       };
 
-      configLoader = new ConfigLoader(config);
-      const spy = sinon.spy();
+      configLoader = new ConfigLoader(config as Configuration);
```
kriswest: as above

test/ConfigLoader.test.ts
```diff
@@ -265,10 +255,10 @@ describe('ConfigLoader', () => {
         },
       };
 
-      configLoader = new ConfigLoader(mockConfig);
-      configLoader.reloadTimer = setInterval(() => {}, 1000);
+      configLoader = new ConfigLoader(mockConfig as Configuration);
```
kriswest: as above

test/ConfigLoader.test.ts
```diff
@@ -310,11 +300,11 @@ describe('ConfigLoader', () => {
         },
       };
 
-      configLoader = new ConfigLoader(mockConfig);
-      configLoader.reloadTimer = setInterval(() => {}, 1000);
-      expect(configLoader.reloadTimer).to.not.be.null;
+      configLoader = new ConfigLoader(mockConfig as Configuration);
```
kriswest: as above

test/ConfigLoader.test.ts
```diff
@@ -310,11 +300,11 @@ describe('ConfigLoader', () => {
...
+      (configLoader as any).reloadTimer = setInterval(() => {}, 1000);
```
kriswest, line 317: Prob worth a comment on this being a private property that we're overriding for test (hence the test is fragile) - unless it can be set properly another way... through the config provided?

test/ConfigLoader.test.ts
```diff
@@ -329,193 +319,162 @@ describe('ConfigLoader', () => {
...
-    it('should throw error if repository is invalid', async function () {
+    it('should throw error if repository is invalid', async () => {
       const source = {
         type: 'git',
         repository: 'invalid-repository',
         path: 'proxy.config.json',
         branch: 'main',
         enabled: true,
-      };
+      } as GitSource;
```
kriswest: Again, why the override? If its needed worth a comment. If its a shortcut, does it make the test more fragile?

test/ConfigLoader.test.ts
```diff
@@ -329,193 +319,162 @@ describe('ConfigLoader', () => {
...
-    it('should throw error if branch name is invalid', async function () {
+    it('should throw error if branch name is invalid', async () => {
       const source = {
         type: 'git',
         repository: 'https://github.com/finos/git-proxy.git',
         path: 'proxy.config.json',
         branch: '..', // invalid branch pattern
         enabled: true,
-      };
+      } as GitSource;
```
kriswest: As above

test/ConfigLoader.test.ts
```diff
@@ -329,193 +319,162 @@ describe('ConfigLoader', () => {
...
-    it('should throw error if configuration source is invalid', async function () {
+    it('should throw error if configuration source is invalid', async () => {
       const source = {
         type: 'invalid',
         repository: 'https://github.com/finos/git-proxy.git',
         path: 'proxy.config.json',
         branch: 'main',
         enabled: true,
-      };
+      } as any;
```
kriswest: This needs to be any as valid type field values are enumerated. However, an override (as) should allow it to compile with GitSource as the type as the TS compiler will trust you that it matches (where an IDE may still show it as an error).

i.e. I thought this was on place where you need the `as GitSource`

test/1.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  // Example test: check db stub is working
+  it('should get the repo from stubbed db', async function () {
+    const repo = await db.getRepo('finos/db-test-repo');
+    expect(repo).toEqual(TEST_REPO);
+  });
+
+  // Example test: use vi.doMock to override the config module
+  it('should return an array of enabled auth methods when overridden', async () => {
+    vi.resetModules(); // Clear module cache
```
kriswest: Why should we clear the cache? What calls in other tests might have affected it and not cleaned up themselves? NOT doing this might be a pattern thats causing issues with subsequent tests (that you mentioned at the last community meeting)

test/ConfigLoader.test.ts
kriswest: I'll stop marking the overrides (`as`) from here.

test/extractRawBody.test.ts
```diff
@@ -0,0 +1,80 @@
...
+      set: vi.fn().mockReturnThis(),
+      status: vi.fn().mockReturnThis(),
+      send: vi.fn(),
+      end: vi.fn(),
+    };
+
+    next = vi.fn();
+
+    (rawBody as Mock).mockClear();
+    (chain.executeChain as Mock).mockClear();
```
kriswest, line 40: Might be better to do this in a beforeAll & afterEach or you'll miss clearing it on the last iteration - not sure that matters.

test/extractRawBody.test.ts
```diff
@@ -0,0 +1,80 @@
+import { describe, it, beforeEach, expect, vi, Mock } from 'vitest';
+import { PassThrough } from 'stream';
+
+// Tell Vitest to mock dependencies
+vi.mock('raw-body', () => ({
```
kriswest, line 5: Do we need to reset the module cache for these afterAll?

test/generated-config.test.ts
```diff
@@ -33,12 +31,12 @@ describe('Generated Config (QuickType)', () => {
...
-      expect(result).to.be.an('object');
-      expect(result.proxyUrl).to.equal('https://proxy.example.com');
-      expect(result.cookieSecret).to.equal('test-secret');
-      expect(result.authorisedList).to.be.an('array');
-      expect(result.authentication).to.be.an('array');
-      expect(result.sink).to.be.an('array');
+      expect(result).toBeTypeOf('object');
+      expect(result.proxyUrl).toBe('https://proxy.example.com');
+      expect(result.cookieSecret).toBe('test-secret');
+      expect(Array.isArray(result.authorisedList)).toBe(true);
```
kriswest: Theres an assertion for this: https://vitest.dev/api/assert.html#isarray

```suggestion
      assert.isArray(result.authorisedList);
```

test/generated-config.test.ts
```diff
@@ -52,27 +50,27 @@ describe('Generated Config (QuickType)', () => {
...
-      };
+      } as GitProxyConfig;
 
       const jsonString = Convert.gitProxyConfigToJson(configObject);
       const parsed = JSON.parse(jsonString);
 
-      expect(parsed).to.be.an('object');
-      expect(parsed.proxyUrl).to.equal('https://proxy.example.com');
-      expect(parsed.cookieSecret).to.equal('test-secret');
+      expect(parsed).toBeTypeOf('object');
```
kriswest:
```suggestion
      assert.isObject(parsed);
```

test/generated-config.test.ts
```diff
@@ -119,18 +117,18 @@ describe('Generated Config (QuickType)', () => {
...
-      expect(result.plugins).to.be.an('array');
-      expect(result.privateOrganizations).to.be.an('array');
-      expect(result.proxyUrl).to.be.a('string');
-      expect(result.rateLimit).to.be.an('object');
-      expect(result.sessionMaxAgeHours).to.be.a('number');
-      expect(result.sink).to.be.an('array');
+      expect(result).toBeTypeOf('object');
+      expect(Array.isArray(result.authentication)).toBe(true);
+      expect(Array.isArray(result.authorisedList)).toBe(true);
+      expect(result.contactEmail).toBeTypeOf('string');
```
kriswest:
```suggestion
      assert.isString(result.contactEmail);
```

test/generated-config.test.ts
```diff
@@ -141,9 +139,9 @@ describe('Generated Config (QuickType)', () => {
 
       try {
         const result = Convert.toGitProxyConfig(JSON.stringify(malformedConfig));
-        expect(result).to.be.an('object');
+        expect(result).toBeTypeOf('object');
```
kriswest: Hmm, if we're expecting the error we should never reach this (previous test was wrong). Perhaps force a failure here instead with a message saying it should have thrown instead - or better rework to use `toThrowError`: https://vitest.dev/api/expect.html#tothrowerror

test/generated-config.test.ts
```diff
@@ -299,18 +297,18 @@ describe('Generated Config (QuickType)', () => {
...
+      expect(result.sessionMaxAgeHours).toBe(0);
+      expect(result.csrfProtection).toBe(false);
+      expect(result.tempPassword).toBeTypeOf('object');
+      expect(result.tempPassword!.length).toBe(12);
     });
 
     it('should test validation error paths', () => {
       try {
         // Try to parse something that looks like valid JSON but has wrong structure
         Convert.toGitProxyConfig('{"proxyUrl": 123, "authentication": "not-array"}');
```
kriswest: test should fail if it didn't throw but doesn't (never has)

test/processors/checkAuthorEmails.test.ts
```diff
@@ -0,0 +1,654 @@
...
+
+      mockAction.commitData = [
+        { authorEmail: 'user@EXAMPLE.COM' } as Commit,
+        { authorEmail: 'user@Example.Com' } as Commit,
+      ];
+
+      const result = await exec(mockReq, mockAction);
+
+      const step = vi.mocked(result.addStep).mock.calls[0][0];
+      // fails because regex is case-sensitive
```
kriswest: It probably shouldn't be case sensitive as most email providers aren't... That should be a quick fix in `isEmailAllowed` (change the flags on the regex's to `'gi'`)

test/processors/scanDiff.test.ts
```diff
@@ -56,252 +55,256 @@ index 8b97e49..de18d43 100644
...
 
-  it('A diff including an AWS (Amazon Web Services) Access Key ID blocks the proxy...', async () => {
+  it('should block push when diff includes AWS Access Key ID', async () => {
     const action = new Action('1', 'type', 'method', 1, 'test/repo.git');
     action.steps = [
       {
         stepName: 'diff',
         content: generateDiff('AKIAIOSFODNN7EXAMPLE'),
-      },
+      } as Step,
```
kriswest: is the override needed?

test/proxy.test.ts
```diff
@@ -0,0 +1,156 @@
+import https from 'https';
+import { describe, it, beforeEach, afterEach, expect, vi } from 'vitest';
+import fs from 'fs';
+
+// TODO: rewrite/fix these tests
```
kriswest: What would you want to achieve in a rewrite?

test/services/routes/auth.test.ts
```diff
@@ -0,0 +1,239 @@
...
+      vi.restoreAllMocks();
+    });
+
+    it('POST /gitAccount returns Unauthorized if authenticated user not in request', async () => {
+      const res = await request(newApp()).post('/auth/gitAccount').send({
+        username: 'alice',
+        gitAccount: '',
+      });
+
+      expect(res.status).toBe(401);
```
kriswest, line 65: If the user is not logged in we should return 401 (Unauthorized), if they are logged in as the wrong user for the action we should return 403 (Forbidden). This may need splitting into two tests one for each response.

kriswest: There is a subsequent test: POST /gitAccount prevents non-admin user changing a different user gitAccount
which accounts for the second case. This is probably just badly named (user not in the request = user is not logged in?)

test/services/routes/auth.test.ts
```diff
@@ -0,0 +1,239 @@
...
+          gitAccount: '',
+          title: '',
+          username: 'bob',
+        },
+      });
+    });
+  });
+
+  describe('/me', () => {
+    it('GET /me returns Unauthorized if authenticated user not in request', async () => {
```
kriswest: if user is not logged in?

test/services/routes/auth.test.ts
```diff
@@ -0,0 +1,239 @@
...
+        email: 'alice@example.com',
+        title: '',
+        gitAccount: '',
+        admin: false,
+      });
+    });
+  });
+
+  describe('/profile', () => {
+    it('GET /profile returns Unauthorized if authenticated user not in request', async () => {
```
kriswest: if user is not logged in?

packages/git-proxy-cli/package.json
```diff
@@ -11,7 +11,8 @@
     "@finos/git-proxy": "file:../.."
   },
   "devDependencies": {
-    "chai": "^4.5.0"
+    "chai": "^4.5.0",
```
kriswest: This doesn't seem right (import of chai and ts-mocha remaining). The CLI tests are importing from vitest.

packages/git-proxy-cli/package.json
```diff
@@ -11,7 +11,8 @@
     "@finos/git-proxy": "file:../.."
```
kriswest: Not this PR, but we can either give the current version number or "0.0.0" instead of the file URL.

test/testActiveDirectoryAuth.test.ts
```diff
@@ -0,0 +1,170 @@
...
+        searchBase: 'ou=users,dc=test,dc=com',
+      },
+    },
+  ],
+});
+
+describe('ActiveDirectory auth method', () => {
+  beforeEach(async () => {
+    vi.clearAllMocks();
+    vi.resetModules();
```
kriswest: Do you also need an afterAll to reset from the last test?

test/testCheckUserPushPermission.test.ts
```diff
@@ -14,24 +12,22 @@ const TEST_USERNAME_2 = 'push-perms-test-2';
 const TEST_EMAIL_2 = 'push-perms-test-2@test.com';
 const TEST_EMAIL_3 = 'push-perms-test-3@test.com';
 
-describe('CheckUserPushPermissions...', async () => {
-  let testRepo = null;
+describe('CheckUserPushPermissions...', () => {
+  let testRepo: any = null;
```
kriswest:
```suggestion
  let testRepo: Repo | null = null;
```

test/testDb.test.ts
```diff
@@ -0,0 +1,667 @@
...
+      // @ts-expect-error dynamic indexing
+      cleanResponse[col] = responses[col];
+    });
+    return cleanResponse as T;
+  } else {
+    throw new Error(`Can only clean arrays or objects, but a ${typeof responses} was passed`);
+  }
+};
+
+// Use this test as a template
```
kriswest:
```suggestion
```

This comment probably came in from the template

test/testParseAction.test.ts
```diff
@@ -1,83 +1,84 @@
...
+    // make sure the test repo exists as the presence of the repo makes a difference to handling of urls
     testRepo = await db.getRepoByUrl(TEST_REPO.url);
     if (!testRepo) {
       testRepo = await db.createRepo(TEST_REPO);
     }
   });
-  after(async function () {
+
+  afterAll(async () => {
     // clean up test DB
```
kriswest:
```suggestion
    // If we created the testRepo, clean it up
```

test/testPush.test.ts
```diff
@@ -0,0 +1,350 @@
...
+                tooltip: {
+                  text: 'Are you happy for this contribution to be pushed upstream?',
+                  links: [],
+                },
+                checked: false,
+              },
+            ],
+          },
+        });
+      expect(res.status).toBe(401);
```
kriswest, line 184: This should probably be a 403 - not an issue for this PR

test/testPush.test.ts
```diff
@@ -0,0 +1,350 @@
...
+                tooltip: {
+                  text: 'Are you happy for this contribution to be pushed upstream?',
+                  links: [],
+                },
+                checked: true,
+              },
+            ],
+          },
+        });
+      expect(res.status).toBe(401);
```
kriswest, line 210: This should probably be a 403

test/testPush.test.ts
```diff
@@ -0,0 +1,350 @@
...
+              tooltip: {
+                text: 'Are you happy for this contribution to be pushed upstream?',
+                links: [],
+              },
+              checked: true,
+            },
+          ],
+        },
+      });
+    expect(res.status).toBe(401);
```
kriswest, line 239: This should probably be a 403

test/testPush.test.ts
```diff
@@ -0,0 +1,350 @@
...
```
kriswest, line 263: This should probably be a 403, etc. will stop marking them here

vitest.config.ts
```diff
@@ -0,0 +1,35 @@
+import { defineConfig } from 'vitest/config';
```
kriswest, line 1: Should there be a vitest config file for the cli tests as well?

test/generated-config.test.ts
kriswest:
```suggestion
        assert.fail('Malformed context should have thrown an error!');
```

kriswest: or 
```suggestion
     expect(Convert.toGitProxyConfig(JSON.stringify(malformedConfig))).toThrowError();
```
and lose the try/catch

or even:

```suggestion
expect(Convert.toGitProxyConfig(JSON.stringify(malformedConfig))).toThrowError('text of error');
```
see: https://vitest.dev/api/expect.html#tothrowerror

I'm liking vitest's expect and assert implementations!

config.schema.json
jescalada: I don't think I have the `generate-schema-doc` program available anywhere (thus `npm run gen-schema-doc` fails). Is there a specific source for it? I haven't seen it in the npm registry...

kriswest: You won't, its a python lib: https://github.com/coveooss/json-schema-for-humans that you have to have installed (not ideal). AFAIK there's no documentation in the project about this... we should put something in /website/README.md

jescalada: Made an issue for this: #1280

src/config/index.ts
jescalada: Setting these top-level properties as required ends up causing a bunch of `ConfigLoader`-related tests to fail because of the missing properties... I lean towards making them optional for flexibility since those test cases were likely based on actual use cases 🤔

src/config/index.ts
jescalada: I looked into setting default values for these in the schema config, but unfortunately the generated types don't enforce the defaults anywhere.

Maybe we could have a `constants.ts` file to centralize these - or perhaps a `defaults.ts` file to make sure these are consistent? Right now we have default values scattered between the `proxy.config.json`, `config/index.ts` and a few other spots as a fallback for the `getSomeConfigValue()` functions.

src/proxy/processors/push-action/checkCommitMessages.ts
jescalada: Made an issue to track this: #1281

src/proxy/processors/push-action/checkCommitMessages.ts
jescalada: Linked this related comment to #1281!

src/service/routes/push.ts
jescalada: I've done a quick cleanup of the really unneeded logs, mostly in `push.ts`.

src/config/index.ts
kriswest: Hence the comment about having two top-level schemas, one for the full config with all required elements required, and one for overrides... Makes maintenance more complex however. I wonder if we could achieve it in Typescript alone with either the `Required<type>` or `Partial<type>` utiulity types? I think wrapping Required<type> around the current generated type would make all the top-level properties required and give us the type safety where we need it without extra schema?

WDYT?

src/config/index.ts
kriswest: See previous comment about using `Required`. Types can't enforce the defaults, but they can enforce a value being set. If the value is required to be set (in a full configuration) then no default is needed. That could be a neat solution, i.e. functions like loadFullConfiguration returning `Required<GitProxyConfig>` - although they might need to confirm that all those values are set on load and throw an error if not (e.g. because someone messed with the default config).

test/ConfigLoader.test.ts
jescalada: That worked!

kriswest: Great! The problem with `as` is that it overrides checks in the TS compiler - you're effectively saying 'trust me this is type X, don't bother checking it'. An IDE can still flag issues, but the TS compiler won't, so regular use of `as` is a bad pattern (but sometimes necessary, e.g. if mocking something with missing properties that you know won't be checked).

kriswest: (at least that's my understanding of overrides)

test/generated-config.test.ts
jescalada: Using `assert` seemed to work for me:

```ts
it('should handle malformed configuration gracefully', () => {
  const malformedConfig = {
    proxyUrl: 123, // Wrong type
    authentication: 'not-an-array', // Wrong type
  };

  assert.throws(() => Convert.toGitProxyConfig(JSON.stringify(malformedConfig)));
});
```

test/generated-config.test.ts
jescalada: This one worked with the following:

```ts
assert.throws(() => Convert.toGitProxyConfig('{"proxyUrl": 123, "authentication": "not-array"}'));
```

test/extractRawBody.test.ts
jescalada: The test file and whole suite seems to pass just fine without those two lines - removing them 🤔

test/extractRawBody.test.ts
jescalada: I think cleaning up in `afterAll` is correct, in case we have other test files making use of the `raw-body` library. I'm pretty sure the `src/proxy/chain` is being used in other places but for some reason not clearing the cache doesn't make those tests fail...

test/processors/scanDiff.test.ts
jescalada: The only way to remove the override is by adding all the missing properties it seems:

```ts
  it('should block push when diff includes AWS Access Key ID', async () => {
    const action = new Action('1', 'type', 'method', 1, 'test/repo.git');
    const diffStep: Step = {
      stepName: 'diff',
      content: generateDiff('AKIAIOSFODNN7EXAMPLE'),
      error: false,
      errorMessage: null,
      blocked: false,
      blockedMessage: null,
      logs: [],
      id: '1',
      setError: vi.fn(),
      setContent: vi.fn(),
      setAsyncBlock: vi.fn(),
      log: vi.fn(),
    };
    action.steps = [diffStep];
```

If that's still acceptable, I could extract the `diffStep` creation to a helper function to avoid casting?

jescalada: I ended up making a `generateDiffStep` helper function for this which also helped remove casting in `scanDiff.emptyDiff.test.ts`, as well as cutting down the line count 👍🏼

test/services/routes/auth.test.ts
jescalada: I added extra tests for the status codes and renamed the existing ones for clarity.

I'm keeping the one you mentioned since it checks the `updateUser` database call - I feel it deserves it's own "should" entry for completeness.

test/testPush.test.ts
jescalada: Made an issue for these: #1285

test/testActiveDirectoryAuth.test.ts
jescalada: I think setting those in an `afterEach` fixes any issues here 🤔

packages/git-proxy-cli/package.json
jescalada: I totally forgot to refactor the CLI to Vitest! 😓 Had some trouble with Vitest executing directly from the CLI, and decided to add the CLI test scripts to the parent package instead (and adjust the CI accordingly).

What are your thoughts on this? @kriswest

vitest.config.ts
jescalada: Apparently this is necessary if having the tests run from the CLI, however, I had trouble getting the tests to work with the workspace/subpackage setup... Running the CLI tests from the parent package works perfectly however.

src/config/index.ts
jescalada: I looked into this in more detail, and I think it's possible to wrap the top-level properties as `Required` in a post-processing script after generating the types. It might be worth documenting this behaviour - however I don't think it's worth making it configurable since it'd add too much complexity to the implementation.

In this case, config overriding tests will still pass, but fallback values for `config.<property>` can be safely removed.

Made an issue for this: #1287

src/config/index.ts
jescalada: Wouldn't this prevent overriding the config (what some config tests are doing)?

packages/git-proxy-cli/package.json
jescalada: After all, I managed to get the CI checks to pass by executing the CLI tests from the parent package. I experimented with Vitest `--coverage` but couldn't get it to accurately cover the CLI files due to some strange issues with package resolution (which apparently `ts-mocha` was handling correctly).

However, this does allow us to cut out a lot of the chai/mocha package bloat, and it's only the CLI coverage that's missing.

vitest.config.ts
jescalada: Not necessary with the current working setup that runs from the parent package 👍🏼

src/config/index.ts
kriswest: No it shouldn't. Where we're passing overrides you use `GitProxyConfig` where the top-level properties are optional. But when we are operating over the full configuration (defaults overlaid by overrides) use `Required<GitProxyConfig>`. The defaults are supposed to meet `Required<GitProxyConfig>` so if we've merged overrides with it (that are valid `GitProxyConfig`s) we should still meet `Required<GitProxyConfig>`.  Hence if `loadConfiguration` returns `Required<GitProxyConfig>` (might have to use an override or do validation) you won't have to check the tpo-level properties are defined all over the place.

src/config/index.ts
kriswest: I think you can get away without post processing and use the `Required<>` utility type where you know everything is set (i.e. after merging with defaults or manually validating all are set) - see the next comment for more on that.

test/1.test.ts
jescalada: I've improved the explanations here:

```ts
// Runs after each test
afterEach(function () {
  // Restore all stubs: This cleans up replaced behaviour on existing modules
  // Required when using vi.spyOn or vi.fn to stub modules/functions
  vi.restoreAllMocks();

  // Clear module cache: Wipes modules cache so imports are fresh for the next test file
  // Required when using vi.doMock to override modules
  vi.resetModules();
});
```

test/proxy.test.ts
jescalada: Explained this in more detail:

```ts
/* 
  jescalada: these tests are currently causing the following error
  when running tests in the CI or for the first time locally:
  Error: listen EADDRINUSE: address already in use :::8000

  This is likely due to improper test isolation or cleanup in another test file
  especially related to proxy.start() and proxy.stop() calls

  Related: skipped tests in testProxyRoute.test.ts - these have a race condition
  where either these or those tests fail depending on execution order
  TODO: Find root cause of this error and fix it
*/
describe.skip('Proxy Module TLS Certificate Loading', () => {
```

src/config/index.ts
jescalada: Makes sense! I'll look into implementing this over the next few days/weeks - not sure if it'll make it into the 2.0.0 release...

I've linked these comments to the related issue (#1287) so we can move this PR forward. 😃

test/proxy.test.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    const ProxyClass = (await import('../src/proxy/index')).default;
+    proxyModule = new ProxyClass();
+  });
+
+  afterEach(async () => {
+    try {
+      await proxyModule.stop();
+    } catch {
+      // ignore cleanup errors
```
kriswest: might be worth logging here to check if we're failing to close the proxy

test/proxy.test.ts
```diff
@@ -0,0 +1,167 @@
...
+  when running tests in the CI or for the first time locally:
+  Error: listen EADDRINUSE: address already in use :::8000
+
+  This is likely due to improper test isolation or cleanup in another test file
+  especially related to proxy.start() and proxy.stop() calls
+
+  Related: skipped tests in testProxyRoute.test.ts - these have a race condition
+  where either these or those tests fail depending on execution order
+  TODO: Find root cause of this error and fix it
+*/
```
kriswest, line 17: If not resolving in this PR can you raise an issue for it and put the number here.

test/ConfigLoader.test.ts
kriswest: still overridden here, does it need to be?

test/ConfigLoader.test.ts
kriswest: still overridden here, does it need to be?

test/ConfigLoader.test.ts
```diff
@@ -329,193 +332,162 @@ describe('ConfigLoader', () => {
...
-      this.timeout(10000);
-
       const source = {
         type: 'git',
         repository: 'https://github.com/finos/git-proxy.git',
         path: 'proxy.config.json',
         branch: 'main',
         enabled: true,
-      };
+      } as GitSource;
```
kriswest: Does this need to be an override rather than declaration?

src/proxy/processors/push-action/checkAuthorEmails.ts
```diff
@@ -14,14 +14,14 @@ const isEmailAllowed = (email: string): boolean => {
 
   if (
     commitConfig?.author?.email?.domain?.allow &&
-    !new RegExp(commitConfig.author.email.domain.allow, 'g').test(emailDomain)
+    !new RegExp(commitConfig.author.email.domain.allow, 'gi').test(emailDomain)
```
kriswest, line 17: :+1

test/ConfigLoader.test.ts
jescalada: Whoops, I totally missed these! I fixed up some the `HttpSource` and `ConfigurationSource` casting too.

test/proxy.test.ts
jescalada: Added an issue for this: #1294 Thanks for the heads-up!

kriswest:
```suggestion
  TODO: Find root cause of this error and fix it 
  https://github.com/finos/git-proxy/issues/1294
*/
```

jescalada: 👍🏼!

test/proxy.test.ts
kriswest: Is this a syntax error? I don't think I've ever seen a catch without arguments. I wonder if its breaking the afterEach or hiding the underlying cause:

```suggestion
    } catch (err) {
      console.error("Error occurred when stopping the proxy: ", err);
```

jescalada: This is odd, I think it's from the previous tests which I converted a bit too carelessly...

I played with it for a while and removing the try/catch entirely doesn't seem to change anything. It's definitely good to log the errors properly though!

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -222,8 +222,6 @@ const getCommitData = (contents: CommitContent[]): CommitData[] => {
     .chain(contents)
     .filter({ type: GIT_OBJECT_TYPE_COMMIT })
     .map((x: CommitContent) => {
-      console.log({ x });
```
coopernetes, line 225: Marking this as related to #399 - these are chatty logs so fine to remove them. However, we need to uplift the overall app's logging and should use a proper logging framework.

test/processors/clearBareClone.test.ts
coopernetes, line 1: As a future exercise for large refactors like this, let's try to use `git mv` so that the file renames are tracked AND any changed content is properly diff'd like this file. Most of the other changes are difficult to reason about since the entire file was "moved" (deleted & recreated as a .ts) from git's POV.

test/testActiveDirectoryAuth.test.ts
coopernetes, line 1: Similar to how we are using supertest to do HTTP integration tests, it would be nice to have "real mocked" authentication providers such as an in-memory LDAP server so we can validate this behaviour without mocks. For now, these tests are fine. Whenever we need to do a lock of mocking like this, we should probably refactor into integration tests with ephemeral backends (replacing DB & LDAP stubs with "real" implementations since we do not want to mock out database driver or server adapter behaviours directly).

test/ui/apiBase.test.ts
```diff
@@ -0,0 +1,50 @@
...
+  beforeAll(() => {
+    globalThis.location = { origin: 'https://lovely-git-proxy.com' } as any;
+  });
+
+  afterAll(() => {
+    globalThis.location = originalLocation;
+  });
+
+  beforeEach(() => {
+    originalEnv = process.env.VITE_API_URI;
```
coopernetes, line 22: This is a common misconception with [Vite environment variables](https://vite.dev/guide/env-and-mode). They are only injected at _build_ time, *not* at run time. My PR (#1165) attempts to make this properly dynamic at run time. It will need proper testing so I'm gonna just make a note here as a reminder to have an issue for this specifically.

https://github.com/RBC/finos-git-proxy/blob/24b91e46fc6288fcba5e8b948937e96a66857aa5/src/ui/services/apiConfig.ts

https://github.com/RBC/finos-git-proxy/blob/24b91e46fc6288fcba5e8b948937e96a66857aa5/src/ui/services/runtime-config.ts

https://github.com/finos/git-proxy/pull/1165/files#diff-79738685a656fe6b25061bb14181442210b599f746faeaba408a2401de45038a

test/testProxy.test.ts
```diff
@@ -0,0 +1,237 @@
+import { describe, it, expect, beforeEach, afterEach, vi, afterAll } from 'vitest';
+
+vi.mock('http', async (importOriginal) => {
```
coopernetes, line 3: This is a code smell and should be refactored into a proper Express integration test. We shouldn't be mocking out server machinery directly. Separating out the creation of the `express.App` instance can help make testing a bit easier (passing around a mock app is much simpler than mocking node internals directly).

test/testActiveDirectoryAuth.test.ts
coopernetes: Added this comment https://github.com/finos/git-proxy/issues/1143#issuecomment-3603326206

test/ui/apiBase.test.ts
coopernetes: Opened new issue #1302 to track this.

test/testProxy.test.ts
coopernetes: Added a comment to #1143 in regards to this.

packages/git-proxy-cli/package.json
coopernetes: Marking this as resolved.

test/processors/clearBareClone.test.ts
jescalada: Thank you for the tip! I didn't know that was the correct way to refactor.

PR #1204, opened 2025-09-15
fix: incorrect error message on cloning unauthorized repos

### LINKED ISSUE #1181: Incorrect error message on cloning unauthorized repo
**Describe the bug**
When cloning a repo that isn't in the `authorisedList` or in the repo database, an incorrect/unrelated error message appears.

**To Reproduce**
Choose a repo that isn't in the `authorisedList` or in the repo database, then clone it:

1. Run GitProxy
2. Execute the following (replace with invalid repo)
```bash
git clone http://localhost:8080/open-webui/open-webui.git`
``` 

This gives the following error:

```
fatal: http://localhost:8000/open-webui/open-webui.git/info/refs not valid: could not determine hash algorithm; is this a git repository?`
```

Adding the repo to the database or `authorisedList` will let the clone go through (therefore the command is correct).

Image

**Expected behavior**
This flow should display a relevant error, such as `REPO_NAME not in authorisedList or database`. 

**Screenshots**

### Proxy output
Image

### `git clone` output

Image

fabiovincenzi:
Fixes #1181 by implementing proper Git protocol error handling for `GET /info/refs requests`, replacing confusing error messages when cloning unauthorized repositories.

  The issue was a Git protocol mismatch in error handling:

  1. GET /info/refs requests (clone operations) require a specific error packet format (https://git-scm.com/docs/gitprotocol-pack)
  2. Previous implementation used POST-oriented sideband format (\x02 + tab formatting) for all requests
  3. Git client received malformed upload-pack data → parsing failed → generic "hash algorithm" error
  
  
Screenshot 2025-09-15 at 16 02 13

### DIFF
--- a/src/proxy/routes/index.ts
+++ b/src/proxy/routes/index.ts
@@ -212,7 +212,8 @@ const getRouter = async () => {
         proxyReqOptDecorator: proxyReqOptDecorator,
         proxyReqBodyDecorator: proxyReqBodyDecorator,
         proxyErrorHandler: proxyErrorHandler,
-      }),
+        stream: true,
+      } as any),
     );
   });
 
@@ -225,7 +226,8 @@ const getRouter = async () => {
     proxyReqOptDecorator: proxyReqOptDecorator,
     proxyReqBodyDecorator: proxyReqBodyDecorator,
     proxyErrorHandler: proxyErrorHandler,
-  });
+    stream: true,
+  } as any);
 
   console.log('proxy keys registered: ', JSON.stringify(proxyKeys));
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -3625,9 +3625,9 @@
       "license": "MIT"
     },
     "node_modules/axios": {
-      "version": "1.11.0",
-      "resolved": "https://registry.npmjs.org/axios/-/axios-1.11.0.tgz",
-      "integrity": "sha512-1Lx3WLFQWm3ooKDYZD1eXmoGO9fxYQjrycfHFC8P0sCfQVXyROp0p9PFWBehewBOdCwHc+f/b8I0fMto5eSfwA==",
+      "version": "1.12.2",
+      "resolved": "https://registry.npmjs.org/axios/-/axios-1.12.2.tgz",
+      "integrity": "sha512-vMJzPewAlRyOgxV2dU0Cuz2O8zzzx9VYtbJOaBgXFeLc4IV/Eg50n4LowmehOOR61S8ZMpc2K5Sa7g6A4jfkUw==",
       "license": "MIT",
       "dependencies": {
         "follow-redirects": "^1.15.6",
@@ -14627,7 +14627,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@finos/git-proxy": "file:../..",
-        "axios": "^1.11.0",
+        "axios": "^1.12.2",
         "yargs": "^17.7.2"
       },
       "bin": {
--- a/packages/git-proxy-cli/package.json
+++ b/packages/git-proxy-cli/package.json
@@ -4,7 +4,7 @@
   "description": "Command line interface tool for FINOS GitProxy.",
   "bin": "./index.js",
   "dependencies": {
-    "axios": "^1.11.0",
+    "axios": "^1.12.2",
     "yargs": "^17.7.2",
     "@finos/git-proxy": "file:../.."
   },
--- a/src/proxy/processors/push-action/checkAuthorEmails.ts
+++ b/src/proxy/processors/push-action/checkAuthorEmails.ts
@@ -30,8 +30,6 @@ const isEmailAllowed = (email: string): boolean => {
 };
 
 const exec = async (req: any, action: Action): Promise<Action> => {
-  console.log({ req, action });
-
   const step = new Step('checkAuthorEmails');
 
   const uniqueAuthorEmails = [
--- a/src/proxy/processors/push-action/checkCommitMessages.ts
+++ b/src/proxy/processors/push-action/checkCommitMessages.ts
@@ -51,8 +51,6 @@ const isMessageAllowed = (commitMessage: string): boolean => {
 
 // Execute if the repo is approved
 const exec = async (req: any, action: Action): Promise<Action> => {
-  console.log({ req, action });
-
   const step = new Step('checkCommitMessages');
 
   const uniqueCommitMessages = [...new Set(action.commitData?.map((commit) => commit.message))];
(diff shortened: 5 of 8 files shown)


src/proxy/routes/index.ts
```diff
@@ -49,6 +49,25 @@ const proxyFilter: ProxyOptions['filter'] = async (req, res) => {
     const action = await executeChain(req, res);
 
     if (action.error || action.blocked) {
+      const errorMessage = action.errorMessage ?? action.blockedMessage ?? '';
+
+      // For GET requests to /info/refs (clone operations), use Git protocol error packet format
```
kriswest: I think this is also used in fetch and a variety of other operations - the code is handling those correctly, but the comment and function name could be improved. 

```suggestion
      // GET requests to /info/refs (used to check refs for many git operations) must use Git protocol error packet format
```

Some notes on /info/refs:

> In Git, requests to the /info/refs endpoint of a repository are typically made during initial interactions with a remote repository, especially when using smart HTTP protocol. These requests help the client discover what references (branches, tags, etc.) are available and what capabilities the server supports.
> 
> Here are the main Git client operations that generate requests to /info/refs:
> 
> 🔹 1. Clone
> Command: git clone <repo-url>
> Purpose: The client queries /info/refs?service=git-upload-pack to discover refs and capabilities before downloading the repository.
>
> 🔹 2. Fetch / Pull
> Command: git fetch, git pull
> Purpose: If the client hasn’t interacted with the remote before or needs to update its understanding of available refs, it may query /info/refs?service=git-upload-pack.
>
> 🔹 3. Push
> Command: git push
> Purpose: The client queries /info/refs?service=git-receive-pack to discover what refs exist on the remote and what capabilities are supported before pushing changes.
>
> 🔹 4. LFS (Large File Storage) Operations
> Command: git lfs fetch, git lfs push
> Purpose: Git LFS may also interact with /info/refs to discover endpoints and capabilities for large file transfers.
>
> 🔹 5. Git Archive (via HTTP)
> Command: git archive --remote=<repo-url>
> Purpose: This can trigger /info/refs?service=git-upload-archive to discover archive capabilities.
>
> Notes:
> These requests are part of the smart HTTP protocol, which is more efficient than the older dumb HTTP protocol.
> The query parameter ?service=git-upload-pack or ?service=git-receive-pack indicates the operation type (fetch/pull vs push).

src/proxy/routes/index.ts
```diff
@@ -104,6 +123,13 @@ const handleMessage = (message: string): string => {
   return `${len}\x02${body}\n0000`;
 };
 
+const handleGetMessage = (message: string): string => {
```
kriswest: The name could be more specific - perhaps more important as we're likely to hit this often when blocking requests as it usually precedes the other requests being made (in clone, fetch, pull and push)

```suggestion
const handleRefsErrorMessage = (message: string): string => {
```

kriswest (commented):
Great work figuring this out! I've dug out some more info on when requests to /info/refs occur, to inform comments and think you should make a function name more specific - but otherwise very happy to see this solved.

src/proxy/routes/index.ts
```diff
@@ -101,7 +101,34 @@ const proxyFilter: ProxyOptions['filter'] = async (req, res) => {
...
+
+    // GET requests to /info/refs (used to check refs for many git operations) must use Git protocol error packet format
+    if (req.method === 'GET' && req.url.includes('/info/refs')) {
+      res.set('content-type', 'application/x-git-upload-pack-advertisement');
+
+      logAction(
+        req.url,
+        req.headers.host,
+        req.headers['user-agent'],
+        'Error occurred in proxy filter function: ' + ((e as Error).message ?? e),
```
kriswest:
```suggestion
       errorMessage,
```

src/proxy/routes/index.ts
```diff
@@ -102,46 +66,57 @@ const proxyFilter: ProxyOptions['filter'] = async (req, res) => {
     return true;
   } catch (e) {
     const errorMessage = `Error occurred in proxy filter function ${e}`;
+    const logMessage = 'Error occurred in proxy filter function: ' + ((e as Error).message ?? e);
```
kriswest: These two lines (`errorMessage` and `logMessage`) are constructing the same message, ones just better written than the other (handling strings or errors). There's no need for both.

I'd replace both with: 
```typescript
const errorMessage = `Error occurred in proxy filter function ${(e as Error).message ?? e}`;
```

src/proxy/routes/index.ts
```diff
@@ -82,20 +65,58 @@ const proxyFilter: ProxyOptions['filter'] = async (req, res) => {
     // this is the only case where we do not respond directly, instead we return true to proxy the request
     return true;
   } catch (e) {
-    const packetMessage = handleMessage(`Error occurred in proxy filter function ${e}`);
+    const errorMessage = `Error occurred in proxy filter function ${e}`;
+    const logMessage = 'Error occurred in proxy filter function: ' + ((e as Error).message ?? e);
```
kriswest: There's no need to duplicate these

```suggestion
    const errorMessage = `Error occurred in proxy filter function: ${(e as Error).message ?? e}`;
```

src/proxy/routes/index.ts
```diff
@@ -82,20 +65,58 @@ const proxyFilter: ProxyOptions['filter'] = async (req, res) => {
...
+
+    sendErrorResponse(req, res, errorMessage, logMessage, null);
```
kriswest:
```suggestion
    sendErrorResponse(req, res, errorMessage, errorMessage, null);
```

kriswest (approved):
One thing to simplify, otherwise LGTM!

src/proxy/routes/index.ts
```diff
@@ -88,7 +86,7 @@ const sendErrorResponse = (
       req.url,
       req.headers.host,
       req.headers['user-agent'],
-      actionErrorMessage,
+      errorMessage,
```
kriswest: Losing the extra arg to `sendErrorMessage` will cause a very minor issue in `logAction`, as `errorMessage` will always be set so It'll always log as "Error" instead of "Blocked". You can either reintroduce the arg, or swap the order of the checks up on lines 20-25 and change it to an if/else.

fabiovincenzi: Good catch! I thought the error handling logic was a bit confusing and hard to follow, so I redesigned it completely. Should be much cleaner now and properly distinguish between errors and blocks in the logs.

src/proxy/routes/index.ts
```diff
@@ -6,23 +6,31 @@ import { executeChain } from '../chain';
...
+  BLOCKED = 'Blocked',
+}
+
 const logAction = (
   url: string,
   host: string | null | undefined,
   userAgent: string | null | undefined,
-  errMsg: string | null | undefined,
-  blockMsg?: string | null | undefined,
+  type: ActionType,
```
kriswest, line 22: 👍

PR #1209, opened 2025-09-19
refactor: eliminate duplicate executeChain calls in push approval flow

### LINKED ISSUE #1196: Investigate duplication of process steps in push approval flow
**Describe the bug**
After some profiling on #985, we found out that when pushing a PR that has been approved, the action chain gets executed twice (see discussion there for more details).

**To Reproduce**

1. Make a PR and push it
2. Log into the UI and approve it
3. Push it to the remote

**Expected behavior**
Unless this is intended, the action chain should execute only once in each push.

**Screenshots**
See profiling screenshots in #985 discussion.

fabiovincenzi:
Fixes #1196 
Refactor of `teeAndValidate` middleware to `extractRawBody` that only extracts request body, centralized all validation logic in `proxyFilter` function, and eliminated duplicate `executeChain()` calls for POST git-pack requests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
         "env-paths": "^2.2.1",
         "express": "^4.21.2",
         "express-http-proxy": "^2.1.2",
-        "express-rate-limit": "^7.5.1",
+        "express-rate-limit": "^8.1.0",
         "express-session": "^1.18.2",
         "history": "5.3.0",
         "isomorphic-git": "^1.33.1",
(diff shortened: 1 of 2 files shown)


fabiovincenzi:
@kriswest 
The stalling wasn't raw-body itself but the PassThrough stream highWaterMark being too small (16KB default).
Fixed by increasing it to 4MB.
Now handles big pushes as well.

kriswest (approved):
LGTM

PR #955, opened 2025-03-25
chore: update to eslint v9

06kellyjac:
Replaces #890
Replaces #893

Raises some warnings but fixing them isn't possible without formatting many files.
I'll do that in another PR.

PR #1218, opened 2025-09-24
fix: linter warnings and CI failure

### LINKED ISSUE #1212: Promise executor functions should not be async in DB code
**Describe the bug**
Linting reports "Promise executor functions should not be async" in [src/db/index.ts](https://github.com/finos/git-proxy/pull/955/files#diff-c7d2c320d1f7fc0aad36b8bf616a7289ca4dcf00e6e057ba8a1d354fb25b74c0) and [src/db/file/repo.ts](https://github.com/finos/git-proxy/pull/955/files#diff-32eee361bff29d74e83df190e4a97f2ea5d2e8701c080e3d499ec3e58b67f7d3). Update their implementations to use promise chains and ensure that if an error occurs the promise rejects (which may not be the case currently where the implementation lacks try/catch clauses).

**To Reproduce**
Steps to reproduce the behaviour:

1. Run linting - currently via #955 (which will be merged shortly)

**Expected behavior**
Clean report from linting

jescalada:
Fixes #1212 (linter warnings/CI failures that popped up after #955).

Notably, some of the errors were due to wrapping promises for some of the database functions (this is only necessary when calling neDB functions directly, because they contain callbacks which can't be intercepted otherwise).


jescalada:
@kriswest Note that the Promise wrappers are not entirely gone, apparently these are required to use `neDB`'s callback functions. An alternative would be to use `promisify`, not sure if this would change anything as the linter errors are already gone 🤔

kriswest (approved):
Approved - easy to review as I've just implemented the same changes and was about to raise the PR. I'd used promise chains in a couple of places and slightly more error handling, but this also looks fine to me.

src/db/file/repo.ts
```diff
@@ -134,21 +133,21 @@ export const addUserCanPush = async (_id: string, user: string): Promise<void> =
...
-  return new Promise(async (resolve, reject) => {
-    const repo = await getRepoById(_id);
-    if (!repo) {
-      reject(new Error('Repo not found'));
-      return;
-    }
+  const repo = await getRepoById(_id);
+  if (!repo) {
+    throw new Error('Repo not found');
+    return;
```
kriswest: this is unreachable

kriswest (changes requested):
theres one unreachable line, I'm surprised the linter isn't complaining about it

src/db/file/repo.ts
06kellyjac: I guess you mean due to how `addUserCanAuthorise` is used the repo with _id will always exist?
The linter isn't fully aware of that & our usage could change (by mistake) going forward.
And there's also probably the potential of a race condition where it did exist when last checked/used & was then deleted before we call `getRepoById` here.

Or if ive overlooked something the eslint config might need fixing for always unreachable code.

kriswest: I meant because of the `throw` right before the `return`...

kriswest (approved):
LGTM

src/db/file/repo.ts
jescalada: This can be fixed by adding `'no-unreachable': 'error'` to our current ESLint config (just tested and it works). Should we do that? 🤔

PR #1225, opened 2025-09-25
chore(experimental): update license-inventory and move to vitest

06kellyjac:
- **chore(experimental): update license-inventory and move to vitest**
- **chore(experimental): update li-cli and move to vitest**

Jest is difficult to use with ESM and git-proxy main will be moving to vitest


coopernetes:
This has been in draft for awhile and I presume the ideas as part of this experimentation hasn't been nailed down yet. Closing. Feel free to re-open if its worth revisiting. It would be good to uplift this to be compatible with the new 2.x conventions  (testing already moved over to vitest).

PR #1227, opened 2025-09-25
feat: improve login page flexibility

### LINKED ISSUE #1206: Login page shows irrelevant auth options & generic errors
## Problem

The configuration only allows one authentication mechanism.

The login page is hard-coded to support all options and the user is expected to know which one to use.

We are using Okta-based SSO, which makes the login form unnecessary. Users can sign in without entering any credentials, but the presence of the form gives the impression that a username and password are required. When they attempt to log in this way, they see a generic error message: `"You do not have the correct access permission."`

In reality, the server is returning a more accurate message: `"Username and password–based login is not enabled at this time."`, but that is being ignored.

## Expected behaviour

Make the login page config-aware:

- Only display Username & Passwords when needed to authenticate (local or active directory)
- Automatically login when OICD based SSO (remove the button)

Always render the errors returned, avoid acronyms (such as OICD)

jescalada:
Fixes #1206.

This PR adds a `/api/auth/config` endpoint to fetch the enabled username/password method, as well as the enabled one-click methods such as OIDC. It also fixes the display for the available config types on the Login page, and improves error handling and visibility of auth methods.

#### When both local* and a custom, non-username/password method

image

#### When only local* is enabled

image

#### When only a custom, non-username/password method is enabled

image

*Defaults to AD if enabled, see `/src/routes/auth` for more details

### DIFF
--- a/src/service/routes/auth.js
+++ b/src/service/routes/auth.js
@@ -66,6 +66,13 @@ const loginSuccessHandler = () => async (req, res) => {
   }
 };
 
+router.get('/config', (req, res) => {
+  res.send({
+    allLoginMethods: getAuthMethods().map((am) => am.type.toLowerCase()),
+    usernamePasswordMethod: getLoginStrategy(),
+  });
+});
+
 // TODO: provide separate auth endpoints for each auth strategy or chain compatibile auth strategies
 // TODO: if providing separate auth methods, inform the frontend so it has relevant UI elements and appropriate client-side behavior
 router.post(
@@ -82,9 +89,9 @@ router.post(
   loginSuccessHandler(),
 );
 
-router.get('/oidc', passport.authenticate(authStrategies['openidconnect'].type));
+router.get('/openidconnect', passport.authenticate(authStrategies['openidconnect'].type));
 
-router.get('/oidc/callback', (req, res, next) => {
+router.get('/openidconnect/callback', (req, res, next) => {
   passport.authenticate(authStrategies['openidconnect'].type, (err, user, info) => {
     if (err) {
       console.error('Authentication error:', err);
--- a/src/ui/views/Login/Login.tsx
+++ b/src/ui/views/Login/Login.tsx
@@ -1,4 +1,4 @@
-import React, { useState, FormEvent } from 'react';
+import React, { useState, FormEvent, useEffect } from 'react';
 import { useNavigate, Navigate } from 'react-router-dom';
 import FormControl from '@material-ui/core/FormControl';
 import InputLabel from '@material-ui/core/InputLabel';
@@ -12,10 +12,10 @@ import CardBody from '../../components/Card/CardBody';
 import CardFooter from '../../components/Card/CardFooter';
 import axios, { AxiosError } from 'axios';
 import logo from '../../assets/img/git-proxy.png';
-import { Badge, CircularProgress, Snackbar } from '@material-ui/core';
-import { getCookie } from '../../utils';
+import { Badge, CircularProgress, FormLabel, Snackbar } from '@material-ui/core';
 import { useAuth } from '../../auth/AuthProvider';
 import { API_BASE } from '../../apiBase';
+import { getAxiosConfig, processAuthError } from '../../services/auth';
 
 interface LoginResponse {
   username: string;
@@ -34,6 +34,19 @@ const Login: React.FC = () => {
   const [success, setSuccess] = useState<boolean>(false);
   const [gitAccountError, setGitAccountError] = useState<boolean>(false);
   const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [authMethods, setAuthMethods] = useState<string[]>([]);
+  const [usernamePasswordMethod, setUsernamePasswordMethod] = useState<string>('');
+
+  useEffect(() => {
+    axios.get(`${API_BASE}/api/auth/config`).then((response) => {
+      const authMethods = response.data.allLoginMethods;
+      const usernamePasswordMethod = response.data.usernamePasswordMethod;
+      const nonUsernamePasswordMethods = authMethods.filter((am) => am !== usernamePasswordMethod);
+
+      setAuthMethods(nonUsernamePasswordMethods);
+      setUsernamePasswordMethod(usernamePasswordMethod);
+    });
+  }, []);
 
   function validateForm(): boolean {
     return (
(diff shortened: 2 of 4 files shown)


src/ui/views/Login/Login.tsx
```diff
@@ -34,33 +34,36 @@ const Login: React.FC = () => {
   const [success, setSuccess] = useState<boolean>(false);
   const [gitAccountError, setGitAccountError] = useState<boolean>(false);
   const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [authMethods, setAuthMethods] = useState<string[]>([]);
+  const [usernamePasswordMethod, setUsernamePasswordMethod] = useState<string>('');
+
+  useEffect(() => {
+    axios.get(`${API_BASE}/api/auth/config`).then((response) => {
```
andypols, line 41: I don’t think the UI needs to be aware of every possible login method. Its concern should just be whether it needs to handle username/password input for the current strategy, or delegate to the strategy’s own UI.

If the UI has to filter or reshape the API response before it can use it, that’s a sign the API might not be the right fit.

jescalada: When I designed/implemented support for multiple login methods, I figured it'd be nice for methods to "co-exist" - as if it were a typical SaaS product that supports whatever method the users have available. That's how I implemented this PR (shows multiple methods). I also like that users need to click to log in - I feel it's more transparent that way.

If you're fairly sure that most organizations using GitProxy would only want to implement *a single* auth method at any given time, then I could implement a more SSO-friendly version (with auto-login).

Alternative: If *only a single auth method* is enabled (say, Okta SSO), then automatically initiate the flow when going to the login page. That way odd organizations with multiple methods can still use the product 😃

jescalada: By the way, the current implementation only shows the `enabled` methods in the UI since potentially multiple strategies are enabled.

jescalada:
@andypols Just fixed this so that it logs in automatically if a single non-user/password method is available. I also fixed up the "${authName} Authentication" label and changed it for "Login", which is only actually shown if a username/password auth method (such as Local or AD) is available. The login page still supports multiple auth methods in the off-chance that some organizations need it.

src/ui/views/Login/Login.tsx
andypols: @jescalada I wasn’t assuming organisations would only ever have one authentication method—it’s just that the schema validation throws an exception if multiple are enabled (because of the oneOf constraint). I ran into this myself when I misconfigured it 🙂

I don’t mind which direction we take—single or multiple methods—I just want to make it consistent. Helps make the code less confusing.

andypols:
Perhaps we should raise a separate issue about single vs. multiple authentication methods? It feels inconsistent that the code and config seem to allow multiple, but the schema prevents it.

The current setup is also creating a poor experience for users:

Seeing a message like “_No username/password authentication method is configured. Please contact an administrator to enable this feature._” is confusing. It makes it look like the organisation is misusing git-proxy rather than just not having that method enabled. It is perfectly valid not to have username/password authentication.

A button that says “LOGIN WITH OPENIDCONNECT” isn’t very user-friendly. Most users won’t know what OpenID Connect means. Even if a button is required, why not just “Login”?

Ideally, SSO methods would simply redirect without needing a button—but that only works if there’s just one method enabled.

@kriswest What do you think?  A second opinion would be welcome :)

src/ui/views/Login/Login.tsx
jescalada: That's odd, on my end the schema validation doesn't seem to error out in this branch 🤔 I'll look into this and then make the UI implementation consistent with that.

andypols: @jescalada - just checked with our deployment - and you are right, it no longer throws an exception (assume this changed as part of migration to quicktype #1140?).  The actual schema still has a `oneOf` constraint added by @kriswest back in June :confused:), its just not getting applied anymore.  🤷🏻‍♂️

kriswest: Added a comment below on this - the schema DOES allow multiple auth methods to be enabled.

andypols: @kriswest thanks for clarifying.  Apologies for misunderstanding

kriswest: np - identified something we can make more intuitive ;-)

kriswest: To cap off this discussion, I'm ok with the new endpoint returning the type names of the enabled auth methods - however @andypols' point that the nonUsernamePasswordMethods element could be produced on the server is valid - perhaps have `usernamePasswordMethods` and `otherMethods` properties in the response?

src/service/routes/auth.js
```diff
@@ -66,6 +66,13 @@ const loginSuccessHandler = () => async (req, res) => {
   }
 };
 
+router.get('/config', (req, res) => {
+  res.send({
+    allLoginMethods: getAuthMethods().map((am) => am.type.toLowerCase()),
+    usernamePasswordMethod: getLoginStrategy(),
+  });
```
kriswest, line 78:
```suggestion
  const usernamePasswordMethod = getLoginStrategy();
  res.send({
    // enabled username /password auth method
    usernamePasswordMethod: usernamePasswordMethod,
    // other enabled auth methods
    otherMethods: getAuthMethods().map((am) => am.type.toLowerCase()).filter((authType) => authType !== usernamePasswordMethod),
  });
```

src/ui/views/Login/Login.tsx
```diff
@@ -34,33 +34,43 @@ const Login: React.FC = () => {
...
+      const nonUsernamePasswordMethods = authMethods.filter(
+        (am: string) => am !== usernamePasswordMethod,
+      );
+
+      setAuthMethods(nonUsernamePasswordMethods);
+      setUsernamePasswordMethod(usernamePasswordMethod);
+
+      // Automatically login if only one non-username/password method is enabled
+      if (!usernamePasswordMethod && nonUsernamePasswordMethods.length === 1) {
+        handleAuthMethodLogin(nonUsernamePasswordMethods[0]);
```
kriswest: To go with the other suggestion

```suggestion
      const usernamePasswordMethod = response.data.usernamePasswordMethod;
      const otherMethods = response.data.otherMethods;

      setUsernamePasswordMethod(usernamePasswordMethod);
      setAuthMethods(otherMethods);
      
      // Automatically login if only one non-username/password method is enabled
      if (!usernamePasswordMethod && otherMethods.length === 1) {
        handleAuthMethodLogin(otherMethods[0]);
```

jescalada:
@andypols The issue with the additional auth methods is that we still need to be able to identify what service we're using to authenticate. I think the best solution would be an extra `name` field in the `proxy.config.json`'s `authentication` list so we can use it for the UI buttons, error messages and such. "Login with OpenIDConnect" could turn into "Google Login" or "Okta Login" depending on the configuration. 🙂 

My interim solution is to display "Login with X" only if multiple methods are present, and "Login" otherwise.

PR #1228, opened 2025-09-26
fix: bug in using API_BASE with URL

andypols:
# Summary

This PR fixes a bug introduced in #1201 related to the URL API.

The `URL` constructor does not support relative paths. When `process.env.VITE_API_URI` is not defined, `API_BASE` falls back to a relative path. As a result, code like this throws an exception in certain environments:

```
const url = new URL(`${API_BASE}/api/auth/gitAccount`);
```

## Fix

* Remove the use of `URL` objects when building url string with `${API_BASE}`.  
  * URLs are now built as plain strings.
* Cleaned up unused query param handling:
  * `/api/v1/user` does not pass any query params.
  * `getUsers` always passed an empty query params object.
  * No params are ever passed into `UserList`

### DIFF
--- a/src/ui/apiBase.ts
+++ b/src/ui/apiBase.ts
@@ -8,4 +8,4 @@ const stripTrailingSlashes = (s: string) => s.replace(/\/+$/, '');
  */
 export const API_BASE = process.env.VITE_API_URI
   ? stripTrailingSlashes(process.env.VITE_API_URI)
-  : '';
+  : location.origin;
--- a/test/ui/apiBase.test.js
+++ b/test/ui/apiBase.test.js
@@ -9,6 +9,14 @@ function loadApiBase() {
 describe('apiBase', () => {
   let originalEnv;
 
+  before(() => {
+    global.location = { origin: 'https://lovely-git-proxy.com' };
+  });
+
+  after(() => {
+    delete global.location;
+  });
+
   beforeEach(() => {
     originalEnv = process.env.VITE_API_URI;
     delete process.env.VITE_API_URI;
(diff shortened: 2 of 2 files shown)


src/service/routes/users.js
```diff
@@ -4,21 +4,8 @@ const db = require('../../db');
...
-    if (k === 'skip') continue;
-    let v = req.query[k];
-    if (v === 'false') v = false;
-    if (v === 'true') v = true;
-    query[k] = v;
-  }
-
-  const users = await db.getUsers(query);
+  console.log(`fetching users`);
+  const users = await db.getUsers({});
```
kriswest, line 8: Some day we need may need to server-side pagination for the various pages, at which point we'll need to handle query params again... On the users API thats a problem I'd like to have, but don't yet and the query arg handling wasn't set up right for it anyway..

src/ui/services/user.ts
```diff
@@ -41,15 +41,14 @@ const getUsers = async (
...
-  const url = new URL(`${API_BASE}/api/v1/user`);
-  url.search = new URLSearchParams(query).toString();
-
   setIsLoading(true);
 
   try {
-    const response: AxiosResponse<UserData[]> = await axios(url.toString(), getAxiosConfig());
+    const response: AxiosResponse<UserData[]> = await axios(
+      `${API_BASE}/api/v1/user`,
+      getAxiosConfig(),
```
kriswest, line 50: I'm worried that if API_BASE is relative here axios still won't be able to resolve it because we're not setting baseUrl in the axiosConfig. which is imported from auth.js:

https://github.com/andypols/git-proxy/blob/455513a64239d9ca0533a1963e66a7b4a3322ec1/src/ui/services/auth.js#L30-L39

In the other services that are still JS they work out baseUrl for themselves and fallback to `location.origin`:

https://github.com/andypols/git-proxy/blob/455513a64239d9ca0533a1963e66a7b4a3322ec1/src/ui/services/repo.js#L4-L6

whjch apiBase.ts does not do...

kriswest (commented):
I'm ok with dropping the unused query args (which may need to come back one day for server-side pagination or filtering). However, I think API_BASE needs to fall back to `location.origin`.

I'll raise an issue for the remaining UI services to be converted to TS - they should also use API_BASE (instead of each working it out themselves), but will need the location.origin fallback (I guess).

PR #1230, opened 2025-09-26
fix: "MongoServerError: The _id cannot be changed" when updating users

andypols:
# Summary

Updating a user (e.g., adding a GitHub username) currently fails with a MongoDB server exception. The issue arises because the user object already includes an _id, and the update attempt tries to overwrite it:

```
MongoServerError: The _id field cannot be changed
```

## Fix

* Removes the `_id` property from the user object before updating.
* Performs an update by `_id` if one exists.
* If no `_id` is present, inserts a new user with a no-op filter (`{ username: user.username }`).

### DIFF
--- a/src/db/mongo/users.ts
+++ b/src/db/mongo/users.ts
@@ -1,4 +1,4 @@
-import { OptionalId, Document } from 'mongodb';
+import { OptionalId, Document, ObjectId } from 'mongodb';
 import { toClass } from '../helper';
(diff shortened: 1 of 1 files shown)


kriswest (approved):
LGTM

Are similar changes needed in any of the other dB clients?

andypols:
@kriswest I did look for other mongo updates, but they all look good.

PR #1233, opened 2025-09-29
chore: update package.json for v2.0.0-rc.3 release

kriswest:


packages/git-proxy-cli/test/testCli.test.js
```diff
@@ -70,7 +70,9 @@ describe('test git-proxy-cli', function () {
     it(`"--version" option prints version details `, async function () {
       const cli = `npx -- @finos/git-proxy-cli --version`;
       const expectedExitCode = 0;
-      const expectedMessages = ['0.1.0'];
+      const packageJson = require('../../../package.json');
+      const version = packageJson.version;
+      const expectedMessages = [version];
```
jescalada, line 75: @kriswest Does this seem like the right solution? This forces us to update the CLI version to match the parent package.

kriswest: Thats what I was going to try, but hadn't thought of checking it against the parent package. If we're keeping the version numbers in sync), and I think we should, then this would be a good guard rail.

PR #1234, opened 2025-09-30
chore: remove defunct release workflow

kriswest:
The release.yml workflow file was replaced with npm.yml a long time ago and its content commented out. Its presence causes a workflow failure email to go out on every PR, hence, it should be removed.


06kellyjac (approved):
I keep seeing this and then forgetting to remove it. Thanks :)

PR #1243, opened 2025-10-08
feat: config schema for missing properties

### LINKED ISSUE #1213: Does anyone use GitHub api baseUrl
The config contains an optional ‘github’ api settings with  `baseUrl`. Does anyone use this for anything?  Looking at refactoring the github api config so we can add api tokens and fork account settings. 

Assume it’s a hangover from something else and not used

### LINKED ISSUE #1237: Update config schema to govern and document config properties not currently captured
The following important config properties are not adequately documented in the Config schema or documentation and TypeScript types generated from it:

https://github.com/finos/git-proxy/blob/292ecadf3794afb253cbd1ee0e8f73a9e00f0028/config.schema.json#L44-L55

All configuration settings should be covered in the schema (and hence generated documentation and typescript types)

kriswest:
resolves #1237
resolves #1213

- Removes the `api.github` config element as it is defunct
- Adds config schema for `commitConfig`, `attestationConfig` and `domains`
- Regenerates the config docs and types from the schema
- Resolves a couple of type errors revealed by the updated types in scanDiff and checkCommitMessages
- Removes the `UserSettings` type as this is no longer used (generated config type `GitProxyConfig` replaced it)


config.schema.json
```diff
@@ -57,16 +44,188 @@
...
+                  "description": "Block commits where the commit diff content matches any of the given regular expressions",
+                  "type": "array",
+                  "items": {}
+                },
+                "providers": {
+                  "title": "MessageBlockProviders",
+                  "description": "Block commits where the commit diff content matches any of the given regular expressions, except where the repository path (project/organisation) matches one of the listed privateOrganisations. The keys in this array are listed as the block type in logs.",
+                  "type": "object",
+                  "additionalProperties": { "type": "string" }
+                }
```
kriswest, line 159: @06kellyjac this is the property that I didn't quite understand - I think I do now... Regex rules that should only be applied to pushes that are not going into orgs in `privateOrganisations` (where some things mmay be allowed that are not allowed externally?). 

I'd be interested to know if its in use and if I have understood it correctly.

proxy.config.json
```diff
@@ -67,11 +67,7 @@
       }
     }
   ],
-  "api": {
-    "github": {
-      "baseUrl": "https://api.github.com"
-    }
-  },
+  "api": {},
```
jescalada, line 70: Should `gitLeaks` from the `api` be present with a default value instead? 🤔

kriswest: I'm wondering that too. I don't see a default value for the `configPath` variable in the defaults in the code:
https://github.com/finos/git-proxy/blob/84d25631e26efdff41443b8b02e9fb8feb0eac0f/src/proxy/processors/push-action/gitleaks.ts#L42-L47

and you can't set an `undefined` value in JSON so I guess it would be:

```JSON
{
  "enabled": false,
  "ignoreGitleaksAllow": true,
  "noColor": false,
}
```

However, I also note the comment:

```
  // adding gitleaks into main git-proxy for now as default off
  // in the future will likely be moved to a plugin where it'll be default on
```

Are we going to move it to a plugin or keep it where it is? That should probably decide if and where it goes into the schema.

If it goes into a plugin, are thre thoughts on configuring plugins? `plugins` is just an array of strings at present - should we be creating something new or is there an existing pattern I missed @coopernetes?

kriswest: I've raised this as a separate issue: #1248
The right answer may depend on where we go with plugins...

config.schema.json
kriswest: resolving so we can merge

PR #1246, opened 2025-10-13
feat: implement hybrid cache architecture

### LINKED ISSUE #1207: Improve push flexibility and pull times
Follow-up issue of #985 (#1189)

Previously, we improved push times by using the `depth: 1` and `singleBranch: true` options in `pullRemote` when cloning. However, this fix doesn't improve pull times (which also happen during pushes) and it locks users into only being able to pull a single branch at a time.

Here are some more details from https://github.com/finos/git-proxy/pull/1189#discussion_r2353440237:

> Note that this effectively means that only the default branch is supported as a diff target. It can be deferred as this speeds up the checkout pretty significantly. However, this does nothing to address overall scalability with giant repos being proxied, pull time and ensuring that the proper diffs are captured by the proxy for presentation in the UI.
>
> As a follow-up to these optimizations, it would be useful to capture the list of tips from a remote repository, persist it in the database and allow for diff's to be "selectable" to a given target branch for proper review. Otherwise, this effectively locks in Git Proxy to supporting one particular branching strategy. Probably fine for most projects but just thought I would raise it here. It can be tracked in an enhancement issue.
>
> Were the proxy to clone all branches, then you can make singleBranch: false while still performing a shallow clone with depth: 1.

fabiovincenzi:
Issue: https://github.com/finos/git-proxy/issues/1207
## Summary

  Two-tier cache system for faster git operations: persistent bare repos + ephemeral working copies.

  ## Architecture

  - `.remote/cache/` - Shared bare repositories (persistent)
  - `.remote/work/<push-id>/` - Per-push working copies (temporary)

  ## Configuration

  Add to `proxy.config.json`:

  ```json
  {
    "cache": {
      "maxSizeGB": 2,
      "maxRepositories": 50,
      "cacheDir": "./.remote/cache"
    }
  }
  ```
   ## Security

  - Each push gets unique working copy (.remote/work/<push-id>/)
  - Bare cache contains only git objects (no user files, no credentials)
  - sanitizeRepositoryConfig() immediately removes credentials from git config
  - Working copies deleted after push completes


src/config/index.ts
```diff
@@ -285,6 +289,17 @@ export const getRateLimit = () => {
...
 };
 
+export const getCacheConfig = () => {
+  const config = loadFullConfiguration();
+  return (
+    config.cache || {
+      maxSizeGB: 2,
+      maxRepositories: 50,
+      cacheDir: './.remote/cache',
+    }
```
06kellyjac: How does it work if in my config I only set `maxSizeGB`?
Do the other values end up unset?
Do we need to do an object merge to have defaults appear or is that handled further up in the config?

src/proxy/processors/push-action/clearBareClone.ts
```diff
@@ -1,16 +1,41 @@
...
+      if (fs.existsSync('./.remote')) {
+        fs.rmSync('./.remote', { recursive: true, force: true });
+        step.log('Test environment: Full .remote directory cleaned');
+      } else {
+        step.log('Test environment: .remote directory already clean');
+      }
+    } catch (err) {
+      step.log(`Warning: Could not clean .remote directory: ${err}`);
+    }
+  } else {
```
06kellyjac: I feel like a safer option that ensures what we're testing more closely matches prod would be to either do some separate cleanup as necessary in the tests, or adjust config to have a much lower max (or even 0) then we can see if CacheManager itself is doing the cleanup.

src/proxy/processors/push-action/git-operations.ts
```diff
@@ -0,0 +1,156 @@
...
+  if (singleBranch) {
+    args.push('--single-branch');
+  } else {
+    // Explicitly clone all branches (needed when using --depth)
+    args.push('--no-single-branch');
+  }
+
+  args.push(`"${authUrl}"`, `"${dir}"`);
+
+  execSync(args.join(' '), { stdio: 'pipe' });
```
06kellyjac: can this be adjusted to use `spawnSync` instead please

https://docs.datadoghq.com/security/code_security/static_analysis/static_analysis_rules/javascript-node-security/detect-child-process/

Should just be a case of feeding it the args & removing "git" from args to be the string `cmd`

src/proxy/processors/push-action/git-operations.ts
```diff
@@ -0,0 +1,156 @@
...
+  const { dir, url, username, password, depth, prune = false, bare = false } = options;
+
+  // Build URL with credentials if provided
+  let authUrl = url;
+  if (username && password) {
+    authUrl = url.replace(
+      /^(https?:\/\/)/,
+      `$1${encodeURIComponent(username)}:${encodeURIComponent(password)}@`,
+    );
+  }
```
06kellyjac: This code is repeated in `clone` above

src/proxy/processors/push-action/git-operations.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  if (prune) {
+    args.push('--prune');
+  }
+
+  args.push(`"${authUrl}"`);
+  args.push('"+refs/heads/*:refs/heads/*"'); // Fetch all branches
+
+  execSync(args.join(' '), { stdio: 'pipe' });
```
06kellyjac: same as `execSync` above

src/proxy/processors/push-action/git-operations.ts
```diff
@@ -0,0 +1,156 @@
...
+      // For non-bare repositories, remote origin should exist
+      try {
+        // Unset the URL with credentials
+        execSync(`git -C "${dir}" config --unset remote.origin.url`, { stdio: 'pipe' });
+      } catch (e) {
+        // Ignore error if already unset
+      }
+
+      // Set clean URL without credentials
+      execSync(`git -C "${dir}" remote set-url origin "${cleanUrl}"`, { stdio: 'pipe' });
```
06kellyjac: same as `execSync` above

src/proxy/processors/push-action/git-operations.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  const args: string[] = ['git', 'clone'];
+
+  if (depth) {
+    args.push('--depth', depth.toString());
+  }
+
+  args.push(`"${sourceDir}"`, `"${targetDir}"`);
+
+  execSync(args.join(' '), { stdio: 'pipe' });
```
06kellyjac: same as `execSync` above

src/proxy/processors/push-action/metrics.ts
```diff
@@ -0,0 +1,37 @@
...
+  private startTime: number = 0;
+  private operation: string = '';
+
+  constructor(step: Step) {
+    this.step = step;
+  }
+
+  start(operation: string): void {
+    this.operation = operation;
+    this.startTime = Date.now();
```
06kellyjac: Can this class be adjusted to use `performance.now()` for safer reporting

https://developer.mozilla.org/en-US/docs/Web/API/Performance/now
https://nodejs.org/api/perf_hooks.html#performancenow
https://stackoverflow.com/questions/111368/how-do-you-performance-test-javascript-code

src/proxy/processors/push-action/pullRemote.ts
```diff
@@ -1,45 +1,137 @@
 import { Action, Step } from '../../actions';
 import fs from 'fs';
-import git from 'isomorphic-git';
-import gitHttpClient from 'isomorphic-git/http/node';
+import { PerformanceTimer } from './metrics';
+import { cacheManager } from './cache-manager';
+import * as gitOps from './git-operations';
 
-const dir = './.remote';
+const BARE_CACHE = './.remote/cache';
```
06kellyjac: The location of the cache can be defined in the config but hardcoded here?

src/config/index.ts
06kellyjac: How does it work if in my config I only set `maxSizeGB`?
Do the other values end up unset?
Do we need to do an object merge to have defaults appear or is that handled further up in the config?

06kellyjac:
```suggestion
    { ...{
      maxSizeGB: 2,
      maxRepositories: 50,
      cacheDir: './.remote/cache',
    }, ...config.cache }
```

**(not formatted)**

src/proxy/processors/push-action/cache-manager.ts
```diff
@@ -0,0 +1,181 @@
...
+   */
+  enforceLimits(): { removedRepos: string[]; freedMB: number } {
+    const stats = this.getCacheStats();
+    const removedRepos: string[] = [];
+    let freedMB = 0;
+
+    // Sort repositories by last accessed (oldest first for removal)
+    const reposToEvaluate = [...stats.repositories].sort(
+      (a, b) => a.lastAccessed.getTime() - b.lastAccessed.getTime(),
+    );
```
06kellyjac: Nice

06kellyjac: I can't remember if we can use `toSorted` yet. If we still cant a comment to use `toSorted` in the future would help to find this later :)

src/proxy/processors/push-action/cache-manager.ts
```diff
@@ -0,0 +1,181 @@
+import fs from 'fs';
+import path from 'path';
+import { getCacheConfig } from '../../../config';
+
+export interface CacheStats {
+  totalRepositories: number;
+  totalSizeMB: number;
```
06kellyjac: We have a bit of a mix of units: bytes from node `statSync`, MB here, GB for the config.

IMO it'd be safest to live by bytes everywhere. Or maybe keep GB for the settings but use bytes internally for consistency.

src/proxy/processors/push-action/cache-manager.ts
```diff
@@ -0,0 +1,181 @@
...
+      }
+    };
+
+    try {
+      calculateSize(dirPath);
+    } catch (error) {
+      return 0;
+    }
+
+    return Math.round(totalBytes / (1024 * 1024)); // Convert to MB
```
06kellyjac:
```suggestion
    return Math.ceil(totalBytes / (1024 * 1024)); // Convert to MB
```

rounding down would hide some amount of KB up to 511KB. Not the end of the world but safest to always round up rather than be shortchanged.

06kellyjac (commented):
I'm looking forward to faster pulls!

Caches are a common place for logic or security problems so we may want added scrutiny and testing for this PR to be extra safe :)

Added some thoughts, not ran this locally just yet.

src/proxy/processors/push-action/cache-manager.ts
```diff
@@ -0,0 +1,181 @@
...
+  totalSizeMB: number;
+  repositories: Array<{
+    name: string;
+    sizeMB: number;
+    lastAccessed: Date;
+  }>;
+}
+
+export class CacheManager {
+  private cacheDir: string;
```
jescalada: Is this `cacheDir` related to the `cachedir` library in `package.json` or the `cacheDir` used in the `ConfigLoader`? If not, we might want to pick a different name to disambiguate them.

src/proxy/processors/push-action/cache-manager.ts
jescalada, line 1: I'm wondering what would happen if two pulls happen simultaneously and execute `touchRepository` or `enforceLimits` at the same time - wouldn't this cause some kind of race condition? Especially with cache statistics in the mix, we wouldn't be able to accurately pick up the latest accessed repository.

src/config/index.ts
fabiovincenzi: The defaults are already properly merged in `mergeConfigurations` (lines 106-108):
 ```typescript
  cache: userSettings.cache
    ? { ...defaultConfig.cache, ...userSettings.cache }
    : defaultConfig.cache,
  ```
The || fallback is actually redundant, it would only trigger if config.cache is null/undefined after the merge, which shouldn't happen.

Removing the fallback and added a test for this in test/testConfig.test.js

src/proxy/processors/push-action/cache-manager.ts
fabiovincenzi: Right, using bytes everywhere except for the config

src/proxy/processors/push-action/cache-manager.ts
```diff
@@ -0,0 +1,181 @@
...
+  /**
+   * Enforce cache limits using LRU eviction
+   */
+  enforceLimits(): { removedRepos: string[]; freedMB: number } {
+    const stats = this.getCacheStats();
+    const removedRepos: string[] = [];
+    let freedMB = 0;
+
+    // Sort repositories by last accessed (oldest first for removal)
+    const reposToEvaluate = [...stats.repositories].sort(
```
jescalada: Is there a reason to sort the repositories twice? The first sort happens inside the `getCacheStats()` call right above.

src/proxy/processors/push-action/cache-manager.ts
```diff
@@ -0,0 +1,181 @@
...
+    const repoPath = path.join(this.cacheDir, repoName);
+    if (fs.existsSync(repoPath)) {
+      fs.rmSync(repoPath, { recursive: true, force: true });
+    }
+  }
+
+  /**
+   * Calculate directory size in MB
+   */
+  private getDirectorySize(dirPath: string): number {
```
jescalada, line 169: Should the error clauses here be logging something instead of failing silently? Such as when there's an error on `calculateSize(dirPath)`.

src/proxy/processors/push-action/cache-manager.ts
fabiovincenzi: I think it should be available from node 20 so yes I'll use it

src/proxy/processors/push-action/cache-manager.ts
fabiovincenzi: There's no conversion anymore since we are using bytes everywhere

src/proxy/processors/push-action/pullRemote.ts
jescalada, line 1: Would be nice to have a short README on how the `pullRemote` works and what the hybrid cache architecture is doing. This'll make things much clearer for future development and finding potential flaws in our logic 🙂

jescalada (commented):
Just a code review for now - I'll be testing the code later!

I was wondering if you could do some profiling to verify/document the speed improvements with this PR. Would be nice to try it out with massive repos (for example https://github.com/backstage/backstage which is >10GB) to see the difference.

I'll try to do some profiling soon as well 🙂

src/proxy/processors/push-action/cache-manager.ts
fabiovincenzi: Good catch! I did some testing to reproduce the race conditions but it's challenging to simulate them (concurrent
  pushes need to overlap during the brief enforceLimits() window).

  However, I agree they can potentially occur. 
  Some race conditions wouldn't be so bad:
  - Duplicate eviction attempts → at worst just a warning log
  - Repository removed between cache check and fetch → falls back to clone instead of fetch

  But others are more concerning:
  - getCacheStats() reading the cache directory while enforceLimits() is actively deleting repositories → could call fs.statSync() on a path that was just removed, causing an unhandled exception and crashing the push request
  - Multiple concurrent enforceLimits() calls reading the same stale cache stats → both might decide to evict the same repositories, leading to inconsistent state

  To address this, I've implemented a Promise-based mutex in CacheManager that serializes touchRepository() and enforceLimits(), ensuring only one operation modifies the cache at a time.

src/proxy/processors/push-action/cache-manager.ts
fabiovincenzi: You are right

src/proxy/processors/push-action/pullRemote.ts
fabiovincenzi: Added comprehensive README documenting the hybrid cache architecture. The README includes a benchmarks section showing performance improvements and instructions to reproduce the results using the included `scripts/cache-benchmark.sh` script.
  See: `src/proxy/processors/push-action/README.md`

src/proxy/processors/push-action/README.md
```diff
@@ -0,0 +1,200 @@
...
+- **Test Repository**: Backstage (medium-large repository, 177MB cached)
+- **Test Method**: 10 consecutive push operations (1 cold + 9 warm)
+- **Cache Configuration**: Bare repositories with `depth: 1` (shallow clone)
+- **Benchmark Script**: [`cache-benchmark.sh`](../../../../scripts/cache-benchmark.sh)
+
+### Results Comparison
+
+| Metric              | Without Cache (main) | With Cache (PR) | Improvement          |
+| ------------------- | -------------------- | --------------- | -------------------- |
+| **Cold Push**       | 20.63s               | 17.58s          | 15% faster           |
```
jescalada: I think the `Improvement` column calculations might be incorrect 🤔 A push that took 20s and is now taking 6.68s is 20/6.68 = 2.99x faster.

I think it feels more impactful and easier to understand when formatting it as a multiplier (3x faster, etc.) rather than a percentage.

benchmark-cache.sh
```diff
@@ -0,0 +1,205 @@
...
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
+NC='\033[0m' # No Color
+
+echo -e "${BLUE}=== Git Proxy Hybrid Cache Benchmark ===${NC}"
+echo ""
+
+# Configuration
+PROXY_URL="http://localhost:8000"
+GITHUB_REPO="${1:-fabiovincenzi/open-webui}"
```
jescalada: Nitpick: We might want to add an explicit error message when executing `benchmark-cache.sh` but the chosen repo is not added to the authorised list.

jescalada: Perhaps we could also throw an error if the user didn't provide their own repo in the arguments rather than defaulting to your fork?

src/proxy/processors/push-action/pullRemote.ts
jescalada: LGTM!

benchmark-cache.sh
```diff
@@ -0,0 +1,205 @@
...
+    cd ..
+
+    echo -e "${GREEN}Results:${NC}"
+    echo "  Clone time: ${CLONE_TIME}s"
+    echo "  Push time:  ${PUSH_TIME}s"
+    echo "  Total time: ${TOTAL_TIME}s"
+    echo ""
+
+    # Store results
+    echo "$push_number,$is_first,$CLONE_TIME,$PUSH_TIME,$TOTAL_TIME" >> results.csv
```
jescalada: Maybe we should add the resulting `.csv`s to `.gitignore`? 🤔

benchmark-cache.sh
```diff
@@ -0,0 +1,205 @@
...
+
+REPO_NAME=$(basename "$GITHUB_REPO")
+
+# Clear cache before starting
+echo -e "${YELLOW}Clearing cache before benchmark...${NC}"
+rm -rf ../.remote/cache/* ../.remote/work/* 2>/dev/null || true
+echo -e "${GREEN}✓ Cache cleared${NC}"
+echo ""
+
+measure_push() {
```
jescalada: Wonder if we could improve error handling here to prevent the script from hanging? I tried pulling my backstage fork but it strangely stopped responding on the 3rd push:

Image

benchmark-cache.sh
```diff
@@ -0,0 +1,205 @@
...
+    local is_first=$2
+
+    echo -e "${BLUE}=== Push #$push_number $([ "$is_first" = "true" ] && echo "(COLD CACHE)" || echo "(WARM CACHE)") ===${NC}"
+
+    # Clone repo through proxy
+    echo "Cloning repository..."
+    START_CLONE=$(date +%s.%N)
+
+    rm -rf "$REPO_NAME" 2>/dev/null || true
+    git clone "$PROXY_REPO_URL" "$REPO_NAME" > clone.log 2>&1
```
jescalada: Perhaps we could output the clone status just to make sure things are going well for larger repos too?

```suggestion
    git clone "$PROXY_REPO_URL" "$REPO_NAME" 2>&1 | tee clone.log
```

jescalada (commented):
Did some manual testing and things seem to work great! When using the profiling script though, I had some issues with `git clone` hanging without any error messages on both the backend and the script execution.

I tried testing with Backstage which is massive and so hiccups like that are expected. However, we do want to make sure that even these large repos can get cloned correctly without fail, or at least give descriptive errors on fail.

src/proxy/processors/push-action/README.md
fabiovincenzi: You're right! The percentages were showing time reduction (66% of time saved) rather than the speedup multiplier. Updated to show it as a multiplier, which is much clearer and more impactful. Thanks!

benchmark-cache.sh
fabiovincenzi: Just edited to show original errors everywhere

fabiovincenzi: Oh, I tought i removed all the references of my github, I just realized I didn't commit those

benchmark-cache.sh
fabiovincenzi: I think now the error handling shoud be much better, I'm sorry I didn't push latest changes

coopernetes:
This PR has been opened awhile and has a much-needed optimization for local checkouts. Marking this for review to include 2.2.0 since 2.1.0 is already committed and almost done.

fabiovincenzi:
Merged latest main. Since this PR was opened, pullRemote has been split into PullRemoteBase / PullRemoteHTTPS / PullRemoteSSH, so the cache now plugs into those classes instead of replacing pullRemote.
 
Three things for review:

- the cache is now optional via `cache.enabled`, with it off, the factory returns the existing implementations untouched
- it works over SSH as well as HTTPS, sharing the same mirror
- the cache key now includes the project, previously collided on a single mirror and fetched each other's history into it

Verified end-to-end on both protocols (cache MISS then HIT).

PR #1017, opened 2025-05-20
Add Apache-2.0 license headers to source files

sam-holmes2:

PR #1253, opened 2025-10-21
fix: Add Apache-2.0 license headers to source files (2)

sam-holmes2:
Cherry picked branch from https://github.com/RBC/finos-git-proxy/tree/chore/eslint-license to hopefully avoid conflicts as requested. 

Changes add ESLint rule with auto-fixable license headers added to source files (more info in original PR #1017)


package.json
coopernetes, line 1: @sam-holmes2 I believe these changes, other than the addition of `eslint-plugin-license-header` (L112), should be reverted or merged from main. Mind tidying this up?

PR #1257, opened 2025-10-25
refactor: remove unnecessary getUserLoggedIn API call in User component

andypols:
# Summary

This PR simplifies and improves the logic of the `User.tsx` view.

Previously, the component was making an unnecessary API call to `getUserLoggedIn` to check if the current user is an admin — information that’s already available through the `UserContext` provided by the Dashboard layout. Other components (e.g., `RepoDetails.tsx `and `Repositories.tsx`) already rely on this context, making the extra call redundant.

Additionally, the component contained several code quality issues:

- Redundant `isAdmin` state variable
- Duplicate logic within the `useEffect` branches
- Unnecessary state for derived values (e.g., `isProfile`)

## Fix

1. Use `UserContext` for admin checks: _Replaced the `getUserLoggedIn` API call with context access to determine admin status._
2. Remove redundant API call and state: _Eliminated the `getUserLoggedIn` call and the `isAdmin` state variable._
3. Simplify derived state: _Replaced the `isProfile` state with a derived constant: `const isProfile = !id;`_
4. Refactor useEffect logic: _Consolidated duplicate if/else branches into a single `getUser` call for cleaner, more maintainable code._

### DIFF
--- a/src/ui/views/User/UserProfile.tsx
+++ b/src/ui/views/User/UserProfile.tsx
@@ -37,7 +37,7 @@ export default function UserProfile(): React.ReactElement {
   const navigate = useNavigate();
   const { id } = useParams<{ id?: string }>();
   const { user: loggedInUser } = useContext<UserContextType>(UserContext);
-  const isProfile = !id;
+  const isOwnProfile = !id;
 
   useEffect(() => {
     getUser(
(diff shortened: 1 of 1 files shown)


andypols:
@jescalada thanks for looking at this - I have:

* renamed file to `UserProfile` to match the component
* and renamed `isProfile` to `isOwnProfile` - its the difference when you are looking at the your profile (My Account) vs looking at a user in the list.

Issue #1149, opened 2025-08-13
Improve plugin loading functionality


jescalada:
**Is your feature request related to a problem? Please describe.**
After the TS + ESM refactor PRs, we've been having all kinds of issues with the `plugin-loader` library. The runtime errors were fixed by using `tsx` for running the app (`"server": "tsx index.ts",`), and the test errors were fixed by using `ts-mocha` to run the tests. However, there are still issues when loading plugins that are written in ESM: #1088 

Based on the discussion there, we should probably look for a way to remove the `load-plugin` library entirely and replace with something without TS/ESM compatibility issues.

jescalada:
Hi @coopernetes, just wanted to confirm whether the latest `PluginLoader` test fix also fixes this issue!

PR #1262, opened 2025-10-29
fix: Plugin loader tests

### LINKED ISSUE #1263: Fix failing plugin loader tests
Ever since `src/plugin` got refactored into TS, the plugin loader tests (`test/plugin/plugin.test.js`) started failing. Those tests were skipped as a temporary measure, but need to be fixed along with actual support for CommonJS plugins.

### Related issues
See the discussions in #1088 and #1149 for more info on the failing plugins and potential fixes.

### Test output
Image

jescalada:
Fixes #1263.

Related: #1149

### DIFF
--- a/src/proxy/processors/pre-processor/parseAction.ts
+++ b/src/proxy/processors/pre-processor/parseAction.ts
@@ -9,24 +9,21 @@ const exec = async (req: {
 }) => {
   const id = Date.now();
   const timestamp = id;
-  const pathBreakdown = processUrlPath(req.originalUrl);
   let type = 'default';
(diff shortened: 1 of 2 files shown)


jescalada:
@coopernetes Thanks for the review! As you mentioned, the TS config plus proper package.json setup seems to fix that side. I'm not sure if this actually fixes the issue that you posted originally. I'd be super grateful if you could check this on your setup as it seems to be (mostly?) working on my side. I experimented with your patch and didn't notice any differences when adding the changes in this PR 🤔 

I'll go ahead and merge this to update the Vitest refactor PR!

PR #1266, opened 2025-10-31
test: add tests for plugins written in ESM

jescalada:
Follow-up PR for #1262.

### DIFF
--- a/test/fixtures/test-package/esm-multiple-export.js
+++ b/test/fixtures/test-package/esm-multiple-export.js
@@ -11,8 +11,8 @@ export default {
     return action;
   }),
   baz: {
-    exec: (async (req, action) => {
+    exec: async (req, action) => {
       console.log('not a real plugin object');
-    }),
+    },
   },
 };


test/fixtures/test-package/esm-multiple-export.js
```diff
@@ -0,0 +1,13 @@
...
+// test multiple exports (ESM syntax)
+export default {
+  foo: new PushActionPlugin(async (req, action) => {
+    console.log('PushActionPlugin: ', action);
+    return action;
+  }),
+  bar: new PullActionPlugin(async (req, action) => {
+    console.log('PullActionPlugin: ', action);
+    return action;
+  }),
```
coopernetes, line 12: Maybe add a dummy object here to validate the load is only picking up proper subclasses of the Push/PullActionPlugin class?
```suggestion
  }),
  baz: {
    exec: (async (req, action) => {
      console.log('not a real plugin object');
    }),
  },
```

coopernetes (approved):
LGTM. One suggested test case for ensuring the module loaded is using the correct objects and ignoring "invalid" ones (that may be unrelated to plugins such as exported functions for unit testing, etc).

test/fixtures/test-package/esm-multiple-export.js
jescalada: The previous tests are implicitly checking that no other objects are being loaded here (I just added comments for clarity):

```ts
expect(loader.pushPlugins.length).to.equal(1);
expect(loader.pullPlugins.length).to.equal(1);
```

PR #1267, opened 2025-10-31
chore: add confirmation dialog

andypols:
# Summary

This PR adds a delete confirmation dialog to prevent accidental repository deletions.

## Background

I unintentionally deleted a repository after clicking “Delete” while moving the mouse toward the “Code” button. To avoid similar mistakes, this PR introduces a confirmation step before deletion.

## Fix

- Adds a confirmation dialog when deleting a repository.
- Requires the user to type (or paste) the repository name to confirm the action.
- Follows the standard UI convention used for similar destructive actions.

CleanShot 2025-10-31 at 14 54 38@2x

The delete button is enabled when the correct repo name has been added.

CleanShot 2025-10-31 at 14 55 29@2x

## Note

There’s a quirk where you can’t remove the default repository. Right after a repo is deleted, the handler restarts the proxy, and during that restart, proxyPreparations() runs again — which re-inserts every repo defined in `proxy.config.json`. If the repo you deleted is listed there, it’s immediately recreated, so the UI still shows it even after a refresh.

### DIFF
--- a/src/ui/views/RepoDetails/Components/DeleteRepoDialog.tsx
+++ b/src/ui/views/RepoDetails/Components/DeleteRepoDialog.tsx
@@ -12,7 +12,7 @@ interface ConfirmDeleteRepoProps {
   onConfirm: () => void;
 }
 
-const ConfirmDeleteRepo: React.FC<ConfirmDeleteRepoProps> = ({
+const DeleteRepoDialog: React.FC<ConfirmDeleteRepoProps> = ({
   repoName,
   open,
   onClose,
@@ -67,4 +67,4 @@ const ConfirmDeleteRepo: React.FC<ConfirmDeleteRepoProps> = ({
   );
 };
 
-export default ConfirmDeleteRepo;
+export default DeleteRepoDialog;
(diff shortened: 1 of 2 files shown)


src/ui/views/RepoDetails/Components/ConfirmDeleteRepo.tsx
```diff
@@ -0,0 +1,70 @@
...
+import DialogActions from '@material-ui/core/DialogActions';
+
+interface ConfirmDeleteRepoProps {
+  repoName: string;
+  open: boolean;
+  onClose: () => void;
+  onConfirm: () => void;
+}
+
+const ConfirmDeleteRepo: React.FC<ConfirmDeleteRepoProps> = ({
```
jescalada: Nit: `ConfirmDeleteRepo` (and related occurrences) could be replaced with something like `DeleteRepoDialog` for clarity + searchability.

andypols: Done

PR #1293, opened 2025-11-28
fix: misleading endpoint status codes and error messages

### LINKED ISSUE #1285: Improve endpoint status codes (`/src/service/routes`)
Many endpoints have misleading error codes, such as the push routes in `src/service/routes/push.ts`:

```ts
router.post('/:id/reject', async (req: Request, res: Response) => {
  if (!req.user) {
    res.status(401).send({
      message: 'not logged in',
    });
    return;
  }

  const id = req.params.id;
  const { username } = req.user as { username: string };

  // Get the push request
  if (!push) return;

  const list = await db.getUsers({ email: committerEmail });

  if (list.length === 0) {
    res.status(401).send({
      message: `There was no registered user with the committer's email address: ${committerEmail}`,
    });
    return;
  }

  if (list[0].username.toLowerCase() === username.toLowerCase() && !list[0].admin) {
    res.status(401).send({
      message: `Cannot reject your own changes`,
    });
    return;
  }

  const isAllowed = await db.canUserApproveRejectPush(id, username);

  if (isAllowed) {
    const result = await db.reject(id, null);
    console.log(`user ${username} rejected push request for ${id}`);
    res.send(result);
  } else {
    res.status(401).send({
      message: 'User is not authorised to reject changes',
    });
  }
});
```

In this case, we're sending 401 responses in cases where the user is already authenticated. We should be sending 403 which shows that the user lacks the correct permissions.

Replace misleading error codes (especially 401 errors) with more suitable ones. Also, adjust tests to reflect these changes.

**Additional context**
https://github.com/finos/git-proxy/pull/1202#discussion_r2514169609
https://github.com/finos/git-proxy/pull/1202#discussion_r2514170916
https://github.com/finos/git-proxy/pull/1202#discussion_r2514171830
https://github.com/finos/git-proxy/pull/1202#discussion_r2514174772

jescalada:
Fixes #1285. Based on #1202, would love to get that one merged first!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
 
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@95d9a5deda9de15063e7595e9719c11c38c90ae2 # v2.13.2
+        uses: step-security/harden-runner@df199fb7be9f65074067a9eb93f12bb4c5547cf2 # v2.13.3
         with:
           egress-policy: audit
 
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -51,7 +51,7 @@ jobs:
 
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@95d9a5deda9de15063e7595e9719c11c38c90ae2 # v2
+        uses: step-security/harden-runner@df199fb7be9f65074067a9eb93f12bb4c5547cf2 # v2
         with:
           egress-policy: audit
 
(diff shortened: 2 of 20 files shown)


src/ui/views/User/UserProfile.tsx
```diff
@@ -46,31 +46,27 @@ export default function UserProfile(): React.ReactElement {
...
-      setUser(updatedData);
-      navigate(`/dashboard/profile`);
-    } catch {
-      setIsError(true);
-    }
+    const updatedData = {
+      ...user,
+      gitAccount: escapeHTML(gitAccount),
+    };
+    await updateUser(updatedData, setErrorMessage, setIsLoading);
```
kriswest, line 68: does not reject anymore so no catch needed 👍 
Perhaps worth a comment.

```suggestion
    //does not reject and will display any errors that occur
    await updateUser(updatedData, setErrorMessage, setIsLoading);
```

kriswest (approved):
LGTM!

Issue #1030, opened 2025-05-23
Reviewers must provide reason for rejection


sam-holmes2:
**Is your feature request related to a problem? Please describe.**
Currently users can reject PRs without providing a reason 

**Describe the solution you'd like**
GitProxy UI should display an error message to reviewers unless a reason for rejecting a PR is supplied

kriswest:
It'd be useful to receive email notifications of rejection/approval/need for review etc.

jescalada:
Duplicate of #605, solved in #1396.

Issue #1188, opened 2025-09-04
Improve UI error handling and API error codes


jescalada:
**Is your feature request related to a problem? Please describe.**
Right now, the UI handles errors rather inflexibly. Usually, we check for a single error code, and then send out a generic error message based on the code.

This is an example from `/src/ui/services/git-push.js`:

```js
const rejectPush = async (id, setMessage, setUserAllowedToReject) => {
  const url = `${baseUrl}/push/${id}/reject`;
  let errorMsg = '';
  let isUserAllowedToReject = true;
  await axios.post(url, {}, getAxiosConfig()).catch((error) => {
    if (error.response && error.response.status === 401) {
      errorMsg = 'You are not authorised to reject...';
      isUserAllowedToReject = false;
    }
  });
  await setMessage(errorMsg);
  await setUserAllowedToReject(isUserAllowedToReject);
};
```

This is the related API endpoint (`src/service/routes/push.ts`):

```ts
router.post('/:id/reject', async (req: Request, res: Response) => {
  if (req.user) {
    const id = req.params.id;

    // Get the push request
    const push = await getValidPushOrRespond(id, res);
    if (!push) return;

    // Get the committer of the push via their email
    const committerEmail = push.userEmail;
    const list = await db.getUsers({ email: committerEmail });

    if (list.length === 0) {
      res.status(401).send({
        message: `There was no registered user with the committer's email address: ${committerEmail}`,
      });
      return;
    }

    if (
      list[0].username.toLowerCase() === (req.user as any).username.toLowerCase() &&
      !list[0].admin
    ) {
      res.status(401).send({
        message: `Cannot reject your own changes`,
      });
      return;
    }

    const isAllowed = await db.canUserApproveRejectPush(id, (req.user as any).username);
    console.log({ isAllowed });

    if (isAllowed) {
      const result = await db.reject(id, null);
      console.log(`user ${(req.user as any).username} rejected push request for ${id}`);
      res.send(result);
    } else {
      res.status(401).send({
        message: 'User is not authorised to reject changes',
      });
    }
  } else {
    res.status(401).send({
      message: 'not logged in',
    });
  }
});
```

Notice how all errors must be set to 401 in order to get an error message to display on the frontend. On top of that, the actual reason is not passed to the frontend.

**Describe the solution you'd like**
We could scope this issue down to these core tasks:
- Fixing up the UI error display so that **any** API error is correctly displayed (popup/toast)
[...]

PR #1295, opened 2025-11-29
feat: Add reject reason

andypols:
# Summary

This PR addresses #1030. Previously, push rejections were recorded without any context. Reviewers now must provide an explanation when rejecting a push, and that explanation is stored and displayed so contributors can understand why their changes were not approved.

## Solution

**Backend**
* Database
    * Updated `db.reject` (file and Mongo) to accept a rejection object.
    * Persist the rejection on the `Action` via a new `rejection` property.
* Push API
    * Updated POST `/push/:id/reject` to require a rejection reason.  
    * Fetches the reviewer email (similar to approvals) and blocks the rejection if it’s missing.
    * Saves the rejection payload including:
       * reason
       * timestamp
       * reviewer

**UI**
* Refactored `PushDetails` into smaller components for clarity and easier maintenance:
   * `PushStatusHeader`
   * `StatusIcon`
   * `Attestation`
   * `Rejection`
   * `ApprovalBadge`
   * `RejectionBadge`
* Added a **Rejection dialog** that requires the reviewer to enter a reason before rejecting a push.
* Added `RejectionBadge` and `ApprovalBadge` components to display:
   * reviewer
   * timestamp
   * rejection details (for rejections)

## Screenshot

Reject button is disabled by default.

CleanShot 2025-12-05 at 12 20 53@2x

Button enabled with a reason is given.

CleanShot 2025-12-05 at 12 25 17@2x

CleanShot 2025-12-05 at 15 43 48@2x

### DIFF
--- a/src/ui/views/PushDetails/components/PushStatusHeader.tsx
+++ b/src/ui/views/PushDetails/components/PushStatusHeader.tsx
@@ -42,7 +42,7 @@ const PushStatusHeader: React.FC<PushStatusHeaderProps> = ({
 
   return (
     <CardHeader color={headerData.color} stats icon>
-      <StatusIcon status={headerData.title} color={headerData.title} />
+      <StatusIcon status={headerData.title} color={headerData.color} />
       {isPending && (
         <div style={{ display: 'inline-flex', padding: '20px' }}>
           <Button color='warning' onClick={onCancel}>


src/db/mongo/pushes.ts
```diff
@@ -63,7 +63,7 @@ export const writeAudit = async (action: Action): Promise<void> => {
   await collection.updateOne({ id: data.id }, { $set: data }, options);
 };
 
-export const authorise = async (id: string, attestation: any): Promise<{ message: string }> => {
+export const authorise = async (id: string, rejection: any): Promise<{ message: string }> => {
```
jescalada, line 66: ~~Same as above, except I'm wondering if it makes sense to include the `rejection` string here - since we're authorising (and the authorisation "reason" would be implicit in the `attestation.questions`).~~

Would be great to have an `Approval` object here instead (formerly `Attestation` type)

src/db/file/pushes.ts
```diff
@@ -112,7 +112,7 @@ export const authorise = async (id: string, attestation: any): Promise<{ message
   return { message: `authorised ${id}` };
 };
 
-export const reject = async (id: string, attestation: any): Promise<{ message: string }> => {
+export const reject = async (id: string, rejection: any): Promise<{ message: string }> => {
```
jescalada, line 115: Might want to add the `Rejection` type here!

src/proxy/processors/types.ts
```diff
@@ -19,6 +19,14 @@ export type Attestation = {
   questions: Question[];
 };
 
+export type Rejection = {
+  reviewer: {
+    username: string;
+  };
+  timestamp: string | Date;
+  reason: string;
+};
```
jescalada, line 28: I understand a bit better now - perhaps we could have a `Review` base object, rename `Attestation` to `Approval` and keep `Rejection`? Then the database function and API routes only need one of them.

src/ui/services/git-push.ts
```diff
@@ -30,86 +30,54 @@ const getPush = async (
 };
 
 const getPushes = async (
-  setIsLoading: (isLoading: boolean) => void,
-  setPushes: (pushes: PushActionView[]) => void,
-  setAuth: (auth: boolean) => void,
-  setIsError: (isError: boolean) => void,
-  setErrorMessage: (errorMessage: string) => void,
```
jescalada, line 37: Wondering if it's okay to remove the error handling here - I know it's not the cleanest right now, but it'd be best to refactor all the `src/ui/services` systematically and in a single PR.

I'm inclined to keep the error handling for now until we start overhauling the whole `src/ui/services` module into something cleaner.

There's an issue to track this: #1188

andypols: I'm not really removing error handling, It's more removing unused code.  The error/auth value are never used. Add back code like this really hurts :)

```
  const [, setIsError] = useState(false);
  const navigate = useNavigate();
  const [, setAuth] = useState(true);
```

PR #1314, opened 2025-12-09
fix: defer import of proxy and service until config file has been set to fix race

### LINKED ISSUE #1313: Race condition in config loading prevents connection to mongoDB
**Describe the bug**
We've recently developed a race condition where the the database adaptors are being initialised on the default configuration, just before the user configuration is loaded. This results in always loading the neDB adaptor and failing to connect to mongo (when configured to do so). 

This is likely the result of recent changes to move to an es module setup. 

**To Reproduce**

1. Create a config that will load mongo, 
     - e.g. `cp proxy.config.json proxy.config.mongo.json` then edit the sink config to disable fd and enable mongo.
2. Add some console.log statements in src/db/index.ts to identify which DB adaptor is being loaded.
3. Add some console.log statements in src/config/index.ts to output the config at the end of `loadFullConfiguration()`
4. Build to dist (`npm run build`)
5. Run with a config file set via commandline arg: 
    - `cd dist`
    - node index.js --config ../proxy.config.mongo.json
6. Check the output - note that the config outputs a couple of times, once with the default config and again with the user config. The database adaptor is loaded between these two on the default config.

**Expected behavior**

User config is loaded before the database adaptor is initialised.

kriswest:
resolves #1313

By deferring the import of the Proxy and Service we can avoid them initialising the database clients before config is loaded, which was preventing configuration to use the mongo adaptor

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -166,7 +166,6 @@
       "integrity": "sha512-e7jT4DxYvIDLk1ZHmU/m/mB19rex9sv0c2ftBtjSBv+kVM/902eh0fINUzD7UwLLNR+jU585GxUJ8/EBfAM5fw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
         "@babel/generator": "^7.28.5",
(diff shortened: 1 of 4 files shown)


kriswest:
Great minds think alike @coopernetes, I have almost exactly the same PR awaiting review in Git Proxy + testing for our local deployment team. It works for me, I just need to check its still works from a bundle.

I'll copy in you initializeFolders change - I have  couple of similar clean-ups in my PR.

PR #1315, opened 2025-12-09
chore: upgrade node & mongo versions in ci, actions upgrades

coopernetes:
- add newer Node versions to build in CI so we test against all major LTS versions
- do the same for mongodb versions; CI is building against an ancient, long EOL version (4.4, Mongo is on version 8.x).
- used [ratchet](https://github.com/sethvargo/ratchet) to upgrade actions to their latest versions

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
         "./packages/git-proxy-cli"
       ],
       "dependencies": {
+        "@aws-sdk/credential-providers": "^3.940.0",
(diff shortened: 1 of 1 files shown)


jescalada (approved):
@coopernetes LGTM, just wondering if the CodeQL action bump to v4 is appropriate here or if we have a reason to stay at v3. I think `ratchet` might've set it to `3.31.7` because it's the "Latest" on the `codeql-action` GH releases page.

PR #1323, opened 2025-12-15
refactor: remove `any` and `as` throughout codebase

### LINKED ISSUE #1174: Clean up `any` and `as` in TS files

jescalada:
Fixes #1174.

### DIFF
--- a/src/proxy/chain.ts
+++ b/src/proxy/chain.ts
@@ -69,7 +69,7 @@ export const executeChain = async (req: Request, _res: Response): Promise<Action
       }
     }
   } catch (error: unknown) {
-    const msg = handleAndLogError(error, 'An error occurred when executing the chain');
+    const msg = handleAndLogError(error, 'An unexpected error occurred when executing the chain');
     action.error = true;
     action.errorMessage = msg;
   } finally {
--- a/src/proxy/processors/push-action/preReceive.ts
+++ b/src/proxy/processors/push-action/preReceive.ts
@@ -90,7 +90,7 @@ const exec = async (
     return action;
   } catch (error: unknown) {
     const msg = getErrorMessage(error);
-    const stdErrSuffix = stderrTrimmed ? `\n${stderrTrimmed}` : '';
+    const stdErrSuffix = stderrTrimmed ? `\nHook stderr: ${stderrTrimmed}` : '';
     step.error = true;
     step.log('Push failed, pre-receive hook returned an error.');
     step.setError(`Hook execution error: ${msg}${stdErrSuffix}`);


packages/git-proxy-cli/index.ts
```diff
@@ -47,13 +47,16 @@ async function login(username: string, password: string) {
...
+    if (isAxiosError(error) && error.response) {
       console.error(`Error: Login '${username}': '${error.response.status}'`);
       process.exitCode = 1;
-    } else {
+    } else if (error instanceof Error) {
       console.error(`Error: Login '${username}': '${error.message}'`);
       process.exitCode = 2;
+    } else {
+      console.error(`Error: Login '${username}': '${error}'`);
+      process.exitCode = 2;
```
kriswest:
```suggestion
    } else if (error instanceof Error) {
        console.error(`Error: Login '${username}': '${error instanceof Error ? error.message : String(error)'`);
        process.exitCode = 2;
```

packages/git-proxy-cli/index.ts
```diff
@@ -165,8 +168,9 @@ async function getGitPushes(filters: Partial<PushQuery>) {
     });
 
     console.log(util.inspect(records, false, null, false));
-  } catch (error: any) {
-    console.error(`Error: List: '${error.message}'`);
+  } catch (error: unknown) {
+    const msg = error instanceof Error ? error.message : String(error);
```
kriswest: this may be a 1-liner, but it is repeated a lot. How about a `getErrorMessage` Util fn that can be used inline instead? If we wanted a switch to enable stack traces  in future (where it is an instance of Error) we'd then be able todo so in that fn

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -503,9 +504,10 @@ const decompressGitObjects = async (buffer: Buffer): Promise<GitObject[]> => {
             offset++;
           }
         });
-      } catch (e) {
-        console.warn(`Error during decompression: ${JSON.stringify(e)}`);
-        error = new Error('Error during decompression', { cause: e });
+      } catch (error: unknown) {
+        const msg = error instanceof Error ? error.message : String(error);
+        console.warn(`Error during decompression: ${msg}`);
+        throw new Error(`Error during decompression: ${msg}`);
```
kriswest:
```suggestion
        const msg = `Error during decompression: ${error instanceof Error ? error.message : String(error)}`;
        console.warn(msg);
        throw new Error(msg);
```

src/proxy/processors/push-action/preReceive.ts
```diff
@@ -69,10 +71,11 @@ const exec = async (
       action.addStep(step);
     }
     return action;
-  } catch (error: any) {
+  } catch (error: unknown) {
+    const msg = error instanceof Error ? error.message : String(error);
     step.error = true;
     step.log('Push failed, pre-receive hook returned an error.');
-    step.setError(`Hook execution error: ${stderrTrimmed || error.message}`);
+    step.setError(`Hook execution error: ${stderrTrimmed || msg}`);
```
kriswest: Perhaps we should always print `msg` here (as it was thrown) and then append stderrTrimmed if its not falsey

src/proxy/processors/push-action/pullRemote.ts
```diff
@@ -24,6 +26,11 @@ const exec = async (req: any, action: Action): Promise<Action> => {
     step.log(`Executing ${cmd}`);
 
     const authHeader = req.headers?.authorization;
+
+    if (!authHeader) {
+      throw new Error('Authorization header is required');
```
kriswest: It is currently, although one day we could apply pullRemote to a pull chain to govern ingress, at which point this could be anonymous... but then I supposed we will want to log who the user pulling was... 

Could expand the error message in case it is sent back to the client... although I wonder if we'll ever hit it - you do get prompted for auth so there is probably a request before the push where that happens.

src/proxy/processors/push-action/preReceive.ts
```diff
@@ -69,10 +71,11 @@ const exec = async (
...
     action.addStep(step);
```
kriswest, line 97: I note we're not throwing here, but do in other cases. We should probably make that consistent. I believe the `continue()` function stops processing when `error` is set so we probably don't need to throw

src/proxy/processors/push-action/pullRemote.ts
```diff
@@ -41,9 +48,10 @@ const exec = async (req: any, action: Action): Promise<Action> => {
 
     step.log(`Completed ${cmd}`);
     step.setContent(`Completed ${cmd}`);
-  } catch (e: any) {
-    step.setError(e.toString('utf-8'));
-    throw e;
+  } catch (error: unknown) {
+    const msg = error instanceof Error ? error.message : String(error);
+    step.setError(msg);
+    throw error;
```
kriswest: probably don't need to throw here

src/proxy/processors/push-action/writePack.ts
```diff
@@ -27,9 +29,10 @@ const exec = async (req: any, action: Action) => {
     ];
     action.newIdxFiles = newIdxFiles;
     step.log(`new idx files: ${newIdxFiles}`);
-  } catch (e: any) {
-    step.setError(e.toString('utf-8'));
-    throw e;
+  } catch (error: unknown) {
+    const msg = error instanceof Error ? error.message : String(error);
+    step.setError(msg);
+    throw error;
```
kriswest: Don't need to throw here

src/service/routes/push.ts
kriswest: Taking care of this todo would ensure that `questions` is typed (or replaced by a typed var from the config), which will improve the typing of attestation below

kriswest: This could be typed - but you'll also need typing on `questions`, see note above about achieving that.

src/service/routes/repo.ts
```diff
@@ -5,11 +5,12 @@ import { getProxyURL } from '../urls';
 import { getAllProxiedHosts } from '../../db';
 import { RepoQuery } from '../../db/types';
 import { isAdminUser } from './utils';
+import { Proxy } from '../../proxy';
 
 // create a reference to the proxy service as arrow functions will lose track of the `proxy` parameter
 // used to restart the proxy when a new host is added
-let theProxy: any = null;
-const repo = (proxy: any) => {
+let theProxy: Proxy | null = null;
```
kriswest: I think we can probably eliminate the `| null` on this with a light refactor - theres no case where the proxy being null is valid. Does repo just need to be a function declaration, allowing it to work with its `proxy` arg instead?

src/ui/services/git-push.ts
```diff
@@ -15,17 +15,19 @@ const getPush = async (
...
-    data.diff = data.steps.find((x: Step) => x.stepName === 'diff');
-    setPush(data as PushActionView);
-  } catch (error: any) {
-    if (error.response?.status === 401) setAuth(false);
-    else setIsError(true);
-  } finally {
-    setIsLoading(false);
-  }
+  await axios<Action>(url, getAxiosConfig())
+    .then((response) => {
```
kriswest: mixing `async`/`await` and `then` is usually an anti-pattern. What did it accomplish here?

src/ui/services/git-push.ts
```diff
@@ -43,26 +45,31 @@ const getPushes = async (
...
-      setErrorMessage(processAuthError(error));
-    } else {
-      const message = error.response?.data?.message || error.message;
-      setErrorMessage(`Error fetching pushes: ${message}`);
-    }
-  } finally {
-    setIsLoading(false);
-  }
+  await axios<Action[]>(url.toString(), getAxiosConfig())
+    .then((response) => {
```
kriswest: mixing async/await and then is usually an anti-pattern. What did it accomplish here?

src/ui/views/PushRequests/components/PushesTable.tsx
```diff
@@ -86,7 +85,7 @@ const PushesTable: React.FC<PushesTableProps> = (props) => {
     <div>
       <Search onSearch={handleSearch} />
       <TableContainer component={Paper}>
-        <Table className={classes.table} aria-label='simple table'>
+        <Table aria-label='simple table'>
```
kriswest, line 117: aria-label should describe the table's purpose - the current label probably came from the template

src/ui/views/PushRequests/PushRequests.tsx
kriswest: differs from the other tab defs

src/ui/views/UserList/Components/UserList.tsx
```diff
@@ -66,7 +62,7 @@ const UserList: React.FC = () => {
       <GridItem xs={12} sm={12} md={12}>
         <Search onSearch={handleSearch} placeholder='Search users...' />
         <TableContainer component={Paper}>
-          <Table className={classes.table} aria-label='simple table'>
+          <Table aria-label='simple table'>
```
kriswest, line 81: aria-label should describe the table.

Was the class declaration not needed? It (`MuiTable-root` ?) does seem to have a small effect on styling.

src/plugin.ts
```diff
@@ -1,4 +1,7 @@
+import { Request } from 'express';
+
 import { Action } from './proxy/actions';
+import Module from 'node:module';
 
 const lpModule = import('load-plugin');
```
kriswest: odd import statement... Doesn't match the docs for load-plugin - there is no default import which may be why... https://www.npmjs.com/package/load-plugin


```suggestion
import {loadPlugin, resolvePlugin} from 'load-plugin'
```

src/plugin.ts
```diff
@@ -101,17 +104,18 @@ class PluginLoader {
...
    * Resolve & load a Node module from either a given specifier (file path, import specifier or package name) using load-plugin.
    * @param {string} target The module specifier to load
-   * @return {Promise<Module>} A resolved & loaded Module
+   * @return {Promise<unknown>} A resolved & loaded Module
    */
-  private async _loadPluginModule(target: string): Promise<any> {
+  private async _loadPluginModule(target: string): Promise<unknown> {
     const lp = await lpModule;
     const resolvedModuleFile = await lp.resolvePlugin(target);
     return await lp.loadPlugin(resolvedModuleFile);
```
kriswest: See above comment about the import, this can be refactored to:


```suggestion
const resolvedModuleFile = await resolvePlugin(target);
return loadPlugin(resolvedModuleFile);
```

No need to await loadPlugin as the function is async and returns a promise anyway

test/processors/gitLeaks.test.ts
```diff
@@ -1,16 +1,18 @@
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import { Request } from 'express';
+
 import { Action, Step } from '../../src/proxy/actions';
 
 vi.mock('../../src/config', async (importOriginal) => {
-  const actual: any = await importOriginal();
+  const actual = await importOriginal<typeof import('../../src/config')>();
```
kriswest, line 23: feels odd, but I see others doing the same...

test/ConfigLoader.test.ts
```diff
@@ -627,9 +627,6 @@ describe('Validation Helpers', () => {
       expect(isValidGitUrl('not-a-git-url')).toBe(false);
       expect(isValidGitUrl('http://github.com/user/repo')).toBe(false);
       expect(isValidGitUrl('')).toBe(false);
-      expect(isValidGitUrl(null as any)).toBe(false);
-      expect(isValidGitUrl(undefined as any)).toBe(false);
-      expect(isValidGitUrl(123 as any)).toBe(false);
```
kriswest, line 753: These cases are still valid to test, even though Typescript is no doubt complaining about them. How about changing to:

`expect(isValidGitUrl(null as unknown as string)).toBe(false);`

test/ConfigLoader.test.ts
```diff
@@ -645,14 +642,6 @@ describe('Validation Helpers', () => {
...
       // Invalid paths
       expect(isValidPath('')).toBe(false);
-      expect(isValidPath(null as any)).toBe(false);
-      expect(isValidPath(undefined as any)).toBe(false);
-
-      // Additional edge cases
-      expect(isValidPath({} as any)).toBe(false);
-      expect(isValidPath([] as any)).toBe(false);
-      expect(isValidPath(123 as any)).toBe(false);
-      expect(isValidPath(true as any)).toBe(false);
```
kriswest, line 776: as above

test/ConfigLoader.test.ts
```diff
@@ -672,8 +661,6 @@ describe('Validation Helpers', () => {
       expect(isValidBranchName('-invalid')).toBe(false);
       expect(isValidBranchName('branch with spaces')).toBe(false);
       expect(isValidBranchName('')).toBe(false);
-      expect(isValidBranchName(null as any)).toBe(false);
-      expect(isValidBranchName(undefined as any)).toBe(false);
```
kriswest, line 797: These should probably also be retained as test cases. 

It occurs to me that if coerced to a string they may become valid, but if they were passing before they should continue to do so if you use `as unknown as string`

packages/git-proxy-cli/index.ts
jescalada: Done, added util functions that might be used in different situations (`handleAndLogError`, `getErrorMessage` and `handleAndThrowError`)

src/proxy/processors/push-action/pullRemote.ts
jescalada: Updated to the following:

"Authorization header is required for pullRemote. Make sure to provide valid credentials as anonymous pulls are not currently supported."

src/service/routes/push.ts
jescalada: I fixed this by adding proper typing for the response: `CompletedAttestation` and `AttestationAnswer`

Also extracted a `validateAttestation` function that compares the questions from the request with the original ones in the config, as well as validating that all checkboxes were ticked.

src/service/routes/repo.ts
jescalada: Converted `repo` into a function, which seems to work fine. 👍🏼

src/ui/services/git-push.ts
jescalada: I originally refactored to use promise chaining since it allows typing the error as `AxiosError` and skipping a check/casting there. Here's another version that does the check inline:

```ts
const getPush = async (
  id: string,
  setIsLoading: (isLoading: boolean) => void,
  setPush: (push: PushActionView) => void,
  setAuth: (auth: boolean) => void,
  setIsError: (isError: boolean) => void,
): Promise<void> => {
  setIsLoading(true);

  try {
    const data: Action & { diff?: Step } = response.data;
    data.diff = data.steps.find((x: Step) => x.stepName === 'diff');
    setPush(data as PushActionView);
  } catch (error: unknown) {
    if (error instanceof AxiosError && error.response?.status === 401) setAuth(false);
    else setIsError(true);
  } finally {
    setIsLoading(false);
  }
};
```

jescalada: I noticed that there are many other places in the code that are mixing `async`/`await` and `then`. Perhaps this could be done in a separate PR? Added an issue for this: #1412

Seems like a tedious albeit simple job that could be done with the agentic implementation command added in #1410 🤔

src/ui/services/git-push.ts
jescalada: Made issue to track this for all other usages: #1412

src/ui/views/PushRequests/components/PushesTable.tsx
jescalada: Good catch! Removing the `className`s didn't affect the styling at all... 👍🏼

test/processors/gitLeaks.test.ts
jescalada: I recall the "recommended" solution (I think from Vitest docs/error message) is to use `any`. This is the only way I found to remove the `any` type...

test/ConfigLoader.test.ts
jescalada: Seems all of the tests still pass when doing the `as unknown as string` casting 👍🏼

src/config/validators.ts
```diff
@@ -103,9 +104,7 @@ export async function loadConfig(
 function parseGitProxyConfig(raw: string, context: string): GitProxyConfig {
   try {
     return Convert.toGitProxyConfig(raw);
-  } catch (error) {
-    throw new Error(
-      `Invalid configuration format in ${context}: ${error instanceof Error ? error.message : 'Unknown error'}`,
-    );
+  } catch (error: unknown) {
+    throw new Error(`Invalid configuration format in ${context}: ${getErrorMessage}`);
```
kriswest: You're not calling the function:


```suggestion
    throw new Error(`Invalid configuration format in ${context}: ${getErrorMessage(error)}`);
```

src/utils/errors.ts
```diff
@@ -0,0 +1,15 @@
+export const getErrorMessage = (error: unknown) => {
+  return error instanceof Error ? error.message : String(error);
+};
+
+export const handleAndLogError = (error: unknown, message: string): string => {
```
kriswest: I'd consider renaming `message` to `messagePrefix` or adding jsdocs to explain how its used (to prefix the error message).

If could also be an optional property (if there are any cases where you just want to log exactly the error).

kriswest (changes requested):
Mostly LGTM, one small issue. The very large diff includes changes from other PRs - that can be avoided by rebasing main rather than merging in future.

I think it best if this goes in ASAP as its getting way too long to review effectively.

One small error to fix + a minor refinement suggested.

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -494,8 +496,8 @@ const decompressGitObjects = async (buffer: Buffer): Promise<GitObject[]> => {
     };
 
     // stop on errors, except maybe buffer errors?
-    const onError = (e: any) => {
-      error = e;
+    const onError = (e: unknown) => {
+      error = e instanceof Error ? e : new Error(String(e));
       console.warn(`Error during inflation: ${JSON.stringify(e)}`);
```
kriswest: `e` could be an error here or a string... you'll need to use `getErrorMessage` on `error` for consistency.

src/proxy/processors/push-action/parsePush.ts
```diff
@@ -521,9 +523,10 @@ const decompressGitObjects = async (buffer: Buffer): Promise<GitObject[]> => {
             offset++;
           }
         });
-      } catch (e) {
-        console.warn(`Error during decompression: ${JSON.stringify(e)}`);
-        error = new Error('Error during decompression', { cause: e });
+      } catch (error: unknown) {
+        const msg = `Error during decompression: ${error instanceof Error ? error.message : String(error)}`;
+        console.warn(msg);
+        throw new Error(msg);
```
kriswest: I'd be cautious with changing this to throw here, this code is a horror to deal with and I'm not sure how well covered all the the error cases are in the tests. 

Note that because you're throwing without a catch you're now not destroying the inflater and you are exiting earlier (where previously we decoded more of the objects - we may not need to but it certainly helped to beug in the past as you could tell which object it blew on - although I've since removed that logging (it was massively bloating the logs). I'd probably revert... otherwise introduce a finally to ensure that the inflator cleanup still happens.

kriswest: I'd be cautious with changing this to throw here, this code is a horror to deal with and I'm not sure how well covered all the the error cases are in the tests. 

Note that because you're throwing without a catch you're now not destroying the inflater and you are exiting earlier (where previously we decoded more of the objects - we may not need to but it certainly helped to beug in the past as you could tell which object it blew on - although I've since removed that logging (it was massively bloating the logs). I'd probably revert... otherwise introduce a finally to ensure that the inflator cleanup still happens.

src/proxy/processors/push-action/preReceive.ts
```diff
@@ -85,10 +88,12 @@ const exec = async (
       action.addStep(step);
     }
     return action;
-  } catch (error: any) {
+  } catch (error: unknown) {
+    const msg = getErrorMessage(error);
+    const stdErrSuffix = stderrTrimmed ? `\n${stderrTrimmed}` : '';
```
kriswest: This is an improvement over the || on line 91. However, it could do with an intro to the stderr (as the message may or may not relate to it)


```suggestion
    const stdErrSuffix = stderrTrimmed ? `\nHook stderr: ${stderrTrimmed}` : '';
```

src/proxy/processors/push-action/pullRemote.ts
```diff
@@ -43,6 +46,11 @@ const exec = async (req: any, action: Action): Promise<Action> => {
       step.log(`Executing ${cmd}`);
 
       const authHeader = req.headers?.authorization;
+      if (!authHeader) {
```
kriswest, line 49: Good catch - this was just confusing previously

src/proxy/chain.ts
```diff
@@ -65,10 +68,10 @@ export const executeChain = async (req: any, res: any): Promise<Action> => {
         checkoutCleanUpRequired = true;
       }
     }
-  } catch (e) {
+  } catch (error: unknown) {
+    const msg = handleAndLogError(error, 'An error occurred when executing the chain');
```
kriswest: If we're now consistent about whether processors throw errors or not (I noticed you fixed a couple of them that would throw rather than setError) then these are **unexpected** errors (where expected errors are those the processors are looking for as opposed to the processors just blowing up).

```suggestion
    const msg = handleAndLogError(error, 'An unexpected error occurred when executing the chain');
```

src/ui/services/git-push.ts
```diff
@@ -14,6 +14,22 @@
  * limitations under the License.
  */
 
+/**
```
kriswest, line 1: Double license header has crept in somehow

kriswest, line 1: Double license header has crept in somehow

src/ui/services/repo.ts
```diff
@@ -14,6 +14,22 @@
  * limitations under the License.
  */
 
+/**
```
kriswest, line 1: Double license header

src/proxy/processors/push-action/parsePush.ts
jescalada: Replaced with the following:

```ts
const onError = (e: unknown) => {
  const msg = getErrorMessage(e);
  console.warn(`Error during inflation: ${msg}`);
  error = new Error(`Error during inflation: ${msg}`);
  inflater.end();
  done = true;
  if (currentWriteResolve) currentWriteResolve();
};
```

jescalada: The `error = e` bit was quite odd since it's immediately overwritten afterwards with the new `Error`.

src/proxy/processors/push-action/parsePush.ts
jescalada: I didn't notice that we were throwing any errors later on:

```ts
// throw any error that was caught as we were not able to read the pack file in full
if (error) {
  throw error;
}
```

Instead of throwing, I'm now assigning it to the `error` that gets checked at the bottom:

```ts
} catch (e: unknown) {
  const msg = getErrorMessage(e);
  console.warn(`Error during decompression: ${msg}`);
  error = new Error(`Error during decompression: ${msg}`);
}
```

src/ui/services/repo.ts
jescalada: Double-checked that the linter script isn't messing up these specific files. Seems to be working correctly ATM 🤷🏼

jescalada:
@kriswest Ready for another look!

PR #1332, opened 2025-12-17
feat(ssh): Add SSH agent forwarding

### LINKED ISSUE #27: SSH support

### LINKED ISSUE #61: SSH: Implement SSH agent forwarding protocol support
The current SSH key capture implementation has a TODO for full SSH agent forwarding support. Implementing this would provide more secure key handling and a better user experience.

fabiovincenzi:
## Summary

  Adds complete SSH protocol support to Git Proxy with SSH agent forwarding.

  ## Key Features

  - **SSH Proxy Server** (`src/proxy/ssh/server.ts`) - Full SSH protocol handling
  - **SSH Agent Forwarding** (`src/proxy/ssh/AgentForwarding.ts`, `AgentProxy.ts`) - Securely forwards client SSH credentials to upstream Git servers
  - **SSH Key Management** - API endpoints, UI components, and CLI tools for user SSH key registration
  - **Git Protocol Handling** (`src/proxy/ssh/GitProtocol.ts`) - Parses and processes `git-upload-pack`/`git-receive-pack` over SSH

  ## Documentation

  - [`docs/SSH_SETUP.md`](docs/SSH_SETUP.md) - User guide for configuring SSH access (key generation, agent forwarding, Git remote setup)
  - [`docs/SSH_ARCHITECTURE.md`](docs/SSH_ARCHITECTURE.md) - Technical architecture details, component overview, and security

  ## Tests

  Comprehensive test coverage in `test/ssh/`:


docs/SSH_ARCHITECTURE.md
```diff
@@ -0,0 +1,293 @@
...
+### Client Setup
+
+**1. Configure Git remote**:
+
+```bash
+# For GitHub
+git remote add origin ssh://git@git-proxy.example.com:2222/github.com/org/repo.git
+
+# For GitLab
+git remote add origin ssh://git@git-proxy.example.com:2222/gitlab.com/org/repo.git
```
jescalada: Chances are the remote is already present - in that case we need to set it with:

```bash
git remote set-url origin <remote-url>
```

docs/SSH_ARCHITECTURE.md
```diff
@@ -0,0 +1,293 @@
...
+The **SSH host key** is the proxy server's cryptographic identity. It identifies the proxy to clients and prevents man-in-the-middle attacks.
+
+**Auto-generated**: On first startup, git-proxy generates an Ed25519 host key stored in `.ssh/host_key` and `.ssh/host_key.pub`.
+
+**Important**: The host key is NOT used for authenticating to GitHub/GitLab. Agent forwarding handles remote authentication using the client's keys.
+
+**First connection warning**:
+
+```
+The authenticity of host '[localhost]:2222' can't be established.
```
jescalada: Since the host IP won't be localhost in production, we could change this to <git-proxy-host-ip> just to keep things consistent!

docs/SSH_ARCHITECTURE.md
```diff
@@ -0,0 +1,293 @@
...
+                        └─────────────┘
+```
+
+---
+
+## SSH Host Key (Proxy Identity)
+
+The **SSH host key** is the proxy server's cryptographic identity. It identifies the proxy to clients and prevents man-in-the-middle attacks.
+
+**Auto-generated**: On first startup, git-proxy generates an Ed25519 host key stored in `.ssh/host_key` and `.ssh/host_key.pub`.
```
jescalada: We might want to think carefully about the `host_key` location. Right now we're generating to the same directory where GitProxy is being executed (normally, an administrator would run it through `npx`), which might not be the best location considering we only need one host key per machine.

In any case, we should update the docs here so that it's clear where these keys are being generated.

docs/SSH_ARCHITECTURE.md
```diff
@@ -0,0 +1,293 @@
...
+ls -la ~/.ssh/id_*.pub
+
+# If no key exists, generate a new Ed25519 key
+ssh-keygen -t ed25519 -C "your_email@example.com"
+# Press Enter to accept default location (~/.ssh/id_ed25519)
+# Optionally set a passphrase for extra security
+```
+
+**3. Start ssh-agent and load key**:
+
```
jescalada: Should explain that this must be executed on the same terminal that you're trying to push! And that if not it might fail unless the user's PC handles the agent sharing.

src/proxy/ssh/AgentForwarding.ts
```diff
@@ -0,0 +1,289 @@
...
+          clientIp: this.client.clientIp,
+          authenticatedUser: this.client.authenticatedUser,
+        });
+
+        // ssh2's AgentContext.init() calls parseKey() on every key we return.
+        // We need to return the raw pubKeyBlob Buffer, which parseKey() can parse
+        // into a proper ParsedKey object.
+        const keys = identities.map((identity) => identity.publicKeyBlob);
+
+        console.log(`[LazyAgent] Returning ${keys.length} identities`);
```
jescalada, line 111: If no identities are found, we should throw a descriptive error:

```suggestion
        console.log(`[LazyAgent] Returning ${keys.length} identities`);
        
        if (keys.length === 0) {
          throw new Error('No identities found. Run ssh-add <key> on this terminal to add your SSH key.');
        }
```

src/proxy/ssh/server.ts
```diff
@@ -0,0 +1,623 @@
...
+            } else {
+              console.log('[SSH] Public key authentication failed - key not found');
+              ctx.reject();
+            }
+          })
+          .catch((err: Error) => {
+            console.error('[SSH] Database error during public key auth:', err);
+            ctx.reject();
+          });
+      } else if (ctx.method === 'password') {
```
jescalada: I have a feeling this should be removed. Apparently this is a fallback for when there are no valid keys present in the user's machine. (?) The user sees a `git@<host-ip>'s password:` prompt that asks for the password, and the `ctx.username` value here would be `git`. Of course, this fails if `git` is not a GitProxy user. This is why my flow was passing the check when using `admin@localhost` with the `admin` password.

I haven't been able to send a message to the user that only public key auth is possible. Wonder if this is possible at all? Simply throwing errors or emitting error events doesn't seem to work.

src/proxy/ssh/GitProtocol.ts
```diff
@@ -0,0 +1,396 @@
...
+
+    remoteGitSsh.on('error', (err: Error) => {
+      console.error(`[executeRemoteGitCommand] Connection error:`, err);
+      clearTimeout(timeout);
+      if (clientStream) {
+        // Provide more helpful error messages based on the error type
+        let errorMessage = `Connection error: ${err.message}\n`;
+
+        // Detect authentication failures and provide actionable guidance
+        if (err.message.includes('All configured authentication methods failed')) {
```
jescalada, line 209: I've gotten this error when having everything set up correctly EXCEPT the GitHub SSH key. For some reason though, I don't get the modified error message...

src/proxy/ssh/server.ts
```diff
@@ -0,0 +1,623 @@
...
+      stream.end();
+      return;
+    }
+  }
+
+  public start(): void {
+    const sshConfig = getSSHConfig();
+    const port = sshConfig.port || 2222;
+
+    this.server.listen(port, '0.0.0.0', () => {
```
jescalada: Double-check if it's okay to allow SSH connections from any IP. 🤔

jescalada (changes requested):
I've been testing with GitProxy and a client running on separate devices, and the main flow is working correctly! 🚀 

We might want to improve the error handling a bit more.

docs/SSH_ARCHITECTURE.md
jescalada: Perhaps we could add this to the ~/.ssh folder (which would be the correct folder UNLESS we're in dev). We should also rename it to `git_proxy_host_key` for clarity.

src/proxy/ssh/server.ts
```diff
@@ -0,0 +1,623 @@
...
+        fullRepoPath = fullRepoPath.substring(1);
+      }
+
+      // Parse full path to extract hostname and repository path
+      // Input: 'github.com/user/repo.git' -> { host: 'github.com', repoPath: '/user/repo.git' }
+      const fullUrl = `https://${fullRepoPath}`; // Construct URL for parsing
+      const urlComponents = processGitUrl(fullUrl);
+
+      if (!urlComponents) {
+        throw new Error(`Invalid repository path format: ${fullRepoPath}`);
```
jescalada: Improved error recovery:

```suggestion
        throw new Error(`Invalid repository path format: ${fullRepoPath} Make sure the repository URL is valid and ends with '.git'.`);
```

docs/SSH_ARCHITECTURE.md
jescalada: Might want to emphasize that the repo URL should contain the `.git` or will otherwise fail!

fabiovincenzi: Also improved the error message and added a test for this case

src/proxy/ssh/server.ts
fabiovincenzi: I've removed the explicit '0.0.0.0' binding. This doesn't change the security behavior - the server still accepts connections from all IPs as before. However, it now aligns with how the HTTP/HTTPS servers work (using Node.js default binding), and also adds IPv6 support.

src/proxy/ssh/GitProtocol.ts
fabiovincenzi: This is weird I've tested this and the custom error message does work correctly for me

src/proxy/ssh/server.ts
fabiovincenzi: Removed password auth completely, now only accepts publickey.
You're right that it was problematic, the password fallback was confusing.
In my opinion this is the correct approach for SSH, password auth doesn't make sense here since we need agent forwarding anyway for remote Git operations.

jescalada (approved):
Retested the flows from before and things seem to be working as expected! 🚀 

Wondering if we can get some other orgs to review this? @kriswest @coopernetes @06kellyjac

fabiovincenzi:
@finos/git-proxy-maintainers This is ready for review!

fabiovincenzi:
The build-windows failure seems to be related to a flaky Windows + Node 24.x test, not related to the changes in this PR. @kriswest  I recall you mentioned a flaky Windows test — is this the same one?

krisstern:
Hi @fabiovincenzi I think you pinged the wrong Kris

fabiovincenzi:
> Hi @fabiovincenzi I think you pinged the wrong Kris

@krisstern Sorry!

ARCHITECTURE.md
jescalada, line 1: I believe this might be an agentic AI artifact from when we didn't yet have the Architecture guide. We should remove this since we now have a curated, high-quality version:
https://github.com/finos/git-proxy/blob/main/website/docs/architecture/architecture.md

.nvmrc
jescalada, line 1: Check if this is necessary - if so, we might want to set at v22 or v24 considering v20 has been deprecated in #1456

jescalada:
I found a few bugs worth noting while testing today, most notably:

### Concurrent SSH requests failing due to nonexistent channels:

This error happens when running concurrent git commands through SSH (client-side output):

```bash
channel_by_id: 2: bad id: channel free
Disconnecting 10.32.1.86 port 2222: data packet referred to nonexistent channel 2
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

I just pushed a fix for this one which was simply due to the `incoming` and `outgoing` channel IDs being flipped. It worked with single requests because both use the same channel ID (1) for packet transport.

### An error when calling `git push` when there aren't any commits to push:

```bash
juan@pop-os:~/git-proxy-experiments$ git push
Access denied: Unable to parse push. Please contact an administrator for support: Your push has been blocked. Please make sure you are pushing to a single branch.
error: failed to push some refs to 'ssh://10.32.1.86:2222/github.com/jescalada/git-proxy-experiments.git'
```

When pushing *without* SSH, this just gives us the expected "Everything up-to-date". I'll look into this further and find a fix 👍🏼 

### An error after fixing merge conflicts and pushing a merge commit. The first "approval" push was apparently skipped, so there wasn't any push for an admin to approve and break the loop:

```bash
juan@pop-os:~/git-proxy-experiments$ git push
Enumerating objects: 10, done.
Counting objects: 100% (10/10), done.
Delta compression using up to 14 threads
Compressing objects: 100% (6/6), done.
Writing objects: 100% (6/6), 690 bytes | 690.00 KiB/s, done.
Total 6 (delta 2), reused 0 (delta 0), pack-reused 0
Access denied: Unreferenced commits in pack (2): e9260e6c519ac964e895ec69d926f844638c9c60, 9ed3a288070f3a7f58494d1a50c99435cff1757c.
This usually happens when a branch was made from a commit that hasn't been approved and pushed to the remote.
Please get approval on the commits, push them and try again.
send-pack: unexpected disconnect while reading sideband packet
fatal: the remote end hung up unexpectedly
```

I'll try reproducing this one again, as it may be due to a buggy push action from before our `rc`s. Note that the buggy state can be resolved by simply rebasing instead of dealing with merge commits...

jescalada:
Just pushed a fix for the error when calling `git push` on a branch that's up-to-date. Basically, HTTPS handles no-op pushes automatically (via GET /info/refs) and never actually triggers the chain, whereas SSH was executing the chain and erroring out in `parsePush`.

I fixed it by modifying the `endHandler` in `src/proxy/ssh/server.ts` to actually parse the packet lines beforehand, detect and skip no-op pushes and keeping the data integrity and chain execution bits intact.

kriswest:
@re-vlad

andypols (approved):
LGTM.  Checked it out and it worked as expected.

config.schema.json
```diff
@@ -291,6 +291,16 @@
...
+    "limits": {
+      "description": "Configuration for various limits",
+      "type": "object",
+      "properties": {
+        "maxPackSizeBytes": {
+          "type": "number",
+          "description": "Maximum size of a pack file in bytes (default 1GB)"
+        }
+      }
+    },
```
kriswest, line 324: Don't allow properties that are not defined.
```suggestion
    "limits": {
      "description": "Configuration for various limits",
      "type": "object",
      "properties": {
        "maxPackSizeBytes": {
          "type": "number",
          "description": "Maximum size of a pack file in bytes (default 1GB)"
        }
      },
      "additionalProperties": false
    },
```

config.schema.json
```diff
@@ -368,6 +378,31 @@
...
+          "type": "string",
+          "description": "Custom error message shown when SSH agent forwarding is not enabled or no keys are loaded in the client's SSH agent. If not specified, a default message with git config commands will be shown. This allows organizations to customize instructions based on their security policies."
+        },
+        "debug": {
+          "type": "boolean",
+          "description": "Enable verbose SSH protocol debug logging (both for the local SSH server and for outbound connections to remote Git servers). Emits one log line per SSH packet, so leave disabled in production.",
+          "default": false
+        }
+      },
+      "required": ["enabled"]
```
kriswest: Don't allow undefined properties:

```suggestion
      "required": ["enabled"],
      "additionalProperties": false
```

src/config/generated/config.ts
```diff
@@ -455,6 +465,17 @@ export interface Domains {
...
 
+/**
+ * Configuration for various limits
+ */
+export interface Limits {
+  /**
+   * Maximum size of a pack file in bytes (default 1GB)
+   */
+  maxPackSizeBytes?: number;
+  [property: string]: any;
```
kriswest: Setting `"additionalProperties": false` in teh schema will eliminate this index type

src/config/generated/config.ts
```diff
@@ -528,6 +549,38 @@ export enum DatabaseType {
...
+   * Enable SSH proxy server. When enabled, clients can connect via SSH and the proxy will
+   * forward their SSH agent to authenticate with remote Git servers.
+   */
+  enabled: boolean;
+  /**
+   * Port for SSH proxy server to listen on. Clients connect to this port instead of directly
+   * to GitHub/GitLab.
+   */
+  port?: number;
+  [property: string]: any;
```
kriswest: Setting `"additionalProperties": false` in the schema will eliminate this index type - worth doing so that misspelt or unknown properties will get picked up in IDEs

src/config/index.ts
```diff
@@ -115,7 +117,8 @@ function loadFullConfiguration(): FullGitProxyConfig {
     return _currentConfig;
   }
 
-  const rawDefaultConfig = Convert.toGitProxyConfig(JSON.stringify(defaultSettings));
+  // Skip QuickType validation for now due to SSH config issues
+  const rawDefaultConfig = defaultSettings as any;
```
kriswest: Whats the config issue? I'd rather not disable validation

src/config/index.ts
```diff
@@ -393,6 +406,63 @@ export const getRateLimit = () => {
...
+    const sshConfig = config.ssh || { enabled: false };
+
+    // The host key is a server identity, not user configuration
+    if (sshConfig.enabled) {
+      sshConfig.hostKey = defaultHostKey;
+    }
+
+    return sshConfig;
+  } catch (error) {
+    // If config loading fails due to SSH validation, try to get SSH config directly from user config
```
kriswest: We should address what the underlying issue is rather than having this workaround... Whats the problem you are having with validation?

src/db/mongo/users.ts
```diff
@@ -79,3 +83,56 @@ export const updateUser = async (user: Partial<User>): Promise<void> => {
   const collection = await connect(collectionName);
   await collection.updateOne(filter, { $set: userWithoutId }, options);
 };
+
+export const addPublicKey = async (username: string, publicKey: PublicKeyRecord): Promise<void> => {
```
kriswest, line 87: If we only need to store the public key for this feature then that probably fine to store in plaintext. Some orgs require all user data to be encyrpted at rest, but we already have an unencrypted email and username, so I think this is fine.

src/proxy/actions/Action.ts
```diff
@@ -54,6 +54,24 @@ class Action {
   lastStep?: Step;
   proxyGitPath?: string;
   newIdxFiles?: string[];
+  protocol?: 'https' | 'ssh';
+  sshUser?: {
+    username: string;
+    email?: string;
```
kriswest: Don't these duplicate the existing `user` and `userEmail` fields?

src/proxy/ssh/GitProtocol.ts
```diff
@@ -0,0 +1,417 @@
...
+      });
+    },
+  );
+
+  return parser.getBuffer();
+}
+
+/**
+ * Forward pack data to remote Git server (used for push operations)
+ * This connects to GitHub, sends the validated pack data, and forwards responses
```
kriswest:
```suggestion
 * This connects to the upstream server, sends the validated pack data, and forwards responses
```

src/proxy/ssh/GitProtocol.ts
```diff
@@ -0,0 +1,417 @@
...
+
+  await executeRemoteGitCommand(
+    command,
+    client,
+    remoteHost,
+    { clientStream: stream, debug: isDebugEnabled(), keepalive: true },
+    (remoteStream) => {
+      console.log(`[SSH] Forwarding pack data for user ${userName}`);
+
+      // Send pack data to GitHub
```
kriswest:
```suggestion
      // Send pack data to upstream
```

src/proxy/ssh/knownHosts.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+/**
+ * Default known host keys for GitHub and GitLab
+ * Last updated: 2025-01-26
+ */
+export const DEFAULT_KNOWN_HOSTS: KnownHostsConfig = {
+  'github.com': 'SHA256:+DiY3wvvV6TuJJhbpZisF/zLDA0zPMSvHdkr4UvCOqU',
+  'gitlab.com': 'SHA256:eUXGGm1YGsMAS7vkcx6JOJdOGHPem5gQp4taiCfCLB8',
+};
```
kriswest: Perhaps these defaults should be drawn from proxy.config.json (like other defaults)

src/service/urls.ts
```diff
@@ -19,18 +19,84 @@ import { Request } from 'express';
 import { serverConfig } from '../config/env';
 import * as config from '../config';
 
-const { GIT_PROXY_SERVER_PORT: PROXY_HTTP_PORT, GIT_PROXY_UI_PORT: UI_PORT } = serverConfig;
+const {
+  GIT_PROXY_SERVER_PORT: PROXY_HTTP_PORT,
+  GIT_PROXY_UI_PORT: UI_PORT,
+  GIT_PROXY_UI_HOST: UI_HOST,
+} = serverConfig;
```
kriswest: n.b. #1554 will affect this by normalising these settings with other config type.

website/docs/ssh-setup.md
```diff
@@ -0,0 +1,252 @@
+# SSH Setup Guide
```
kriswest, line 1: Could you integrate this into the user manual document now that its been merged?

kriswest (commented):
This looks pretty good to me, a few minor comments and one concern - the disabling of config validation. What is the validation issue with the SSH config thats causing that ? If its a JSON schema issue I may be able to help resolve.

src/config/index.ts
fabiovincenzi: Good catch! The validation was disabled as a workaround during the initial SSH implementation and was never revisited. The underlying issue no longer exists, so I've restored the validation and removed the try/catch fallback in `getSSHConfig()`.

src/proxy/actions/Action.ts
fabiovincenzi: You're right, these are duplicates, `sshUser` was originally added as part of the SSH Key Retention which was removed. Removed it along with encryptedSSHKey and sshKeyExpiry.

src/config/generated/config.ts
fabiovincenzi: For the SSH related interfaces such as `SSH` and `HostKey` I set `"additionalProperties": false` in the schema.
For the remaining interfaces (`Gitleaks`, `AuthenticationElement`, `Database`, `TLS`, etc.) this is a valid improvement but I think is beyond the scope of this PR. Happy to address it in a follow-up PR.

fabiovincenzi:
Thanks for the review @kriswest! I believe all your comments have been addressed. Let me know if anything else needs attention.

kriswest (approved):
LGTM after comments were dealt with

kriswest:
@jescalada's early comments still need to be resolved to merge: https://github.com/finos/git-proxy/pull/1332#pullrequestreview-3591440411

src/proxy/ssh/server.ts
jescalada: Makes sense, although we should revisit this during a security audit down the line 👍🏼

jescalada:
Just verified that the older comments have been dealt with - ready to merge! 🚀

PR #1353, opened 2026-01-16
fix: error message on fetching auth'd user

### LINKED ISSUE #1345: Error messages displayed in admin UI when not logged in and broken login link
**Describe the bug**
At present on main, with the default config and if the user is not logged in, then the below error is displayed in the header:

Image

Using the login link in the user menu (top-right_ results in the following being displayed instead of the login page:

Image

If you manually enter the login page URL it does load, and once you login the errors disappear. Returning when you log out. This may have been an unexpected side-effect of #1293 @jescalada as it changed the response codes, which need to handle 401 differently in various places.

Further, the error message appears to assume that jwtAuth is enabled (it is not by default), I believe thats due to: https://github.com/finos/git-proxy/blob/ecd616139ec2826ea72ab01f732ac67c403fdda2/src/ui/services/auth.ts#L51-L55

The error messages in both branches of that if statement assume jwtAuth is enabled, but thats not the case in the default config where its turned off. You first need to check whether jwtAuth is enabled before appending content about it to the message...

**To Reproduce**

1. Checkout main
2. run `npm install; npm run start`
3. Visit the admin UI URL, e.g. http://localhost:3000/dashboard/repo
4. observe the error message displayed top right
5. Attempt to login via the link in the user menu (top-right)
6. observe content displayed instead of the login page.

**Expected behaviour**
401 response codes from the auth API calls are correctly handled.

jescalada:
Fixes #1345


kriswest (approved):
LGTM

PR #1355, opened 2026-01-17
fix: add invalid regex validation for `commitConfig`

### LINKED ISSUE #1336: Invalid regex in `commitConfig` causes pushes to crash/error out
When setting invalid regex for the patterns in `commitConfig`, pushes either fail (in the case of `commitConfig.author.email.local.block`) or give a misleading error (in the case of `commitConfig.message.block.patterns`.

### Invalid `author` config:

```json
"author": {
  "email": {
    "local": {
      "block": "[invalid"
    },
    "domain": {
      "allow": "(?<invalid"
    }
  }
},
```

#### Output:

Image

### Invalid `message` config:

```json
"message": {
  "block": {
    "literals": ["safe"],
    "patterns": [
      "[unclosed",
      "(?<incomplete",
      "*invalid",
      "(?P<invalid>",
      "\\k<notdefined>"
    ]
  }
},
```

#### Output:

Image

### Invalid `diff` config:

```json
"diff": {
  "block": {
    "literals": [],
    "patterns": [
      "[unclosed",
      "(?<incomplete"
    ],
    "providers": {
      "Bad Provider": "*starts-with-quantifier",
      "Unclosed Group": "(forgot to close"
    }
  }
}
```

#### Output:

Image

These are likely fixed by improving the error handling in the `checkCommitMessages.ts`, `checkAuthorEmails.ts` and `scanDiff.ts` processors. We should make sure each action gives a descriptive error on invalid regex but blocks the push instead of crashing.

**Additional context**
Found when writing documentation for the chain in #1335

jescalada:
Fixes #1336.

Most of the changes are unit tests for the new code and `/src/config/index.ts`.


src/config/ConfigLoader.ts
```diff
@@ -189,6 +190,11 @@ export class ConfigLoader extends EventEmitter {
       } else {
         console.log('Configuration has not changed, no update needed');
       }
+
+      if (!validateConfig(newConfig)) {
```
fabiovincenzi: shoudn't we validate before emitting the event?

src/config/validators.ts
```diff
@@ -0,0 +1,69 @@
...
+      try {
+        new RegExp(pattern);
+      } catch (error: unknown) {
+        console.error(
+          `Invalid regular expression for commitConfig.diff.block.patterns: ${pattern}`,
+        );
+        return false;
+      }
+    }
+  }
```
fabiovincenzi, line 74: Maybe validate diff.block.providers as well?

src/config/ConfigLoader.ts
jescalada: Thanks for catching this! I also fixed up the related test that was suspiciously passing...

kriswest: +1

src/config/index.ts
kriswest: This can throw an error...

kriswest: The first step in `initializeConfigLoader` is to call `loadFullConfigration` - so this might be double loading...

src/config/validators.ts
jescalada: Thanks! I didn't notice the `diff.block.providers` object was using regex internally too.

src/config/index.ts
jescalada: @kriswest Any `getConfigValue` getter function could blow up since all of them call `loadFullConfiguration`. Should we perhaps catch these errors in `loadFullConfiguration` instead of bubbling them up? I don't think having try/catch blocks everywhere would help in this case...

jescalada: Good catch - fixed!

jescalada: By the way, `initializeConfigLoader` is wrapped in a try/catch when it's actually used (and it's the only usage):

```ts

// Initialize configuration on module load
try {
  initializeConfigLoader();
  console.log('Configuration loaded successfully');
} catch (error) {
  console.error('Failed to load configuration:', error);
  throw error;
}
```

`loadFullConfiguration` may require tweaks, however...

src/config/ConfigLoader.ts
kriswest: Could rename this variable, as the individual configs are only partially validated? Otherwise it might be better to move teh subsequent validation on line 185 into the load?

src/config/validators.ts
```diff
@@ -0,0 +1,80 @@
+import { GitProxyConfig } from './generated/config';
+
+const validationChain = [validateCommitConfig];
```
kriswest, line 3: You could move the Quicktype validation here, however, its often used to parse a string containing the config. You could have two entry points, one for a string and one for an object (stringify it before processing).

Alternatively, the input could always be a string and this validator could be used to parse it then valdiate?

src/config/validators.ts
```diff
@@ -0,0 +1,80 @@
...
+function validateCommitConfig(config: GitProxyConfig): boolean {
+  if (config.commitConfig?.author?.email?.local?.block) {
+    try {
+      new RegExp(config.commitConfig.author.email.local.block);
+    } catch (error: unknown) {
+      console.error(
+        `Invalid regular expression for commitConfig.author.email.local.block: ${config.commitConfig.author.email.local.block}`,
+      );
+      return false;
+    }
```
kriswest: This could be extracted to a helper function `validateConfigRegex(config: GitProxyConfig, path:string)` extracts the value (checking existence), checks if its an array, validates and logs. That would reduce repeated code.

src/config/validators.ts
jescalada: I managed to extract the error handling and parsing logic into a few helpers:

```ts
export async function loadConfig(
  context: string,
  loader: () => Promise<string>,
): Promise<GitProxyConfig> {
  const raw = await loader();
  return parseGitProxyConfig(raw, context);
}

function parseGitProxyConfig(
  raw: string,
  context: string
): GitProxyConfig {
  try {
    return Convert.toGitProxyConfig(raw);
  } catch (error) {
    throw new Error(`Invalid configuration format in ${context}: ${error instanceof Error ? error.message : 'Unknown error'}`);
  }
}
```

This allows injecting the unique loading logic into the parser which also handles errors.

src/config/validators.ts
jescalada: I tried doing this previously - had a bunch of failing tests so I gave up on it. Will take another look now with fresh eyes 🙂

src/config/ConfigLoader.ts
jescalada: I figured we could rename it instead of adding it to the validation chain, since the validation chain is meant to be applied to a single, merged `GitProxyConfig` object with proper formatting (although potentially erroneous entries). The earlier processes are either filtering or merging configs and this would happen only once - meanwhile validation should be extensible as we start finding more bugs with invalid config values.

src/config/validators.ts
jescalada: Managed to extract a few functions out of it - potentially useful for extending `commitConfig` checks or other regex validation down the line:

```ts
/**
 * Validates that a regular expression is valid.
 * @param pattern The regular expression to validate
 * @param context The context of the regular expression
 * @returns true if the regular expression is valid, false otherwise
 */
function isValidRegex(pattern: string, context: string): boolean {
  try {
    new RegExp(pattern);
    return true;
  } catch {
    console.error(`Invalid regular expression for ${context}: ${pattern}`);
    return false;
  }
}

/**
 * Validates that a value in the configuration is a valid regular expression.
 * @param config The configuration to validate
 * @param path The path to the value to validate
 * @returns true if the value is a valid regular expression, false otherwise
 */
function validateConfigRegex(config: GitProxyConfig, path: string): boolean {
  const getValueAtPath = (obj: unknown, path: string): unknown => {
    return path.split('.').reduce((current, key) => {
      if (current == null || typeof current !== 'object') {
        return undefined;
      }
      return (current as Record<string, unknown>)[key];
    }, obj);
  };

  const value = getValueAtPath(config, path);

  if (!value) return true;

  if (typeof value === 'string') {
    return isValidRegex(value, path);
  }

  if (Array.isArray(value)) {
    for (const pattern of value) {
      if (!isValidRegex(pattern, path)) return false;
    }
    return true;
  }

  if (typeof value === 'object') {
    return Object.values(value).every((pattern) => isValidRegex(pattern as string, path));
  }

  return true;
}
```

The main validator function turns into this:

```ts
function validateCommitConfig(config: GitProxyConfig): boolean {
  return (
    validateConfigRegex(config, 'commitConfig.author.email.local.block') &&
    validateConfigRegex(config, 'commitConfig.author.email.domain.allow') &&
    validateConfigRegex(config, 'commitConfig.message.block.patterns') &&
    validateConfigRegex(config, 'commitConfig.diff.block.patterns') &&
    validateConfigRegex(config, 'commitConfig.diff.block.providers')
  );
}
```

PR #1356, opened 2026-01-18
test: improve test coverage

jescalada:
Improves test coverage on route files and MongoDB handlers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
           fetch-depth: 0
 
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6.2.0
+        uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
         with:
           node-version: ${{ matrix.node-version }}
 
@@ -112,7 +112,7 @@ jobs:
           fetch-depth: 0
 
       - name: Use Node.js 24.x
-        uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6.2.0
+        uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
         with:
           node-version: 24.x
 
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -34,14 +34,14 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@b895512248b1b5b0089ac3c33ecf123c2cd6f373 # ratchet:github/codeql-action/init@v4
+        uses: github/codeql-action/init@a6594f96a3c88bcd1537795a61816854dc8ccf20 # ratchet:github/codeql-action/init@v4
         with:
           languages: ${{ matrix.language }}
 
       - name: Autobuild
-        uses: github/codeql-action/autobuild@b895512248b1b5b0089ac3c33ecf123c2cd6f373 # ratchet:github/codeql-action/autobuild@v4
+        uses: github/codeql-action/autobuild@a6594f96a3c88bcd1537795a61816854dc8ccf20 # ratchet:github/codeql-action/autobuild@v4
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@b895512248b1b5b0089ac3c33ecf123c2cd6f373 # ratchet:github/codeql-action/analyze@v4
+        uses: github/codeql-action/analyze@a6594f96a3c88bcd1537795a61816854dc8ccf20 # ratchet:github/codeql-action/analyze@v4
         with:
           category: '/language:${{matrix.language}}'
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -20,7 +20,7 @@ jobs:
 
       - name: Log in to Docker Hub
         if: github.repository == 'finos/git-proxy'
-        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3
+        uses: docker/login-action@b45d80f862d83dbcd57f89517bcf500b2ab88fb2 # v4
         with:
           username: finos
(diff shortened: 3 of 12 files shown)


test/testRepoApi.test.ts
```diff
@@ -298,3 +298,265 @@ describe('add new repo', () => {
...
+
+  beforeAll(async () => {
+    proxy = new Proxy();
+    app = await Service.start(proxy);
+
+    await cleanupRepo(TEST_REPO.url);
+    await db.deleteUser('testadmin');
+    await db.deleteUser('nonadmin');
+
+    await db.createUser('testadmin', 'password', 'testadmin@test.com', 'Test Admin', true);
```
fabiovincenzi: testadmin is created here but seems to be never used if needed we could also delete in the afterAll block, which only cleans up testuser and nonadmin.

test/db/mongo/push.test.ts
```diff
@@ -0,0 +1,292 @@
...
+            commitData: 1,
+            commitFrom: 1,
+            commitTo: 1,
+            error: 1,
+            method: 1,
+            project: 1,
+            rejected: 1,
+            repo: 1,
+            repoName: 1,
+            timepstamp: 1,
```
fabiovincenzi: Nit: this replicates the pre-existing typo timepstamp (instead of timestamp) from src/db/mongo/pushes.ts. We could fix the typo in both the source and the test.

test/testRepoApi.test.ts
jescalada: Whoops, didn't notice this one - probably confused it with the built-in `admin`

test/db/mongo/push.test.ts
jescalada: Code completion gone wrong 😄 Makes me wonder how many other typos like these we have around the app...

test/db/mongo/helper.test.ts
```diff
@@ -0,0 +1,315 @@
...
+
+      vi.clearAllMocks();
+      mockDb.collection.mockReturnValue(mockCollection);
+
+      await connect('collection2');
+
+      expect(MongoClient).not.toHaveBeenCalled();
+      expect(mockClient.connect).not.toHaveBeenCalled();
+      expect(mockDb.collection).toHaveBeenCalledWith('collection2');
+    });
```
andypols, line 99: This isn't wrong but may be clearer to use something like:

```
await connect('collection1');
expect(MongoClient).toHaveBeenCalledTimes(1);
expect(mockClient.connect).toHaveBeenCalledTimes(1);

await connect('collection2');
expect(MongoClient).toHaveBeenCalledTimes(1);
expect(mockClient.connect).toHaveBeenCalledTimes(1);
expect(mockDb.collection).toHaveBeenLastCalledWith('collection2');
```

jescalada: There's something odd going on here with the database connection not closing properly... I believe it's due to the mock bypassing the connection logic.

I rewrote the test based on the `should connect to MongoDB` one above so it works with your valid assertions: 

```ts
it('should reuse existing connection', async () => {
  mockGetDatabase.mockReturnValue({
    connectionString: 'mongodb://localhost:27017/testdb',
    options: {},
  });

  const { connect } = await import('../../../src/db/mongo/helper');

  const result = await connect('collection1');

  expect(MongoClient).toHaveBeenCalledWith('mongodb://localhost:27017/testdb', {});
  expect(mockClient.connect).toHaveBeenCalledTimes(1);
  expect(mockClient.db).toHaveBeenCalledTimes(1);
  expect(mockDb.collection).toHaveBeenCalledWith('collection1');
  expect(result).toBe(mockCollection);

  // Accessing a different collection should reuse the existing db connection
  await connect('collection2');
  expect(MongoClient).toHaveBeenCalledTimes(1);
  expect(mockClient.connect).toHaveBeenCalledTimes(1);
  expect(mockClient.db).toHaveBeenCalledTimes(1);
  expect(mockDb.collection).toHaveBeenCalledWith('collection2');
});
```

test/testRepoApi.test.ts
```diff
@@ -404,3 +404,263 @@ describe('add new repo', () => {
...
+      const res = await request(app)
+        .get('/api/v1/repo')
+        .set('Cookie', adminCookie)
+        .query({ '': 'empty' });
+
+      expect(res.status).toBe(200);
+    });
+  });
+
+  afterAll(async () => {
```
dcoric, line 660: Teardown is duplicated in one suite. There is `afterAll` around line 115.
I’d recommend consolidating to one afterAll (or split into two top-level describe blocks if you want separate teardown scopes).
This can create a stability risk and can cause nondeterministic failures in integration tests.

test/testPush.test.ts
```diff
@@ -213,153 +378,292 @@ describe('Push API', () => {
...
+            ],
+          },
+        });
+
+      expect(res.status).toBe(403);
+      expect(res.body.message).toBe(
+        `User ${testUsername} not authorised to approve pushes on this project`,
+      );
+
+      // Cleanup
```
dcoric: Not the best way to do a cleanup. If assertion fails, this will not be executed.
Prefer `try/finally` or centralized `afterEach` cleanup

test/testPush.test.ts
```diff
@@ -213,153 +378,292 @@ describe('Push API', () => {
...
-    testPush.userEmail = TEST_EMAIL_1;
-    await db.writeAudit(testPush as any);
-    await loginAsApprover();
-    const res = await request(app)
-      .post(`/api/v1/push/${TEST_PUSH.id}/reject`)
-      .set('Cookie', `${cookie}`);
-    expect(res.status).toBe(403);
-    expect(res.body.message).toBe('Cannot reject your own changes');
-  });
+      // Cleanup
```
dcoric: Same issue as on 513 line - 
Prefer `try/finally` or centralized `afterEach` cleanup in a case assertion fails, and this part does not get executed

dcoric (changes requested):
Great coverage improvement overall. Added small comments for double teardown and to reduce flakiness if an assertion fails mid-test.

test/testPush.test.ts
jescalada: Thanks for the suggestion! I've reverted the state of `testPush` since the merge conflicts were too confusing to deal with. I'll keep your suggestion in mind when adding a few tests for extra coverage here.

test/testRepoApi.test.ts
jescalada: You're right, it seems `testPush` had this problem! 👍🏼

Issue #1294, opened 2025-11-28
Failing proxy tests (test cleanup issues)


jescalada:
After the Vitest migration in #1202, we're skipping a few tests that call `proxy.start`/`proxy.stop` due to some kind of race condition that makes them fail depending on execution order.

This is the list of problematic test files/tests:

- All tests in `proxy.test.ts` (3)
- "proxy express application" tests in `testProxyRoute.test.ts` (8)

Running the `proxy.test.ts` file results in the following error:

```bash
Serialized Error: { code: 'EADDRINUSE', errno: -98, syscall: 'listen', address: '::', port: 8000 }
This error originated in "test/testProxyRoute.test.ts" test file. It doesn't mean the error was thrown inside the file itself, but while it was running.
The latest test that might've caused the error is "should restart and proxy for a new host when project is ADDED". It might mean one of the following:
- The error was thrown, while Vitest was running this test.
- If the error occurred after the test had been completed, this was the last documented test before it was thrown.

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Uncaught Exception ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Error: listen EADDRINUSE: address already in use :::8000
 ❯ Server.setupListenHandle [as _listen2] node:net:1937:16
 ❯ listenInCluster node:net:1994:12
 ❯ Server.listen node:net:2099:7
 ❯ Proxy.start src/proxy/index.ts:76:8
     74|     this.httpServer = http
     75|       .createServer(getServerOptions() as any, this.expressApp)
     76|       .listen(proxyHttpPort, () => {
       |        ^
     77|         console.log(`HTTP Proxy Listening on ${proxyHttpPort}`);
     78|       });
 ❯ processTicksAndRejections node:internal/process/task_queues:105:5
 ❯ src/service/routes/repo.ts:135:9
```

Meanwhile, running all of the files results in varying errors depending on test execution order. When re-running all tests locally, they pass without fail. The first test run always fails in the CI.

[...]

PR #1357, opened 2026-01-19
fix: enable skipped proxy tests

fabiovincenzi:
Fixes: #1294

  While investigating the issue, I noticed the `stop()` method in `src/service/index.ts` was not properly waiting for the server to close:

  ```typescript
  // BEFORE
  async function stop() {
    console.log(`Stopping Service Listening on ${uiPort}`);
    _httpServer.close();  // Async call, but we don't wait for it!
  }
  ```
  This is a potential race condition, so I changed it to resolve only when the callback of `_httpServer.close()` is called

  Also changed `_httpServer` to be created in `start()` instead of at module level, making the Service restartable.

  However, this was not the cause of the failing tests.

## proxy.test.ts

  This test mocks the HTTP/HTTPS servers, so it doesn't use the real `stop()`. The issue was in the close mock wich didn't call the callback, the real `stop()` method wraps server.close() in a Promise that only resolves when the callback is called:

  ```typescript
  // BEFORE (buggy mock)
  vi.spyOn(https, 'createServer').mockReturnValue({
    listen: vi.fn().mockReturnThis(),
    close: vi.fn(),  // Never calls callback!
  } as any);
  ```

  If the mock's `close()` never calls the callback, the Promise never resolves, and the test hangs until it times out.

  **Fix**: Mock both http and https, using the pre-defined `mockHttpServer`/`mockHttpsServer` that properly call callbacks

## testProxyRoute.test.ts 

  After enabling testProxyRoute.test.ts, users created in testPush.test.ts were "NOT FOUND" during passport deserialization.

  Root cause:
  - `testProxyRoute.test.ts` has `afterAll(() => { vi.resetModules(); })`
  - This creates a new instance of the in-memory database
  - Passport callbacks captured a reference to the old db module at configuration time
  - New users were created in the new database, but passport looked in the old one

  **Fix**: Use dynamic import in passport callbacks (src/service/passport/local.ts):
  
 Note: This PR fixes the immediate issue, but we should consider https://github.com/finos/git-proxy/issues/1294#issuecomment-3597760233 from @coopernetes about refactoring the test architecture.


src/service/index.ts
```diff
@@ -20,7 +20,7 @@ const { GIT_PROXY_UI_PORT: uiPort } = serverConfig;
 const DEFAULT_SESSION_MAX_AGE_HOURS = 12;
 
 const app: Express = express();
-const _httpServer = http.createServer(app);
+let _httpServer: http.Server | null = null;
```
jescalada, line 23: It makes a lot of sense to move HTTP server initialization into the `start` function! 👍🏼

Wonder if there was a specific reason why it was initialized on definition previously.

test/proxy.test.ts
```diff
@@ -82,12 +83,17 @@ describe.skip('Proxy Module TLS Certificate Loading', () => {
...
-    }));
+    vi.doMock('../src/db', async (importOriginal) => {
+      const actual: any = await importOriginal();
+      return {
+        ...actual,
+        getRepos: mockDb.getRepos,
+        createRepo: mockDb.createRepo,
+        addUserCanPush: mockDb.addUserCanPush,
+        addUserCanAuthorise: mockDb.addUserCanAuthorise,
+        getAllProxiedHosts: vi.fn().mockResolvedValue([]),
```
jescalada, line 94: Oddly, tests still pass on my end after removing the `getAllProxiedHosts` mock. Any specific reason to include mock this function?

src/service/index.ts
fabiovincenzi: The httpServer was originally exposed at module level in e95949cc to allow closing it in tests (see the after hook in test/testLogin.js from the same commit). The initialization was probably done at the module level just for simplicity at the time.
I don't think moving it inside start() will causes any issues

test/proxy.test.ts
fabiovincenzi: Interesting — tests fail on my end without it (vitest throws "No 'getAllProxiedHosts' export is defined on the mock").   
  Could be a caching issue or different vitest/node versions. Could you try clearing the cache first? (rm -rf                
  node_modules/.vitest && npm test)

fabiovincenzi:
Created https://github.com/finos/git-proxy/pull/1378 to make the tests runnable in random order.

PR #1316, opened 2025-12-09
fix: defer reading of database configuration until needed to fix race

kriswest:
resolves #1313
supercedes #1314 

By deferring the setup of the database adaptors until they are first used, we can defer the retrieval of the DB config allowing the user configuration to be loaded before we attempt to use it.

PR #1358, opened 2026-01-20
fix: file permission issue in e2e tests & regression on db init

### LINKED ISSUE #1354: Fix broken build — E2E tests failing due to missing local git repo
Hey folks — can we get the build green again? It hurts my soul a little seeing it red 😄

The failures are coming from the E2E test suite. All failing tests appear to hit the same error:

```
fatal: repository 'http://localhost:8000/git-server:8443/coopernetes/test-repo.git/' not found
```

This looks like the test environment expects a local git server/repo that isn’t available (or the URL is wrong). I’m not sure whether this is a regression/bug or if these tests were never actually working.

**Possible paths forward:**

- Fix the E2E setup so the expected repo exists (or correct the URL/config).
- If these tests are obsolete or flaky, we can temporarily mark them as ignored/skipped while we sort out the underlying cause.
- Remove them if they’re no longer relevant (last resort).

@coopernetes would you mind taking a look / pointing me at the intended setup for the git server + test repo?

Thanks!

coopernetes:
Closes #1354

Two fixes to e2e tests as well as a regression caused by the app
initialization.
- `pullRemote - Error: EACCES: permission denied, mkdir './.remote'` errors during e2e test runs (Dockerfile fix of file permissions)
- Fix issue where `getAllProxiedHosts()` does not complete until after the Express router starts up, leading the `git-server:8443` origin to _not_ be proxied (regression from #1316) which causes the proxy to default to github.com. The following logs are observed in the git-proxy container:

```
$ docker compose logs git-proxy
Created runtime configuration with:
  API URL: auto-detect
  Allowed Origins: *
  Environment: test
Configuration sources are disabled
Configuration loaded successfully
Setting config file to: /app/test-e2e.proxy.config.json
validating config
Setting up the proxy and Service
No plugins configured
Found 0 plugin modules
Loading MongoDB database adaptor
creating new repo {"project":"coopernetes","name":"test-repo","url":"https://git-server:8443/coopernetes/test-repo.git","users":{"canPush":[],"canAuthorise":[]}}
creating new repo {"project":"finos","name":"git-proxy","url":"https://git-server:8443/finos/git-proxy.git","users":{"canPush":[],"canAuthorise":[]}}
Initializing proxy router for origins: '[]'
        setting up catch-all route (github.com) for backwards compatibility
        
       # !!! Note how git-server:8443 is never added as a proxied origin ^
...
checkRepoInAuthorisedList - repo https://git-server:8443/coopernetes/test-repo.git is in the authorisedList
Action processed: Allowed
    Request URL: /git-server:8443/coopernetes/test-repo.git/info/refs?service=git-upload-pack
    Host:        localhost:8000
    User-Agent:  git/2.52.0
Request resolved to https://github.com/git-server:8443/coopernetes/test-repo.git/info/refs?service=git-upload-pack

       # !!! Request incorrectly proxied to github.com instead of git-server:8443 ^

```

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -9,21 +9,13 @@ on:
   push:
     branches: [main]
   pull_request:
-  issue_comment:
-    types: [created]
+  workflow_dispatch:
 
 jobs:
   e2e:
     runs-on: ubuntu-latest
-    # Run on push/PR or when a maintainer comments "/test e2e" or "/run e2e"
-    if: |
-      github.event_name != 'issue_comment' || (
-        github.event.issue.pull_request &&
-        (contains(github.event.comment.body, '/test e2e') || contains(github.event.comment.body, '/run e2e')) &&
-        (github.event.comment.author_association == 'OWNER' || 
-         github.event.comment.author_association == 'MEMBER' || 
-         github.event.comment.author_association == 'COLLABORATOR')
-      )
+    # Only run on workflow_dispatch (manual trigger), but always register check on PRs
+    if: github.event_name == 'workflow_dispatch'
 
     steps:
       - name: Checkout code


src/proxy/index.ts
```diff
@@ -51,14 +51,14 @@ export class Proxy {
     const defaultAuthorisedRepoList = getAuthorisedList();
     const allowedList: Repo[] = await getRepos();
 
-    defaultAuthorisedRepoList.forEach(async (x) => {
```
coopernetes, line 54: FYI @kriswest 

This forEach is not "blocking" in that these functions are not awaited, causing this function to return early and for the app to continue starting up without the list of repositories added to the database. This causes the proxy middleware instances to not be created (due to an "empty" list of repository origins). It then falls back to proxying only https://github.com.

- `proxyPreparations()` is awaited before `createApp()`
- DB seeding should finish first. 
- However, there is a subtle bug: inside `proxyPreparations()`, it uses forEach(async (x) => { ... }) for seeding, but forEach does not await async callbacks. This means proxyPreparations() can finish before all repos are actually created.

Issue #1318, opened 2025-12-11
Fix failing tests on Windows


jescalada:
Various tests are failing on Windows, likely due to differences in file paths:

<img width="740" height="102" alt="Image" src="https://github.com/user-attachments/assets/889c0151-8365-483c-a13a-301a9e49740a" />

Here's the complete output on my end:

<details>

```bash
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Failed Suites 1 ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 FAIL  test/plugin/plugin.test.ts > loading plugins from packages
Error: ENOENT: no such file or directory, lstat 'C:\Users\mahye\git-proxy\test\fixtures\test-package\node_modules'
 ❯ test/plugin/plugin.test.ts:123:5
    121|
    122|   afterAll(() => {
    123|     rmSync(join(testPackagePath, 'node_modules'), { recursive: true });
       |     ^
    124|   });
    125| });

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/13]⎯


[...]

PR #1361, opened 2026-01-23
ci: add Windows to test matrix and fix cross-platform compatibility

fabiovincenzi:
Fixes: https://github.com/finos/git-proxy/issues/1318.
This PR adds Windows to the CI test matrix and fixes cross-platform compatibility issues.                                
                                                                                                                           
  CI Changes                                                                                                               
                                                                                                                           
  - Add windows-latest to the OS matrix                                                   
  - Enable Windows Developer Mode in CI for npm symlinks support                                                           
  - Skip Cypress e2e tests on Windows                           
  - Remove unnecessary MongoDB dependency and version matrix (see below)                                                   
                                                                                                                           
  Cross-Platform Fixes                                                                                                     
                                                                                         
  - Replace bash script fix-shebang.sh with Node.js fix-shebang.js            
  - Add cross-env package for Windows-compatible environment variables in npm scripts                                      

 The preReceive tests are skipped because they test Unix shell hook execution, which is not applicable on Windows.

### DIFF
--- a/src/ui/components/Navbars/DashboardNavbarLinks.tsx
+++ b/src/ui/components/Navbars/DashboardNavbarLinks.tsx
@@ -27,11 +27,10 @@ const DashboardNavbarLinks: React.FC = () => {
   const [openProfile, setOpenProfile] = useState<HTMLElement | null>(null);
   const [, setAuth] = useState<boolean>(true);
   const [, setIsLoading] = useState<boolean>(true);
-  const [errorMessage, setErrorMessage] = useState<string>('');
   const [user, setUser] = useState<PublicUser | null>(null);
 
   useEffect(() => {
-    getUser(setIsLoading, setUser, setAuth, setErrorMessage);
+    getUser(setIsLoading, setUser, setAuth);
   }, []);
 
   const handleClickProfile = (event: React.MouseEvent<HTMLElement>) => {
@@ -66,7 +65,6 @@ const DashboardNavbarLinks: React.FC = () => {
 
   return (
     <div>
-      {errorMessage && <div className={classes.errorMessage}>{errorMessage}</div>}
       <div className={classes.manager}>
         <Button
           color={window.innerWidth > 959 ? 'transparent' : 'white'}
--- a/src/ui/views/User/UserProfile.tsx
+++ b/src/ui/views/User/UserProfile.tsx
@@ -16,6 +16,7 @@ import { LogoGithubIcon } from '@primer/octicons-react';
 import CloseRounded from '@material-ui/icons/CloseRounded';
 import { Check, Save } from '@material-ui/icons';
 import { TextField, Theme } from '@material-ui/core';
+import Danger from '../../components/Typography/Danger';
 
 const useStyles = makeStyles((theme: Theme) => ({
   root: {
@@ -52,11 +53,13 @@ export default function UserProfile(): React.ReactElement {
   }, [id]);
 
   if (isLoading) return <div>Loading...</div>;
-  if (errorMessage) return <div>{errorMessage}</div>;
 
   if (!auth && window.location.pathname === '/dashboard/profile') {
     return <Navigate to='/login' />;
   }
+
+  if (errorMessage) return <Danger>{errorMessage}</Danger>;
+
   if (!user) return <div>No user data available</div>;
 
   const updateProfile = async (): Promise<void> => {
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -5,7 +5,7 @@ name: CI
 
 on:
   push:
-    branches: [main]
+    branches: [main, windows-failing-tests]
```
jescalada: We might want to revert this if not testing anymore!

```suggestion
    branches: [main]
```

.github/workflows/ci.yml
```diff
@@ -37,10 +37,11 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
-      - name: Start MongoDB
```
jescalada, line 40: Any specific reason to remove MongoDB from the test matrix and steps?

jescalada: I noticed that we already have 6 CI combinations (20, 22, 24) x (Windows, Ubuntu), so adding the 3 Mongo versions would result in 18 CI runs, spending way too many resources in CI runs.

Perhaps we could keep the CI execution as it is, and then add a `ci-windows.yml` file separately? Not sure if this makes the most sense - Windows is not really meant to be a "supported" OS beyond development since GitProxy would usually be deployed on a Unix machine.

test/preReceive/preReceive.test.ts
```diff
@@ -6,7 +6,7 @@ import { exec } from '../../src/proxy/processors/push-action/preReceive';
 // TODO: Replace with memfs to prevent test pollution issues
 vi.mock('fs', { spy: true });
 
-describe('Pre-Receive Hook Execution', () => {
+describe.skipIf(process.platform === 'win32')('Pre-Receive Hook Execution', () => {
```
jescalada, line 10: Might as well add a warning here and in `prereceive.ts` in the rare case this is being used on Windows 🤔 

```suggestion
// Pre-receive is not currently supported on Windows
describe.skipIf(process.platform === 'win32')('Pre-Receive Hook Execution', () => {
```

test/processors/writePack.test.ts
```diff
@@ -66,14 +68,14 @@ describe('writePack', () => {
         1,
         'git',
         ['config', 'receive.unpackLimit', '0'],
-        expect.objectContaining({ cwd: '/path/to/repo' }),
+        expect.objectContaining({ cwd: path.join('path', 'to', 'repo') }),
```
jescalada, line 71: 👍🏼

jescalada (commented):
The tests look good!

Wondering if we could keep the MongoDB combinations in the CI and think about whether we need a Windows CI execution since Windows test support is more of a quality-of-life thing for Windows development...

.github/workflows/ci.yml
fabiovincenzi: MongoDB was originally added to the CI in November 2020 (f8d5f9f4) when it was the default database.                                                               
                                                                                                                           
However, at some point the default was changed to use the file-based database, and now MongoDB is `enabled: false` by default in proxy.config.json. Additionally, when NODE_ENV=test, the tests use neDB in-memory (see `src/db/file/*.ts` with inMemoryOnly: true), so they never actually connect to MongoDB.                                                     
                                                                                                                           
This means MongoDB has been running in CI without being used by tests for quite some time.

Regarding Windows: I agree it's more of a quality-of-life feature for development. Having 6 combinations (3 Node × 2 OS) seems reasonable to me, but I'm open to moving Windows to a separate workflow if preferred. What do you think?

.github/workflows/ci.yml
fabiovincenzi: You are right used for testing

.github/workflows/ci.yml
jescalada: The MongoDB CI matrix was added in #1315, since we want to check if the build still works with various recent versions. We do have a few tests already for Mongo, and we're adding more in #1356. Plus we want to support E2E testing soon as well.

I think the OS combinations might not be as necessary as the Mongo ones - we could have a single separate execution for Windows on Node 24 just for Windows development support.

CC: @coopernetes for a better explanation 🙂

fabiovincenzi: Okay, thanks for explaining this! I'll revert these changes🙂

jescalada (approved):
LGTM! 👍🏼

Issue #1362, opened 2026-01-26
Tests fail when shuffling test execution


jescalada:
When shuffling test execution, a number of tests are flaky or outright fail:

<img width="1219" height="139" alt="Image" src="https://github.com/user-attachments/assets/4379b8e1-b504-4d15-a449-4f622de30667" />

A lot of them seem to be related to `testRepoApi.test.ts`:

<img width="1230" height="872" alt="Image" src="https://github.com/user-attachments/assets/10f57bc6-45eb-43d2-86ca-ca6893ef990e" />

Other files that typically error out include: `testLogin`, `testDb`.

**Describe the solution you'd like**
Fix the flaky tests so that they pass even when shuffling the test execution order via
```json
"test-shuffle": "NODE_ENV=test vitest --run --dir ./test --sequence.shuffle",
```

PR #1378, opened 2026-02-02
fix: make tests independent of execution order

fabiovincenzi:
Fixes: https://github.com/finos/git-proxy/issues/1362.
Make tests independent of execution order so they pass with `--sequence.shuffle`. The root causes were shared mutable state across tests running in a single process: DB records, mock leaks, module cache, network ports, and filesystem artifacts.                                                                                                
  - **Fix fire-and-forget async operations in production code:**                                                               
    - clearBareClone.ts: `fs.rm()` with callback was not awaited, causing race conditions. Switched to await `fs.promises.rm()`.                                                                                                       
    - Proxy.start(): `server.listen()` was not awaited and errors were silently ignored. Wrapped in a Promise with proper error handling.                                                                                                          
    - Service.stop(): `_httpServer` was not reset to null after close, preventing clean restarts.                            
    - repo.ts POST route: proxy restart happened after `res.send()`, so clients received `200` before the proxy was ready. Moved restart before the response, consistent with the DELETE route.                                                     
  - **Fix mock/spy leaks between tests:**                                                                                      
    - extractRawBody.test.ts: moved `mockClear()` from `afterAll` to `beforeEach`.                                               
    - testAuthMethods.test.ts: added `vi.doUnmock('fs')` before `vi.resetModules() `since `doMock` persists across module resets.
    - preReceive.test.ts: replaced `vi.restoreAllMocks()` with `vi.clearAllMocks()` to preserve module-level spies.            
    - 1.test.ts: added `vi.resetModules()` before `vi.doMock` to ensure fresh imports.                                         
    - proxy.test.ts: added missing `on()` method to mock HTTP servers.                                                       
  - **Make each test self-contained (no implicit ordering dependencies):**                                                     
    - testDb.test.ts: each test now creates its own prerequisites and cleans up conflicting state.                         
    - testLogin.test.ts: extracted `loginAsAdmin()` helper so each test gets its own session cookie.                         
    - testRepoApi.test.ts: moved login to `beforeAll`, added `ensureTestRepoExists()` helper, proper cleanup with `proxy.stop() ` and  `Service.stop()`.                                                                                                        
    - testProxyRoute.test.ts: each test ensures its own DB prerequisites exist.                                            
  - **Fix filesystem state dependencies:**                                                                                     
    - ConfigLoader.test.ts: clean cached git clone directory before test to force git clone over git pull.                 
    - getDiff.test.ts: each test creates its own git commits with unique filenames.

### DIFF
--- a/test/ConfigLoader.test.ts
+++ b/test/ConfigLoader.test.ts
@@ -432,7 +432,7 @@ describe('ConfigLoader', () => {
       );
     });
 
-it(
+    it(
       'should throw error if repository is a valid URL but not a git repository',
       async () => {
         const source: ConfigurationSource = {


test/testDb.test.ts
```diff
@@ -239,6 +262,10 @@ describe('Database clients', () => {
   });
 
   it('should be able to retrieve a repo by id', async () => {
+    // Ensure repo exists
+    const existing = await db.getRepoByUrl(TEST_REPO.url);
+    if (!existing) await db.createRepo(TEST_REPO);
```
jescalada: Would be good to extract this into its own helper function since we're repeating this a lot throughout the file.

test/testDb.test.ts
```diff
@@ -353,6 +391,18 @@ describe('Database clients', () => {
   });
 
   it('should throw an error when creating a duplicate username', async () => {
+    // Ensure user exists
+    const existingUser = await db.findUser(TEST_USER.username);
+    if (!existingUser) {
```
jescalada: Same as the existing repo check - good to extract into its own function!

It might even be worth looking into making the helper function work for all test files since the `TEST_USER` and `TEST_REPO` variables are often very similar.

test/testDb.test.ts
```diff
@@ -527,57 +645,91 @@ describe('Database clients', () => {
...
+      await db.writeAudit(TEST_PUSH as any);
+    }
     await db.deletePush(TEST_PUSH.id);
     const pushes = await db.getPushes({});
     const cleanPushes = cleanResponseData(TEST_PUSH, pushes as any);
     expect(cleanPushes).not.toContainEqual(TEST_PUSH);
   });
 
   it('should be able to authorise a push', async () => {
+    await db.deletePush(TEST_PUSH.id);
```
jescalada: Is there a reason why this can't be run in a `beforeEach()` for all the push-related tests?

test/testRepoApi.test.ts
```diff
@@ -184,18 +240,36 @@ describe('add new repo', () => {
...
 
   it('add authorise user that does not exist', async () => {
+    await ensureTestRepoExists();
+
+    // Ensure 2 authorise users exist
+    const repoBefore = await fetchRepoOrThrow(TEST_REPO.url);
+    if (!repoBefore.users.canAuthorise.includes('u1'))
+      await db.addUserCanAuthorise(repoIds[0], 'u1');
+    if (!repoBefore.users.canAuthorise.includes('u2'))
+      await db.addUserCanAuthorise(repoIds[0], 'u2');
```
jescalada: Might be good to extract these too for `testRepoApi.test.ts`

test/testDb.test.ts
fabiovincenzi: Done, extracted `ensureRepoExists()`

test/testDb.test.ts
fabiovincenzi: extracted `ensureUserExists()`

test/testRepoApi.test.ts
fabiovincenzi: extracted `ensureUsersCanPush(users)` and `ensureUsersCanAuthorise(users)`.

test/testDb.test.ts
fabiovincenzi: You're right, fixed

test/testDb.test.ts
fabiovincenzi: Only `testDb` and `testRepoApi` use these patterns. They have different constants and different setup strategies (direct DB vs HTTP API). Also testRepoApi doesn't repeat this pattern, it sets up users once in beforeAll.  SoI would leave them  as-is for now and revisit if other test files start reproducing this pattern.

jescalada: Makes sense 👍🏼

PR #1381, opened 2026-02-04
chore: add CODEOWNERS with basic setup

jescalada:
Sets up a basic `CODEOWNERS` file.

@finos/git-proxy-maintainers  Feel free to make these more granular! 👍🏼

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,17 +1,2 @@
 # Default — maintainers own everything
 * @finos/git-proxy-maintainers
-
-# Sub-areas (must ALSO include maintainers!)
-/plugins/ @finos/git-proxy-maintainers @coopernetes
-
-/src/proxy/processors @finos/git-proxy-maintainers @kriswest @grovesy
-
-/src/service/passport @finos/git-proxy-maintainers @jescalada
-/src/service/passport/activeDirectory.ts @finos/git-proxy-maintainers @06kellyjac @kriswest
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,17 @@
+# Default — maintainers own everything
+* @finos/git-proxy-maintainers
+
+# Sub-areas (must ALSO include maintainers!)
+/plugins/ @finos/git-proxy-maintainers @coopernetes
```
coopernetes: This is a logical OR (not AND). An approval from anyone in the maintainer team will satisfy the CODEOWNERS check. As such, I wonder if it's worth carving up the paths in this way considering that all the individuals in the FINOS team are the same mentioned in specific paths.

I think we can simply use the first line in the file only then handle specific parts of the code base via PR assignee or review requests.

We likely won't ever have committers with `write` access that _aren't_ in the team.

kriswest: Fair enough. All this is achieving is review assignments that will already overlap with @finos/git-proxy-maintainers.

@jescalada if you're ok with keeping it simple we can apply that to the repo and I'll contact FINOS to make the settings changes for:

- Require review from Code Owners
- Require an approval from someone other than the last person to push to a branch

on the main branch.

jescalada: This sounds good to me - since it's mostly meant to protect the repo from self-merging and prevent us from missing fresh PRs. 🙏🏼

PR #1382, opened 2026-02-04
feat: Add MongoDB integration tests with CI support

### LINKED ISSUE #1168: Set-up testing of the mongo DB client
One or more recent PRs were merged with errors in the mongoDB client implementation, which is currently not used when testing and has little test coverage (31.46% vs 91.41% for the file DB client, see https://app.codecov.io/gh/finos/git-proxy/tree/main/src%2Fdb)

At least the DB tests are repeated in CI against the mongo DB client. That will require a mongoDB instance to be stood up during the test (that can't be mocked and produce a valid result AFAIK).

**Additional context**
I note the CI job does start a Mongo DB instance, but it is not then used when testing... See: 
https://github.com/finos/git-proxy/blob/4bf227674c78b1c19a08ad150ea5a949d2e007bb/.github/workflows/ci.yml#L39-L42

dcoric:
### Summary

Adds comprehensive MongoDB integration tests for the database layer (repos, users, pushes) with CI integration. Tests run against a real MongoDB instance to validate database operations beyond unit test coverage.
Fixes #1168

### What's New

- Adds integration test coverage for MongoDB operations
- Test infrastructure with automatic setup/teardown and database cleanup
- CI integration running tests across MongoDB 6.0, 7.0, and 8.0
- Enhanced database helper with connection management utilities
- Documentation for running integration tests locally with Docker

### Running Locally

Integration tests require MongoDB running on `localhost:27017`. The connection string is configured in `vitest.config.integration.ts`.

```bash
# Start MongoDB with Docker
docker run -d --name mongodb-test -p 27017:27017 mongo:7

# Run integration tests
npm run test:integration

# Cleanup
docker stop mongodb-test && docker rm mongodb-test
```

The `RUN_MONGO_TESTS` environment variable is automatically set by the vitest config, but can be overridden if needed.

### Running in CI

Integration tests run automatically in the CI pipeline after unit tests. MongoDB is already started via `supercharge/mongodb-github-action` in the build matrix (versions 6.0, 7.0, 8.0), so the tests execute against the existing MongoDB instance.

The CI step simply runs:
```yaml
- name: MongoDB Integration Tests
  env:
    RUN_MONGO_TESTS: 'true'
  run: npm run test:integration
```

### Configuration

Connection string is set in `vitest.config.integration.ts`:
- **Local/CI**: `mongodb://localhost:27017/git-proxy-test`
- **Database**: `git-proxy-test` (isolated from development data)
- **Config file**: `test-integration.proxy.config.json` (loaded via `CONFIG_FILE` env var)

### Benefits

- Validates real database operations against live MongoDB
- Runs automatically in CI across all MongoDB versions in the test matrix
- Uses isolated test database to avoid conflicts with development data
- Clean test isolation with automatic cleanup between tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,6 @@
     "test": "cross-env NODE_ENV=test vitest --run --dir ./test",
     "test:e2e": "vitest run --config vitest.config.e2e.ts",
     "test:e2e:watch": "vitest --config vitest.config.e2e.ts",
-<<<<<<< ours
     "test-coverage": "cross-env NODE_ENV=test vitest --run --dir ./test --coverage",
     "test-coverage-ci": "cross-env NODE_ENV=test vitest --run --dir ./test --coverage.enabled=true --coverage.reporter=lcovonly --coverage.reporter=text",
     "test:integration": "NODE_ENV=test vitest --run --config vitest.config.integration.ts",


grovesy:
@kriswest - Reviewed, looks good to me. Happy to close.

-Paul

test/setup-integration.ts
```diff
@@ -0,0 +1,90 @@
...
+    const dbName = getTestDbName(
+      process.env.GIT_PROXY_MONGO_CONNECTION_STRING ||
+        `mongodb://localhost:27017/${DEFAULT_TEST_DB_NAME}`,
+    );
+    const db = client.db(dbName);
+    for (const collection of COLLECTIONS) {
+      try {
+        await db.collection(collection).deleteMany({});
+      } catch {
+        // Collection might not exist yet, ignore
```
jescalada: Is it okay to have this fail silently? Just wondering if there are any edge cases where `deleteMany` could fail and leave the subsequent test with a polluted database.

test/setup-integration.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  if (client) {
+    try {
+      const dbName = getTestDbName(
+        process.env.GIT_PROXY_MONGO_CONNECTION_STRING ||
+          `mongodb://localhost:27017/${DEFAULT_TEST_DB_NAME}`,
+      );
+      await client.db(dbName).dropDatabase();
+    } catch {
+      // Ignore if database doesn't exist
```
jescalada: Same here and the other `catch` clauses above - should we be showing some kind of debug message in case of failure?

test/setup-integration.ts
dcoric: Yes, you are correct. `deleteMany` returns 0 if there are no data to delete, and fails if there is more serious problem, like DB not reachable, connection issues, auth or other problems.
I have fixed it to display error

test/setup-integration.ts
dcoric: Added error logging and applied to remaining `catch` clauses

PR #1335, opened 2025-12-30
docs: architecture guide

jescalada:
Closes #1325.

Feel free to add anything that's missing! Let me know if you'd like me to split a section into its own page.

The architecture diagram can be edited from this file on [diagrams.net](https://app.diagrams.net/)
[GitProxy_Architecture.drawio](https://github.com/user-attachments/files/24558301/GitProxy_Architecture.drawio)

### To-do
- [x] Add section for authentication methods
- [x] Add section for databases
- [x] Add plugin guide
- [x] Add deployment guide - to be added later by @tabathad 
- [x] Fix all the to-dos

Issue #1325, opened 2025-12-15
Document internal architecture, contributing & hacking on Git Proxy


coopernetes:
A follow-up from #425 and document the current architecture of Git Proxy in preparation for a full 2.0 release.

Some details which should be included:

- Architecture & internals
- Extending Git Proxy (plugins, already documented in #638 and just needs to be validated against main for v2.0)
- Deploying Git Proxy into production

PR #1385, opened 2026-02-06
docs: add deployment guide

tabathad:
Part of #1325, complements #1335.

Adds a deployment guide covering:

- Prerequisites
- Quick start walkthrough
- Docker deployment 
- Git client configuration
- Production considerations 

Also adds the deployment page to the docs sidebar.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -52,6 +52,27 @@ jobs:
       - name: Run E2E tests
         run: npm run test:e2e
 
+      - name: Run Cypress E2E tests
+        run: npm run cypress:run:docker
+        timeout-minutes: 10
+        env:
+          CYPRESS_BASE_URL: http://localhost:8081
+          CYPRESS_API_BASE_URL: http://localhost:8081
+          CYPRESS_GIT_PROXY_URL: http://localhost:8000
+          CYPRESS_GIT_SERVER_TARGET: git-server:8443
+
+      - name: Dump git-proxy logs on failure
+        if: failure()
+        run: docker compose logs git-proxy --tail=100
+
+      - name: Upload Cypress screenshots on failure
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
+        if: failure()
+        with:
+          name: cypress-screenshots
+          path: cypress/screenshots
+          retention-days: 7
+
       - name: Stop services
         if: always()
         run: docker compose down -v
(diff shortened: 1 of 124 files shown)


website/docs/deployment.mdx
```diff
@@ -0,0 +1,315 @@
...
+
+GitProxy supports two database backends (configured in `proxy.config.json` under `"sink"`):
+
+1. **NeDB (File-based)** — Default, suitable for development and single-server use
+   - No external dependencies
+   - Data stored in `./.data/db/` (hardcoded path, not configurable)
+   - Sessions are in-memory only (lost on restart)
+
+2. **MongoDB** — Recommended for production
+   - Requires MongoDB 7.0+
```
jescalada: I believe GitProxy is compatible with Mongo 6.0+, as shown in the latest [CI build matrix](https://github.com/finos/git-proxy/actions/runs/21772815409/job/62823566433).

```suggestion
   - Requires MongoDB 6.0+
```

website/docs/deployment.mdx
```diff
@@ -0,0 +1,315 @@
...
+```
+
+:::warning
+Repository URLs **must** include the `.git` suffix. Without it, you will get `fatal: info/refs not valid` errors.
+:::
+
+### 3. Start GitProxy
+
+```bash
+git-proxy --config ./proxy.config.json
```
jescalada, line 79: Oddly, this command doesn't seem to be correctly loading the config on my end. When calling `git-proxy --validate`, I see that the config file being loaded is `~/.nvm/versions/node/v22.13.1/lib/node_modules/@finos/git-proxy/proxy.config.json` which is not the "current working directory" that's mentioned in the option description...

I wonder if this is exclusive to the GitProxy global install and if there's an easy way to fix it.

website/docs/deployment.mdx
```diff
@@ -0,0 +1,315 @@
...
+```
+remote: GitProxy has received your push
+remote: Shareable Link
+remote: http://localhost:8080/dashboard/push/000000__b12557
+```
+
+### 6. Approve the Push
+
+1. Navigate to http://localhost:8080 in your browser
+2. Log in with default credentials (**development only**):
```
jescalada, line 125: We might want to include a link to the auth documentation in the architecture guide: https://github.com/finos/git-proxy/blob/1e98d9dbef27f6e338cf18354b0bbbc418b8a070/docs/Architecture.md#Authentication

The proper link will be available after that PR is merged!

website/docs/deployment.mdx
tabathad: I get the same error if I use the 1.19.2 package / latest. Using rc4 yeilds a different error, installing from the main branch works as expected

kriswest (approved):
LGTM - although the node version info is about to become out-of-date - @jescalada @1462 will need rebasing after this merges and will need to update the node version range in deployment.mdx.

This should merge now so that it goes out with 2.0.0

There is a check failing, but I think its not the PRs fault, but a timeout and just needs to be re-run.

PR #1389, opened 2026-02-06
fix: reduce DB and log bloat by not logging writePack content

### LINKED ISSUE #1388: writePack unnecessarily bloats the database and logs
**Describe the bug**
the writePack push processor unnecessarily logs two copies of the updated git index files in a very inefficient format used to produce a diff, unnecessarily and massively bloating both the database and logs. The content logged is not used for anything and can be removed.

I measured a DB record for a push with a two line commit at: 81,752 chars. By removing the logging and content from the writePack step this was reduced to 6,468 chars. I.e. the unnecessary logging increased the DB and log storage requirements by 13x - a factor I believe will be higher in more complex commits.

**To Reproduce**

Look at a database record for a push...

kriswest:
resolves #1388 

Reduces bloat of DB records and server logs (by 13x!) by not logging a byte array encoding the git index files created in the writePack step.

### DIFF
--- a/docs/Upgrading to v2.md
+++ b/docs/Upgrading to v2.md
@@ -1,0 +1,149 @@
+# Upgrading to GitProxy v2
+
+This guide attempts to cover everything needed for a seamless upgrade from GitProxy v1 (`1.19.2`) to v2.
+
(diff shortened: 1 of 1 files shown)


jescalada:
@kriswest Thanks for this!

PR #1393, opened 2026-02-09
fix: improve handling of repository clones in the .remote folder

### LINKED ISSUE #1386: Repository clones are not cleaned up individually or reliably
**Describe the bug**
As discussed in another issue (see https://github.com/finos/git-proxy/issues/1384#issuecomment-3855382042 and https://github.com/finos/git-proxy/issues/1384#issuecomment-3859225836), repository clones are not cleaned up individually ([the entire .remote folder is deleted in clearBareClone](https://github.com/finos/git-proxy/blob/ccb5d1df43c5a418dd161e80a09a0cb331857fb8/src/proxy/processors/push-action/clearBareClone.ts#L8)) or reliably (the clone is not cleared up at all if a processor before clearBareClone produces an error, the proxy being stopped, restarted or crashing can produce the same result). This can result in a push being permanently blocked by the error, e.g.:

```
$ git push upstream-dev-finos
Enumerating objects: 15, done.
Counting objects: 100% (15/15), done.
Delta compression using up to 4 threads
Compressing objects: 100% (8/8), done.
Writing objects: 100% (9/9), 879 bytes | 175.00 KiB/s, done.
Total 9 (delta 7), reused 3 (delta 1), pack-reused 0 (from 0)
remote:         An error occurred when executing the chain: CheckoutConflictError: Your local changes to the following files would be overwritten by checkout: sr
/components/Navbars/DashboardNavbarLinks.tsx, src/ui/views/User/UserProfile.tsx
fatal: the remote end hung up unexpectedly
fatal: the remote end hung up unexpectedly
error: failed to push some refs to 'https://*****/github.com/finos/git-proxy.git'
```

Requiring manual intervention (delete the .remote folder) to unpick.

Further: 
- multiple concurrent requests **relating to the same push** can conflict with each other by attempting to use the same folder (whose name is set via action.id, which is set based on the commit SHA).
- multiple concurrent requests **relating to the different pushes** can conflict with each other as, due to async functions used to handle them, they can be processed concurrently and `clearBareClone` deletes the whole .remote folder rather than just the one relating to that particular push.

**To Reproduce**

1. Failing to clean-up on errors:
    @fabiovincenzi reports:
    
    > if any step between pullRemote and clearBareClone fails, cleanup never happens. I tested this with gitleaks blocking a push .remote remains on disk with a full clone.
    
2. Failing to clean-up on shutdown/restart:
   - Push a change to a large repository
   - Stop the proxy while it is processing, after it has cloned the repository (easier to do if its a large change...)_
   - Confirm that the .remote/* folder for the push exists 
   - Check that local changes have been applied in order to calculate the diff - or apply a local change of your own
   - Restart the proxy
   - Push the change again - it should fail as shown above (refusing to overwrite local changes).

**Expected behaviour**

- The .remote folder is cleared up on startup (in case it was left in a poor state on restart
- .remote/* checkout folders are cleared up reliably on errors and after commits are successfully processed
- .remote/* folders are cleaned up individually, so that multiple concurrent requests for different pushes do not conflict with each other
- Multiple concurrent requests relating to the same repository/SHAs should should be handled such that the first request succeeds and subsequent requests should fail until the first has finished processing.

kriswest:
resolves #1386

Resolves a number of issues with .remote folder handling (git clones used for diffs), by:
- Deleting the .remote folder on start-up - clones are not currently supposed to persist so this should have been cleaned up
- Creating the .remote folder on start-up
- Not deleting the entire .remote folder when a single push is processed - there could be concurrent pushes to other repos/SHAs that are in progress (due to the async implementation of filtering) - now the individual clone is cleaned up
- ensuring successful clones are always cleaned up at the end of chain processing (where they were not previously if an error occurred)
- ensuring failed clones are cleaned up immediately in pull Remote
- rationalizing the naming of clearBareClone and audit as post-processors, rather than push-action processors (clearBareClone is no longer a direct part of the push chain, but invoked after processing if pullRemote ran successfully - which might one day be part of other chains)
- Improving the implementation  of the relevant tests.


src/proxy/chain.ts
```diff
@@ -32,6 +30,7 @@ const defaultActionChain: ((req: any, action: Action) => Promise<Action>)[] = [
 ];
 
 let pluginsInserted = false;
+let checkoutCleanUpRequired = false;
```
fabiovincenzi: We should move this variable inside executeChain, otherwise once it's set to true it will remain true forever — causing clearBareClone to run on every subsequent request.

src/proxy/actions/Action.ts
```diff
@@ -94,7 +94,8 @@ class Action {
   }
 
   /**
-   * Set the commit range for the action.
+   * Set the commit range for the action. Changes the action.id to be based on the
+   * the commit details.
```
fabiovincenzi, line 98:
```suggestion
   *commit details.
```

src/proxy/chain.ts
kriswest: will do

kriswest:
CI failed with: @jescalada @coopernetes CI failed with: 

```
  Unable to find image 'mongo:7.0' locally
  7.0: Pulling from library/mongo
  docker: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit.
```

Is the matrix unnecessarily large and should we be authenticating?

kriswest:
I also note the e2e tests depend on test-repo in @coopernetes namespace:

https://github.com/finos/git-proxy/blob/2452f10025afba85d47633a78afae55df1bf1a42/tests/e2e/push.test.ts#L271

Should we fix that? Its a real repo: https://github.com/coopernetes/test-repo
If we need a real repo thats not _finos/git-proxy_ we could ask help@finos.org to host one under [@finos-labs](https://github.com/finos-labs/) for us (with https://github.com/orgs/finos/teams/git-proxy-maintainers as the maintainers for it).

src/proxy/chain.ts
kriswest: that fixed the e2e failure - hence nice spot + nice validation of the value of the e2e tests.

jescalada:
@kriswest Suggestions on the doc are much appreciated! 😃 I'll go ahead and merge that PR afterwards.

PR #1404, opened 2026-02-13
chore: add apache license header script & CI check

### LINKED ISSUE #1002: Add Apache-2.0 license headers to source code
As part of FINOS activation (https://github.com/finos/community/issues/361), all relevant source code files must reflect the project's license as a license header in each relevant file.

### Tasks

- [ ] Identify all files that are missing the Apache-2.0 file header
- [ ] Append the Apache-2.0 license header to files

### Reference

[licenseheaders - a Python 3 tool to update, change or add license headers to all files](https://pypi.org/project/licenseheaders/)

jescalada:
**New PR at #1447**

Fixes #1002. Inspired by #1253, credit goes to @sam-holmes2 for the original PR 🙂 

### Changes
- Added `add-license-headers.ts` script to `/scripts` in order to add license headers automatically
  - Executed by running `npm run add-license-headers`
  - Checks if the license header is already present
  - Included special handling to prevent executable files (with shebangs in first line) from breaking
  - Excludes `/src/config/generated` files from having the license appended, due to infinite loop with `quicktype` generated config
- Prepended the licenses to all TS, JS, TSX and JSX files
  - We may want to extend this to cover other file extensions
- Added `check-license-headers.ts` for checking headers in CI
  - Added header check to Code Cleanliness workflow (`lint.yml`)

### DIFF
--- a/scripts/add-license-headers.ts
+++ b/scripts/add-license-headers.ts
@@ -38,7 +38,7 @@ function extractShebang(fileContent: string): string {
 }
 
 function hasLicenseHeader(fileContent: string, licenseHeader: string): boolean {
-  const normalizedContent = fileContent.trim().substring(0, 500);
+  const normalizedContent = fileContent.trim().substring(0, 200);
 
   return (
     normalizedContent.includes('@license') ||
--- a/scripts/check-license-headers.ts
+++ b/scripts/check-license-headers.ts
@@ -1,3 +1,23 @@
+/**
+ * @license
+ * Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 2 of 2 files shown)


scripts/add-license-headers.ts
jescalada, line 1: This is where the interesting stuff happens. So you don't have to dig through the diff! 🙂

jescalada:
@06kellyjac Here's a new PR implementing this using the ESLint plugin, I wasn't aware it existed! It's quite easy to use, too. #1447 

@kriswest Will close this PR if the new one is more appropriate 👍🏼

Issue #1402, opened 2026-02-12
UI silently ignores 403 errors on push approve/reject actions


fabiovincenzi:
**Describe the bug**
When a user without `canAuthorise` permission attempts to approve or reject a push via the admin UI dashboard, the server correctly returns `403 Forbidden`. However, the UI silently ignores the error and navigates the user back to the push list as if the action succeeded. No error message or feedback is shown.     

**To Reproduce**
  1. Login as a user with `canPush` permission but **without** `canAuthorise` permission
  2. Go to `/dashboard/push` and click on a pending push
  3. Click "Approve", check the attestation checkboxes, and click "Approve" to confirm (or click "Reject")
  4. The user is navigated back to the push list without any error message
  5. The push remains in "Pending" state — the action silently failed

**Expected behavior**
 A snackbar error message should inform the user that they lack permission to approve/reject the push (e.g. "You are not authorised to approve this push"). The user should remain on the push detail page.

 **Root cause**

  `authorisePush` and `rejectPush` in `src/ui/services/git-push.ts` only handle `401` (not authenticated) responses. `403` (not authorized) is silently swallowed in the `.catch()` block:

  ```ts
  // src/ui/services/git-push.ts - authorisePush (line 88-92)
  .catch((error) => {
    if (error.response && error.response.status === 401) { // only 401, not 403
      errorMsg = 'You are not authorised to approve...';
      isUserAllowedToApprove = false;
    }
  });
  ```

  The same pattern exists in rejectPush (lines 107-111).

  The server-side correctly returns 403 in these cases (src/service/routes/push.ts):
  - "Cannot approve/reject your own changes" (self-approval)
  - "User not authorised to approve/reject" (missing canAuthorise permission)

**Additional context**
Discovered during implementation of e2e tests for #1390. The negative test cases document this behavior with TODO comments in cypress/e2e/pushActions.cy.js.

PR #1406, opened 2026-02-13
feat: enhance error handling in git-push and repo services

### LINKED ISSUE #1188: Improve UI error handling and API error codes

### LINKED ISSUE #1392: Poor error messaging in the admin UI when the user is not logged in
**Describe the bug**
The admin UI has poor error messaging when the user is not logged in, in some cases. For example, a reviewer attempting to approve or reject a push is told 'You are not authorized to approve/reject" rather than being told they are not logged in or redirected to the login form.

Similarly an a committer attempting to cancel a commit is told 'You are not authorized to cancel" rather than being told they are not logged in or redirected to the login form.

**To Reproduce**

1. Attempt to reject, approve or cancel a push when not logged in.

**Expected behavior**
An accurate error message is reports OR (better) the user is redirected to the login form.

dcoric:
## Summary

This PR fixes both:
- Fixes #1392
- Fixes #1188

## Improve error handling and test coverage for UI services

This PR refactors error handling across the UI service layer and adds comprehensive test coverage. Error handling is now consistent, type-safe, and properly tested.

### What's changed

**Error handling improvements:**
- Added `src/ui/services/errors.ts` with reusable error utilities (`getServiceError`, `errorResult`, `successResult`)
- Refactored `git-push.ts` and `repo.ts` services to use the new error handling utilities
- Updated UI views (`PushDetails.tsx`, `RepoDetails.tsx`, etc.) to properly handle and display error states
- All service functions now return consistent `ServiceResult<T>` objects with success/failure states

**Test coverage (14 → 79 tests):**
- `test/ui/errors.test.ts` - comprehensive tests for error utility functions (18 tests)
- `test/ui/user.test.ts` - tests for user service functions (13 tests)
- `test/ui/git-push.test.ts` - expanded coverage for all git-push functions (13 tests)
- `test/ui/repo.test.ts` - expanded coverage for all repo functions (21 tests)

Tests cover success paths, HTTP errors (401/403/404/409/500), network failures, and edge cases.

### Benefits

- **Consistent error handling** - all services use the same pattern
- **Better UX** - proper error messages shown to users
- **Type safety** - TypeScript ensures correct error handling
- **Test coverage** - prevents regressions and documents expected behavior

Screenshot 2026-02-13 at 18 20 59

## Addition to this PR - Error Boundary

New file `src/ui/components/ErrorBoundary/ErrorBoundary.tsx`

ErrorBoundary class component with getDerivedStateFromError (render capture) + componentDidCatch (logging)

Wraps all route content in <ErrorBoundary name='Dashboard'> and catches any render error from any view without taking down the sidebar/navbar 

It has 2 modes - `dev` and `production`. In `dev` mode it has option to display stacktrace of the error, in production mode only offers option to reload:

DEV: 
Screenshot 2026-02-20 at 17 10 46

PROD:
Screenshot 2026-02-20 at 17 10 58

### DIFF
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -214,6 +214,17 @@ function Home() {
             </div>
           </div>
         </div>
+        <div className='row margin-bottom--xl'>
+          <div className='col col--4'>
+            <div className='col-demo'>
+              <Avatar
+                name='Andy Pols'
+                description='Engineer at Qube Research & Technologies'
+                username='andypols'
+              />
+            </div>
+          </div>
+        </div>
       </div>
       <div className='container text--center'>
         <h1 className='margin-top--md'>


src/ui/services/git-push.ts
```diff
@@ -1,133 +1,89 @@
...
+import { getApiV1BaseUrl } from './apiConfig';
 import { Action, Step } from '../../proxy/actions';
 import { PushActionView } from '../types';
+import { ServiceResult, errorResult, successResult } from './errors';
 
-const getPush = async (
-  id: string,
-  setIsLoading: (isLoading: boolean) => void,
-  setPush: (push: PushActionView) => void,
-  setAuth: (auth: boolean) => void,
```
jescalada, line 11: Was `setAuth` actually doing anything here before? I think it might've been used for handling expired sessions, but it's hard to tell since not all of the old fetch handlers make use of the hook... 🤔

src/ui/services/git-push.ts
```diff
@@ -1,133 +1,89 @@
...
   const apiV1Base = await getApiV1BaseUrl();
   const url = new URL(`${apiV1Base}/push`);
   url.search = new URLSearchParams(query as any).toString();
 
-  setIsLoading(true);
-
   try {
     const response = await axios<Action[]>(url.toString(), getAxiosConfig());
-    setPushes(response.data as PushActionView[]);
+    return successResult(response.data as PushActionView[]);
```
jescalada: Wonder if there's a way to prevent casting here... It seems that the enhanced `Action` object (`PushActionView`) isn't accurately typed to begin with, since `PushActionView` should be more of a "data container" rather than an actual `Action` with various methods (`addStep`, etc.). That's why we get this error when trying to remove the casting:

```
Type '{ diff: Step; id: string; type: string; method: string; timestamp: number; project: string; repoName: string; url: string; repo: string; steps: Step[]; error: boolean; errorMessage?: string | null | undefined; ... 19 more ...; newIdxFiles?: string[] | undefined; }' is missing the following properties from type 'PushActionView': addStep, getLastStep, setCommit, setBranch, and 5 more.ts(2740)
```

jescalada: I'd suggest to change the `PushActionView` type to contain the `Action` type but without the unneeded methods - then we could just unpack the `response.data` and initialize a proper `PushActionView` object:

```ts
const getPush = async (id: string): Promise<ServiceResult<PushActionView>> => {

  try {
    const data: Action = response.data;
    const actionView: PushActionView = {
      ...data,
      diff: data.steps.find((x: Step) => x.stepName === 'diff')!,
    };
    return successResult(actionView);
  } catch (error: any) {
    return errorResult(error, 'Failed to load push');
  }
};
```

src/ui/services/user.ts
```diff
@@ -27,14 +28,12 @@ const getUser = async (
     setUser?.(user);
     setIsLoading?.(false);
   } catch (error) {
-    const axiosError = error as AxiosError;
-    const status = axiosError.response?.status;
+    const { status, message } = getServiceError(error, 'Unknown error');
     if (status === 401) {
       setAuth?.(false);
```
jescalada, line 33: Wondering about why we keep `setAuth` here in `user.ts` but not in the other service files 🤔 Might want to double-check that we weren't using `setAuth` for expired session handling or something else entirely.

src/ui/services/git-push.ts
kriswest: I think it may have been vestigial... Where it was used in PushDetails.tsx the state variable wasn't actually used:

`const [, setAuth] = useState(true);`

src/ui/views/PushDetails/PushDetails.tsx
```diff
@@ -19,65 +19,85 @@ import TableHead from '@material-ui/core/TableHead';
...
-    }
+    if (!id) return;
+    const load = async () => {
+      setIsLoading(true);
+      const result = await getPush(id);
+      if (result.success && result.data) {
+        setPush(result.data);
+      } else if (result.status === 401) {
+        setIsLoading(false);
+        navigate('/login', { replace: true });
```
kriswest, line 57: 👍

src/ui/views/PushDetails/PushDetails.tsx
```diff
@@ -19,65 +19,85 @@ import TableHead from '@material-ui/core/TableHead';
...
-
-  const setUserAllowedToApprove = (userAllowedToApprove: boolean) => {
-    isUserAllowedToApprove = userAllowedToApprove;
-  };
-
-  const setUserAllowedToReject = (userAllowedToReject: boolean) => {
-    isUserAllowedToReject = userAllowedToReject;
+  const handleActionFailure = (result: ServiceResult) => {
+    if (result.status === 401) {
+      navigate('/login', { replace: true });
```
kriswest, line 42: 👍

src/ui/services/errors.ts
```diff
@@ -0,0 +1,35 @@
...
+  fallbackMessage: string,
+): { status?: number; message: string } => {
+  const status = error?.response?.status;
+  const responseMessage = error?.response?.data?.message;
+  const message =
+    typeof responseMessage === 'string' && responseMessage.trim().length > 0
+      ? responseMessage
+      : error?.message || fallbackMessage;
+  return { status, message };
+};
```
kriswest, line 21: Happy to see the API determine the message returned to the user. The error messages from the API are generally informative (with a few small exceptions).

I wonder if all you need is a default 'unknown error occurred, response code: XXX' message if the API doesn't return a message. That should mostly happen on 500/server errors I imagine.

src/ui/services/git-push.ts
dcoric: Fixed — it was vestigial. As @kriswest noted, the state variable was declared as const [, setAuth] = useState(true) in PushDetails.tsx and the setter was never actually consumed.

src/ui/services/git-push.ts
dcoric: Done - changed `PushActionView` to extends `Omit<Action, ActionMethods>` in `types.ts`, stripping the class methods from the required interface. `getPush` now constructs `actionView` with the spread + explicit `diff` as you suggested, with no casting needed. For `getPushes` (list), the cast is now `as unknown as PushActionView[]` to be explicit that list items don't have `diff` computed - the table view never uses it.

src/ui/services/user.ts
dcoric: The `setAuth` in `user.ts` is intentional, unlike `git-push.ts`, it's actively used to handle 401 responses by setting auth state to `false`, which drives the expired session / redirect-to-login flow. The pattern is different from the vestigial one in `PushDetails`.

src/ui/services/errors.ts
dcoric: Done. When the API returns a status code but no message, `getServiceError` now returns `Unknown error occurred, response code: ${status}` instead of falling through to the axios error message.

src/ui/views/RepoDetails/RepoDetails.tsx
```diff
@@ -67,22 +82,37 @@ const RepoDetails: React.FC = () => {
...
+    } else if (result.status === 401) {
+      navigate('/login', { replace: true });
+    } else {
+      setIsError(true);
+      setErrorMessage(result.message || 'Failed to refresh repository data');
+    }
   };
 
   const removeRepository = async (id: string) => {
     await deleteRepo(id);
```
fabiovincenzi: Add a try-catch here as well?

src/ui/views/RepoDetails/RepoDetails.tsx
```diff
@@ -67,22 +82,37 @@ const RepoDetails: React.FC = () => {
   const removeUser = async (userToRemove: string, action: 'authorise' | 'push') => {
     if (!repoId) return;
     await deleteUser(userToRemove, repoId, action);
```
fabiovincenzi: We could add a try-catch here to notify the user, since deleteUser can throw, something like:
```
try {
        await deleteUser(userToRemove, repoId, action);
 } catch (err: any) {
        setIsError(true);
        setErrorMessage(err.message || 'Failed to remove user');
        return;
 }
```

fabiovincenzi (commented):
Looks good to me, and it covers #1402 , thanks @dcoric!

jescalada (approved):

PR #1409, opened 2026-02-19
fix: update @types/express-serve-static-core to 5.1.1 and fix typing of Request params

kriswest:
While testing I noticed that theres an update to express-serve-static-core 5.06 -> 5.1.1 that contains a correction to express's Request type - which looks breaking (but the change actually happened in a major release, the type was just wrong). 

At present `Request.params` has type `Record<string, string>` but its being 'corrected' to `Record<string, string | string []>` (as path params in express 5 can have multiple components and provide an array of values in those cases).

To prevent type issues that break the build, we need to strongly type Request, which I've done in this PR. @jescalada this may have a bearing on #1323 where you are adding in the Request type in may places (and may also need to add the types for the params object there if accessing its properties).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -318,44 +318,44 @@
       }
     },
     "node_modules/@aws-sdk/client-sso": {
-      "version": "3.980.0",
-      "resolved": "https://registry.npmjs.org/@aws-sdk/client-sso/-/client-sso-3.980.0.tgz",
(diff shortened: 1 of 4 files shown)


jescalada (approved):
LGTM! 👍🏼

I think it's best to merge this in, I can deal with the conflicts in #1323 which still has a few pending comments to fix.

PR #1414, opened 2026-02-23
feat: add TLS support to API/UI service

### LINKED ISSUE #1408: Enable TLS for UI and API
Currently, TLS is configurable only for the proxy. We could extend the implementation to add secure connections to the API/UI server.

The current implementation allows configuring the TLS parameters as follows (`config.schema.json`):

```json
"tls": {
  "description": "TLS configuration for secure connections",
  "type": "object",
  "properties": {
    "enabled": { "type": "boolean" },
    "key": { "type": "string" },
    "cert": { "type": "string" }
  },
  "required": ["enabled", "key", "cert"]
},
```

These values are fetched and used when creating the server:

```ts
if (getTLSEnabled()) {
  await new Promise<void>((resolve, reject) => {
    const server = https.createServer(getServerOptions(), this.expressApp!);
    server.on('error', reject);
    server.listen(proxyHttpsPort, () => {
      console.log(`HTTPS Proxy Listening on ${proxyHttpsPort}`);
      resolve();
    });
    this.httpsServer = server;
  });
}
```

It would be nice to extend this implementation so that the API/UI server can make use of TLS as well.

**Additional context**
https://github.com/finos/git-proxy/pull/1391#discussion_r2783316165

fabiovincenzi:
Fixes #1408 
Add HTTPS on port 8444 (configurable via `GIT_PROXY_HTTPS_UI_PORT`) to the API/UI service, using the existing `tls.enabled/key/cert` config fields already used by the proxy. HTTP and HTTPS run in parallel.


src/service/index.ts
```diff
@@ -15,12 +17,24 @@ import { configure } from './passport';
...
+
+const getServiceTLSOptions = () => ({
+  key:
+    config.getTLSEnabled() && config.getTLSKeyPemPath()
+      ? fs.readFileSync(config.getTLSKeyPemPath()!)
+      : undefined,
+  cert:
+    config.getTLSEnabled() && config.getTLSCertPemPath()
+      ? fs.readFileSync(config.getTLSCertPemPath()!)
+      : undefined,
```
jescalada, line 52: Is it okay to reuse the proxy TLS credentials for the service and UI or should these be separate configurable parameters 🤔

jescalada (commented):
Code-wise LGTM, might want another look from @kriswest to make sure.

jescalada (approved):
LGTM! Should be good to ship in 2.1 👍🏼

Issue #1397, opened 2026-02-09
23rd Feb 2026 - GitProxy Meeting Minutes


kriswest:
## Date

20260223 - 4pm GMT / 11am EST

## Meeting info

- [Meeting link](https://zoom-lfx.platform.linuxfoundation.org/meeting/95849833904?password=99413314-d03a-4b1c-b682-1ede2c399595)

- [Register for future meetings](https://zoom-lfx.platform.linuxfoundation.org/meeting/95849833904?password=99413314-d03a-4b1c-b682-1ede2c399595&invite=true)

## Meeting notices

- FINOS **Project leads** are responsible for observing the FINOS guidelines for [running project meetings](https://community.finos.org/docs/governance/meeting-procedures/). Project maintainers can find additional resources in the [FINOS Maintainers Cheatsheet](https://community.finos.org/docs/finos-maintainers-cheatsheet).

- **All participants** in FINOS project meetings are subject to the [LF Antitrust Policy](https://www.linuxfoundation.org/antitrust-policy/), the [FINOS Community Code of Conduct](https://community.finos.org/docs/governance/code-of-conduct) and all other [FINOS policies](https://community.finos.org/docs/governance/#policies).

- FINOS meetings involve participation by industry competitors, and it is the intention of FINOS and the Linux Foundation to conduct all of its activities in accordance with applicable antitrust and competition laws. It is therefore extremely important that attendees adhere to meeting agendas, and be aware of, and not participate in, any activities that are prohibited under applicable US state, federal or foreign antitrust and competition laws. Please contact legal@finos.org with any questions.

- FINOS project meetings may be recorded for use solely by the FINOS team for administration purposes. In very limited instances, and with explicit approval, recordings may be made more widely available.

## Agenda

- [x] Convene & roll call (5mins)
- [x] Display [FINOS Antitrust Policy summary slide](https://community.finos.org/Compliance-Slides/Antitrust-Compliance-Slide.pdf) and Review Meeting Notices (see above)
- [x] Approve past meeting minutes and review actions
  - #1367
- [x] 2.0 release status
- [x] Should attestation bypass bug be reported as vulnerability?
- [x] #1400 and #1332 
- [x] AOB, Q&A & Adjourn (5mins)

## Minutes

- **Approval of Past Meeting Minutes and Review Actions**
    - Previous minutes approved; actions reviewed with updates and clarifications. Several items closed or rolled over as appropriate.

- **2.0 Release Status**
[...]

PR #1417, opened 2026-02-23
docs: move usage to quickstart

tabathad:
Moves the [usage page](https://git-proxy.finos.org/docs/usage) under the Quick Start section. The current usage page is really about kicking-off rather than a proper deployment guide, so it fits better in the quickstart section.

Discussed in meeting: #1397

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
 
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@e3f713f2d8f53843e71c69a996d56f51aa9adfb9 # v2.14.1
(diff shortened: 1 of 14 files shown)


website/docs/quickstart/usage.mdx
```diff
@@ -1,11 +1,11 @@
 ---
 title: Usage
-description: How to run GitProxy in your environment
+description: How to run GitProxy using npm or npx
 ---
 
 ### Run from global install
 
-Once you have followed the [installation](installation) steps, run:
+Once you have followed the [installation](../installation) steps, run:
```
kriswest: I think we might as well also move installation into quickstart (as the first link), its a prerequisite for usage.

```suggestion
Once you have followed the [installation](installation) steps, run:
```

PR #1421, opened 2026-02-25
refactor: replace async/await and promise chaining

### LINKED ISSUE #1412: Remove async/await and promise chaining mixing
Right now, there are several places in the code where we mix `async`/`await` and promise chaining (`.then`). Since it's harder to reason about the code, we should make these consistent.

Here's an example from `/src/ui/services/repo.ts`:

```ts
const getPush = async (
  id: string,
  setIsLoading: (isLoading: boolean) => void,
  setPush: (push: PushActionView) => void,
  setAuth: (auth: boolean) => void,
  setIsError: (isError: boolean) => void,
): Promise<void> => {
  setIsLoading(true);

  await axios<Action>(url, getAxiosConfig())
    .then((response) => {
      const data: Action & { diff?: Step } = response.data;
      data.diff = data.steps.find((x: Step) => x.stepName === 'diff');
      setPush(data as PushActionView);
    })
    .catch((error: AxiosError<string>) => {
      if (error.response?.status === 401) setAuth(false);
      else setIsError(true);
    })
    .finally(() => {
      setIsLoading(false);
    });
};
```

Which can be refactored as follows:

```ts
const getPush = async (
  id: string,
  setIsLoading: (isLoading: boolean) => void,
  setPush: (push: PushActionView) => void,
  setAuth: (auth: boolean) => void,
  setIsError: (isError: boolean) => void,
): Promise<void> => {
  setIsLoading(true);

  try {
    const data: Action & { diff?: Step } = response.data;
    data.diff = data.steps.find((x: Step) => x.stepName === 'diff');
    setPush(data as PushActionView);
  } catch (error: unknown) {
    if (error instanceof AxiosError && error.response?.status === 401) setAuth(false);
    else setIsError(true);
  } finally {
    setIsLoading(false);
  }
};
```

Note that errors must be of type `unknown` and properly parsed without casting or using `any` types.

### Additional context
Related PR: #1174

fabiovincenzi:
Fixes #1412

### DIFF
--- a/src/ui/services/repo.ts
+++ b/src/ui/services/repo.ts
@@ -8,20 +8,18 @@ import { ServiceResult, getServiceError, errorResult, successResult } from './er
 const canAddUser = async (repoId: string, user: string, action: string) => {
   const apiV1Base = await getApiV1BaseUrl();
   const url = new URL(`${apiV1Base}/repo/${repoId}`);
-  return axios
-    .get<Repo>(url.toString(), getAxiosConfig())
-    .then((response) => {
-      const repo = response.data;
-      if (action === 'authorise') {
-        return !repo.users.canAuthorise.includes(user);
-      } else {
-        return !repo.users.canPush.includes(user);
-      }
-    })
-    .catch((error: any) => {
-      const { message } = getServiceError(error, 'Failed to validate repo permissions');
-      throw new Error(message);
-    });
+  try {
+    const response = await axios.get<Repo>(url.toString(), getAxiosConfig());
+    const repo = response.data;
+    if (action === 'authorise') {
+      return !repo.users.canAuthorise.includes(user);
+    } else {
+      return !repo.users.canPush.includes(user);
+    }
+  } catch (error: any) {
+    const { message } = getServiceError(error, 'Failed to validate repo permissions');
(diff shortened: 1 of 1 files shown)


src/ui/views/Login/Login.tsx
```diff
@@ -36,20 +36,21 @@ const Login: React.FC = () => {
   const [usernamePasswordMethod, setUsernamePasswordMethod] = useState<string>('');
 
   useEffect(() => {
-    getBaseUrl().then((baseUrl) => {
-      axios.get(`${baseUrl}/api/auth/config`).then((response) => {
-        const usernamePasswordMethod = response.data.usernamePasswordMethod;
-        const otherMethods = response.data.otherMethods;
+    const fetchAuthConfig = async () => {
+      const baseUrl = await getBaseUrl();
+      const response = await axios.get(`${baseUrl}/api/auth/config`);
```
dcoric, line 41: This is a very small nitpick - if API call fails, error will not be handled. As it is was not handled in a previous version, it is not regression.

jescalada: I'll fix this up in the any/as refactor PR which also refactors how we handle errors #1323

PR #1431, opened 2026-03-02
chore: fix push sort order to be latest-first

andypols:
When viewing push requests, the current implementation assumes MongoDB (or the file-based DB) returns results in a consistent order, and the UI simply reverses whatever it receives. Because the database query does not guarantee ordering, pushes could be displayed in an inconsistent (often effectively random) order.

This PR fixes that by:

- Adding an explicit sort in the database layer so the most recent pushes are returned first.

- Simplifying the push table rendering to rely on backend ordering rather than reordering in the UI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,7 +84,7 @@ jobs:
           path: build
 
       - name: Download the build folders
-        uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0 # v5
+        uses: actions/download-artifact@70fc10c6e5e1ce46ad2ea6f2b72d43f7d47b13c3 # v8
         with:
           name: build-ubuntu-node-${{ matrix.node-version }}-mongo-${{ matrix.mongodb-version }}
           path: build
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -34,14 +34,14 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@c0fc915677567258ee3c194d03ffe7ae3dc8d741 # ratchet:github/codeql-action/init@v4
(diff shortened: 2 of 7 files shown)


src/db/file/pushes.ts
```diff
@@ -34,19 +34,21 @@ const defaultPushQuery: Partial<PushQuery> = {
...
-        reject(err);
-      } else {
-        resolve(
-          _.chain(docs)
-            .map((x) => toClass(x, Action.prototype))
-            .value(),
-        );
-      }
-    });
+    db.find(query)
```
kriswest, line 37: @andypols I wonder if we should do this in the API service, or even UI rather than the DB client? That way we can have controls to flip sort order, use a different sort order etc.. We already accept a PushQuery object thats supposed to define such things, but its currently empty and just passed to `find`. We'd have to extend [PushQuery](https://github.com/finos/git-proxy/blob/9bb8059e17abc80bac6b4db0cdcda2e86108da09/src/db/types.ts#L4-L13) (which was defined to cover our current usage rather than than intentionally excluding sorting criteria I believe) and then inspect it to apply to the query.

Regardless, I think this is better than the current situation and would accept it. I just wanted to push the idea of going a little further in the architecture as we're going to need pagination and a few other things in due course.

andypols: @kriswest I agree this is a stepping stone and there’s more we can improve here. I don’t think we should do this in the API service layer, since that likely implies pulling all rows from the database and then sorting/filtering in memory.

Our implementation is close to the point where we should implement paging, column sorting, and filtering at the database/query level so we can reduce the volume of data returned and keep performance predictable.

andypols:
@dcoric  Good shout on the tests.  Ok if I add a separate PR for that?

src/db/file/pushes.ts
kriswest: Apologies for not being clear (I do 100% agree) - I meant to suggest that we add the desired sort column and direction to the PushQuery object in the API layer. In the long run we definitely need pagination etc. also, which has to come from there, so I was trying to suggest we start on that (which will involve separating the filter elements from pagination and sorting elements before passing to `find`, `sort`, `limit` and `skip`). I think right now all we've got in any of the services is something that ignores skip and limit if set. 

Were no sort supplied, this would be the default to apply - so agreed this is a valid stepping stone

PR #1446, opened 2026-03-05
feat: Add action summary to push view

### LINKED ISSUE #1387: Pushes in the 'error' category don't render in the admin UI unless a diff was produced
**Describe the bug**
Pushes that end up in the error category can't be viewed in the admin UI if the diff step failed:

```
"lastStep": {
    "logs": [
      "diff - Executing \"git diff c338e9a0035b755f4f5c6fff00fa8b8f206d8c05 cee0c02d1cb4c22a7f014b370f00f83868a7c746\" in ./.remote/0000000000000000000000000000000000000000__cee0c02d1cb4c22a7f014b370f00f83868a7c746/fdc3.git",
      "diff - Error: fatal: Invalid revision range c338e9a0035b755f4f5c6fff00fa8b8f206d8c05..cee0c02d1cb4c22a7f014b370f00f83868a7c746\n"
    ],
    "id": "c5d80af9-465d-4e75-8d2a-47300261c034",
    "stepName": "diff",
    "content": null,
    "error": true,
    "errorMessage": "Error: fatal: Invalid revision range c338e9a0035b755f4f5c6fff00fa8b8f206d8c05..cee0c02d1cb4c22a7f014b370f00f83868a7c746\n",
    "blocked": false,
    "blockedMessage": null
  }
```

Image

This results in a white screen with no UI rendered. 

**To Reproduce**

- Create a branch where the base is not in the upstream repo (e.g. create a branch, add some commits to it, then create a new branch from it, add commits to that and then try to push the result to the proxy)
- The proxy should reject the push, but add it to the error tab in the dashboard
- Attempt to view the diff.

**Expected behavior**
Details of the push should be displayed, including the error message, but with no Diff.

andypols:
# Summary

It can be hard to understand why a push was rejected. This PR adds a view of each step involved in a push, making it easier to identify failures. It also fixes #1387 and prevents the UI from crashing when a push fails before a diff is generated.

We use "steps" instead of actions in the UI, since “actions” was confusing for users.

## Solution

* Introduces a tabbed view for push data, including Commits, Diff, and Steps.  This reduces scrolling and provides a foundation for future additions, such as an AI-generated change summary (future PR).
CleanShot 2026-03-06 at 08 38 11@2x

* Adds dedicated components for each tab.
* Extended `CustomTabs` to support an optional badge.
* Calculates the total error count across steps and displays it as a badge to make failures more visible.
* Sums the error count from the steps and displays an error count as a badge to highlight there are errors.
* Adds a StepsTimeline component featuring an accordion layout, status chips, and log output to better surface errors, blocked steps, and successful execution.

CleanShot 2026-03-06 at 08 39 41@2x

* Clicking on an accordion step  the info about that action 

CleanShot 2026-03-06 at 08 42 19@2x

* Prevents large step payloads from being rendered.

### DIFF
--- a/.github/workflows/pr-lint.yml
+++ b/.github/workflows/pr-lint.yml
@@ -16,6 +16,7 @@ permissions:
 jobs:
   pr_title:
     permissions:
+      contents: write
       pull-requests: write
       statuses: write
     name: Validate & Label PR
@@ -45,5 +46,7 @@ jobs:
             test
             break
       - uses: release-drafter/release-drafter@139054aeaa9adc52ab36ddf67437541f039b88e2 # v7
+        with:
+          commitish: main
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


src/ui/views/PushDetails/PushDetails.tsx
```diff
@@ -249,44 +250,29 @@ const Dashboard: React.FC = () => {
...
+              },
+              {
+                tabName: 'Changes',
+                tabIcon: CodeIcon,
+                tabContent: <Diff diff={push.diff?.content || ''} />,
+              },
+              {
+                tabName: 'Steps',
+                tabIcon: TimelineIcon,
+                tabContent: <StepsTimeline steps={push.steps} />,
```
re-vlad: Can we pass `steps={push.steps ?? []}` into `StepsTimeline` so the component never receives undefined

src/ui/views/PushDetails/PushDetails.tsx
```diff
@@ -115,6 +114,8 @@ const Dashboard: React.FC = () => {
   if (isError) throw new Error(message || 'Something went wrong ...');
   if (!push) return <div>No push data found</div>;
 
+  const errorCount = push.steps.filter((step) => step.error).length;
```
re-vlad: Small non-blocking suggestion: if `push.steps` is ever undefined (e.g. old API shape or malformed response), this throws. As for me, optional is more appropriate, something like this: `const errorCount = push.steps?.filter((step) => step.error).length ?? 0`

src/ui/views/PushDetails/components/CommitDataTable.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <TableHead>
+        <TableRow>
+          <TableCell>Timestamp</TableCell>
+          <TableCell>Committer</TableCell>
+          <TableCell>Author</TableCell>
+          <TableCell>Message</TableCell>
+        </TableRow>
+      </TableHead>
+      <TableBody>
+        {commitData.map((c) => (
```
re-vlad, line 47: When `commitData` is empty, the table renders only headers with no body rows. Can you add an explicit empty state (e.g. a single row with colSpan and "No commits" or a short message)

src/ui/views/PushDetails/PushDetails.tsx
fabiovincenzi: Also noticed that StepsTimeline already has an `expandStepId` prop that auto-expands a specific accordion, but it's never passed from `PushDetails`. If we calculate the first failing step, we can auto-expand it when the user opens the Steps tab:                                                                                                                                                                                      
  `const firstErrorStep = (push.steps ?? []).find((step) => step.error);`                                                                             
                                                                                                                                                                                                                                                                           
  `tabContent: <StepsTimeline steps={push.steps ?? []} expandStepId={firstErrorStep?.id} />`                                                      
                                                                                                                                                    
  This way errors are immediately visible without the user having to manually click the failing step.

andypols: @fabiovincenzi  auto-opening the steps with errors is a nice idea, but feels strange passing  firstErrorStep in when StepsTimeline has all the information to do itself?

andypols: @fabiovincenzi I've added your suggestion to auto-open on the first error if there are errors.

src/ui/views/PushDetails/components/CommitDataTable.tsx
jescalada: `commitData` shouldn't be empty under normal circumstances, the `checkEmptyBranch` action handles this:

```ts
  const step = new Step('checkEmptyBranch');

  if (action.commitData && action.commitData.length > 0) {
    return action;
  }

  if (await isEmptyBranch(action)) {
    step.setError('Push blocked: Empty branch. Please make a commit before pushing a new branch.');
    action.addStep(step);
    step.error = true;
    return action;
  } else {
    step.setError(
      'Push blocked: Commit data not found. Please contact an administrator for support.',
    );
    action.addStep(step);
    step.error = true;
    return action;
  }
};
```

jescalada: That said, it would be nice to have a short message there so it's obvious to the end user that no commit data was found (they might assume an error happened while fetching the commits).

src/ui/views/PushDetails/components/StepsTimeline.tsx
```diff
@@ -0,0 +1,302 @@
...
+  steps: StepData[];
+}
+
+const StepsTimeline: React.FC<StepsTimelineProps> = ({ steps }) => {
+  const classes = useStyles();
+  const [expanded, setExpanded] = useState<string | false>(
+    () => steps.find((s) => s.error)?.id ?? false,
+  );
+
+  const isLargeStep = (stepName: string) => stepName === 'writePack' || stepName === 'diff';
```
jescalada, line 146: It might be worth revisiting this now that a [few logging enhancements (#1448)](https://github.com/finos/git-proxy/pull/1448) have gone in. Some actions may have gotten longer due to replacing `console.log` with `step.log`.

src/ui/views/PushDetails/PushDetails.tsx
jescalada: `Action.steps` shouldn't be undefined as it's initialized as an empty array in `src/proxy/actions/Action.ts`:

```ts
class Action {
  id: string;
  type: string;
  ...
  steps: Step[] = [];
```

src/ui/views/PushDetails/components/StepsTimeline.tsx
andypols: @jescalada I tried it in a few pushes and looks good.  Please can you review again.

PR #1338, opened 2026-01-08
refactor: Improve logging in processors/action chain

RichardUri:
Resolves log data not being recorded due to use of console.log. Now properly uses step.log.
Fixes #1281

PR #1448, opened 2026-03-08
refactor: improve logging in `src/proxy/processors`, remove log-related stubs/spies in tests

### LINKED ISSUE #1281: Improve logging in `processors`/action chain
Some action files are simply logging errors to the GitProxy console instead of adding errors as a `Step` (thus the error data would get lost instead of getting saved to the Push DB).

**Example**
`checkCommitMessages.ts` has a few of these:

```ts
// Commit message is empty, i.e. '', null or undefined
if (!commitMessage) {
  console.log('No commit message included...');
  return false;
}

// Validation for configured block pattern(s) check...
if (typeof commitMessage !== 'string') {
  console.log('A non-string value has been captured for the commit message...');
  return false;
}
```

> There are several other places that info on why the message has failed the check is just being logged. We should refactor this to return an error message or null and then carry that forward into step.log messages, as that'll make illegal messages way less annoying to debug and we can drop all the debug logging here.

Remove unnecessary logging across `src/proxy/processors` and replace with `Step.log` wherever it makes sense (especially places where errors should be thrown).

**Additional context**
https://github.com/finos/git-proxy/pull/1202#discussion_r2511134308
https://github.com/finos/git-proxy/pull/1202#discussion_r2511141424

jescalada:
Fixes #1281, supersedes #1338

Also improves wording and log handling in general, and fixes log stubs in tests.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -52,6 +52,27 @@ jobs:
       - name: Run E2E tests
         run: npm run test:e2e
 
+      - name: Run Cypress E2E tests
+        run: npm run cypress:run:docker
+        timeout-minutes: 10
+        env:
+          CYPRESS_BASE_URL: http://localhost:8081
+          CYPRESS_API_BASE_URL: http://localhost:8081
+          CYPRESS_GIT_PROXY_URL: http://localhost:8000
+          CYPRESS_GIT_SERVER_TARGET: git-server:8443
+
+      - name: Dump git-proxy logs on failure
+        if: failure()
+        run: docker compose logs git-proxy --tail=100
+
+      - name: Upload Cypress screenshots on failure
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
+        if: failure()
+        with:
+          name: cypress-screenshots
+          path: cypress/screenshots
+          retention-days: 7
+
       - name: Stop services
         if: always()
         run: docker compose down -v
--- a/cypress.config.js
+++ b/cypress.config.js
@@ -19,7 +19,13 @@ const { defineConfig } = require('cypress');
 module.exports = defineConfig({
   e2e: {
     baseUrl: process.env.CYPRESS_BASE_URL || 'http://localhost:3000',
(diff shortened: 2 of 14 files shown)


src/proxy/processors/push-action/parsePush.ts
```diff
@@ -63,7 +63,7 @@ async function exec(req: any, action: Action): Promise<Action> {
         'Your push has been blocked. Please make sure you are pushing to a single branch.',
       );
     } else {
-      console.log(`refUpdates: ${JSON.stringify(refUpdates, null, 2)}`);
+      console.log(`parsePush refUpdates: ${JSON.stringify(refUpdates, null, 2)}`);
```
re-vlad: please consider replacing with `step.log` so all parsePush diagnostics go into the Push DB

src/proxy/processors/push-action/scanDiff.ts
re-vlad: look like leftover debug logging (replacing with `step.log` only when needed?)

src/proxy/processors/push-action/parsePush.ts
jescalada: I feel that removing that line entirely might be the best choice as it reveals too much about the internal implementation...

@kriswest Just wondering if you recall the reasoning behind logging the `refUpdates`, and whether you think it's appropriate to add into the step logs.

src/proxy/processors/push-action/scanDiff.ts
jescalada: Whoops, I missed this one! It doesn't seem like these are being used meaningfully, just cluttering the output.

src/proxy/processors/push-action/parsePush.ts
andypols: @jescalada I would be tempted to remove it if we can't think of a good reason to keep it.

andypols (approved):
LGTM -  @jescalada I think we merge this in if you can resolve the conversation.

src/proxy/processors/push-action/parsePush.ts
jescalada: Done!

PR #1457, opened 2026-03-14
feat: add password reset for default admin/user accounts

### LINKED ISSUE #1022: Improve default admin user creation

jescalada:
Fixes #1022.

I opted for the password reset solution instead of relying on the config, since I didn't want to cause any extra confusion for new users (GitProxy administrators) or friction for testing/development which relies on these dummy accounts.

The bulk of the changes are test files and frontend additions/refactoring.

I also added tests to improve coverage for auth routes.

**Disclaimer:** This PR was created with agentic AI and refactored manually to improve code quality. I validated the password reset flow manually and with Cypress tests. Please keep this in mind during review - there may be some obvious mistakes I missed!

### Screenshot

Shown automatically when logging in with admin/admin for the first time, and when trying to navigate to other pages. Note that this is only visible when `NODE_ENV` is set to production.

image


src/service/routes/auth.ts
```diff
@@ -151,6 +178,84 @@ router.post('/logout', (req: Request, res: Response, next: NextFunction) => {
...
+      return;
+    }
+
+    const hashedPassword = await bcrypt.hash(newPassword, 10);
+    await db.updateUser({
+      username: user.username,
+      password: hashedPassword,
+      mustChangePassword: false,
+    });
+    (req.user as User).mustChangePassword = false;
```
dcoric, line 242: The session object is mutated in-place and not regenerated. In a session-fixation scenario (attacker knows a pre-login session ID), the attacker's session remains valid post-password-change. It would be good to call `req.session.regenerate()` after a successful password change.

src/ui/components/RouteGuard/RouteGuard.tsx
```diff
@@ -57,6 +57,10 @@ const RouteGuard = ({ component: Component, fullRoutePath }: RouteGuardProps) =>
     return <Navigate to='/login' />;
   }
 
+  if (user?.mustChangePassword) {
+    return <Navigate to='/login?forcePasswordChange=1' />;
```
dcoric: Login.tsx never reads this query param and the password change form is shown based on authContext.user?.mustChangePassword (line 74). The query param has no effect. It should be wired up or removed

dcoric (changes requested):
The overall approach is solid. I spotted a few small things worth addressing

src/ui/components/RouteGuard/RouteGuard.tsx
jescalada: You're right, this query param was hallucinated and never actually used... The current flow simply lets you login with the "admin/admin" default for the first time, and then shows you the "password change" form after logging in (and keeps showing it even if you navigate elsewhere). So the param isn't needed. 👍🏼 

We have to be careful when developing with agents since questionable code might get introduced despite the code working as intended...

dcoric (approved):
Looks good!

kriswest:
@andypols have a quick look at this please

jescalada:
@andypols Should be ready for a look now!

andypols (approved):
LGTM - there is one thing I noticed:

The password reset currently updates the stored password and clears `mustChangePassword`, but it does not regenerate/destroy the current session or invalidate other existing sessions for the same user.

If someone logs in with the default admin password before the legitimate admin resets it, their session may remain valid after the password is changed. Once the DB user has mustChangePassword=false, that stale session can deserialize as the now-reset admin user.

If this was a real production config I would say we need to fix this, but its only ever used for testing.  Arguably we should remove the file-based db and test from docker... but I will leave that for another PR :)

jescalada:
Thanks for the review @andypols! Denis pointed out the session issue as well - and I also figured we could ignore the issue for now since it's an edge case within local auth, and nobody really uses local auth to begin with.

I'm sure this will come up again during a formal security review of GitProxy (coming soon!) - I promise to fix it up when that happens.

PR #1458, opened 2026-03-16
feat: add upstream proxy configuration for outbound requests

### LINKED ISSUE #759: Expose proxy support within GitProxy itself for air gapped environments
It's very common in organizations to have web proxies configured for developer machines for security. Currently, GitProxy itself does not expose a way to enable "proxying the proxy" - that is, when GitProxy forwards requests to the upstream destinations, it presumes direct Internet access with no ability to hook in `http_proxy` settings.

It is supported in the middleware by overriding the `agent` property in the reqOpt via the `proxyReqOptDecorator` function. See https://github.com/villadora/express-http-proxy/pull/337#issuecomment-391757860 for details.

A GitProxy user should be able to specify the appropriate HTTP proxy agent options via configuration to enable the git server to connect to a destination via an external proxy.

There are no other viable alternatives with the current middleware in use. The current design presumes that the GitProxy server itself has direct network access to upstream locations (https://github.com). For some organizations, this may not be compatible with their network security.

**Additional context**
This was reported by a user getting started with the project so it's best to include it in our introductory documentation.

Andreybest:
Resolves #759.

Adds upstream proxy support. Uses values from both configuration file in `upstreamProxy` fields or environment variables (`HTTPS_PROXY/HTTP_PROXY`, `NO_PROXY`)


config.schema.json
```diff
@@ -367,6 +367,29 @@
           }
         }
       }
+    },
+    "upstreamProxy": {
```
jescalada, line 371: Sample values or just empty entries should be added to `proxy.config.json` so it's easier to see that the `upstreamProxy` is configurable and how to configure it:

```json
"upstreamConfig": {
  "enabled": true,
  "url": "",
  "noProxy": []
}
```

src/proxy/routes/index.ts
```diff
@@ -144,7 +147,100 @@ const getRequestPathResolver: (prefix: string) => ProxyOptions['proxyReqPathReso
...
+    return false;
+  });
+};
+
+const buildUpstreamProxyAgent = (
+  proxyReqOpts: Omit<RequestOptions, 'headers'> & {
+    headers: OutgoingHttpHeaders;
+  },
+) => {
+  const upstreamProxyConfig = getUpstreamProxyConfig();
```
jescalada: Refactored for simplicity:

```suggestion
  const { enabled, url, noProxy } = getUpstreamProxyConfig();
```

src/proxy/routes/index.ts
```diff
@@ -144,7 +147,100 @@ const getRequestPathResolver: (prefix: string) => ProxyOptions['proxyReqPathReso
...
+
+  const configuredUrl = upstreamProxyConfig.url;
+  const envUrl = getEnvProxyUrl();
+
+  const proxyUrl = configuredUrl || envUrl;
```
jescalada: Short-circuiting so we don't execute getEnvProxyUrl() all the time

```suggestion
  const proxyUrl = url || getEnvProxyUrl();
```

(must also remove `const envUrl = getEnvProxyUrl();` from above)

src/proxy/routes/index.ts
```diff
@@ -144,7 +147,100 @@ const getRequestPathResolver: (prefix: string) => ProxyOptions['proxyReqPathReso
...
+};
+
+const buildUpstreamProxyAgent = (
+  proxyReqOpts: Omit<RequestOptions, 'headers'> & {
+    headers: OutgoingHttpHeaders;
+  },
+) => {
+  const upstreamProxyConfig = getUpstreamProxyConfig();
+
+  const configuredUrl = upstreamProxyConfig.url;
```
jescalada:
```suggestion
```

src/proxy/routes/index.ts
```diff
@@ -144,7 +147,100 @@ const getRequestPathResolver: (prefix: string) => ProxyOptions['proxyReqPathReso
...
+  const envUrl = getEnvProxyUrl();
```
jescalada:
```suggestion
```

src/proxy/routes/index.ts
```diff
@@ -144,7 +147,100 @@ const getRequestPathResolver: (prefix: string) => ProxyOptions['proxyReqPathReso
...
+  const proxyUrl = configuredUrl || envUrl;
+
+  // If nothing is configured, do not use a proxy
+  if (!proxyUrl) {
+    return undefined;
+  }
+
+  // If config explicitly disabled the proxy, do not use it
+  if (upstreamProxyConfig.enabled === false) {
+    return undefined;
```
jescalada, line 241: Can be combined into a single check:

```ts
if (!enabled || !proxyUrl)
```

src/proxy/routes/index.ts
```diff
@@ -144,7 +147,100 @@ const getRequestPathResolver: (prefix: string) => ProxyOptions['proxyReqPathReso
...
+  },
+) => {
+  const upstreamProxyConfig = getUpstreamProxyConfig();
+
+  const configuredUrl = upstreamProxyConfig.url;
+  const envUrl = getEnvProxyUrl();
+
+  const proxyUrl = configuredUrl || envUrl;
+
+  // If nothing is configured, do not use a proxy
```
jescalada: Code here is self-descriptive 🙂 

```suggestion
```

src/proxy/routes/index.ts
```diff
@@ -144,7 +147,100 @@ const getRequestPathResolver: (prefix: string) => ProxyOptions['proxyReqPathReso
...
+  if (!proxyUrl) {
+    return undefined;
+  }
+
+  // If config explicitly disabled the proxy, do not use it
```
jescalada: Same as above

```suggestion
```

src/proxy/routes/index.ts
```diff
@@ -144,7 +147,100 @@ const getRequestPathResolver: (prefix: string) => ProxyOptions['proxyReqPathReso
...
+  if (upstreamProxyConfig.enabled === false) {
+    return undefined;
+  }
+
+  const host: string | null | undefined = proxyReqOpts.host || proxyReqOpts.hostname;
+
+  const configNoProxy = upstreamProxyConfig.noProxy ? upstreamProxyConfig.noProxy : [];
```
jescalada: Not sure what was the intention here 🤔

src/proxy/routes/index.ts
```diff
@@ -144,7 +147,100 @@ const getRequestPathResolver: (prefix: string) => ProxyOptions['proxyReqPathReso
...
+  const envNoProxy = getEnvNoProxyList();
+  const combinedNoProxy = [...configNoProxy, ...envNoProxy];
```
jescalada: Do we want to combine the configuration + env values, or do we want to default to the env values if the `noProxy` from the config is missing despite `upstreamProxy` being enabled?

jescalada: If we want to combine, this could be done in a single line like such:

```ts
  const combinedNoProxy = [...noProxy || [], ...getEnvNoProxyList()]
```

If we want to use the env values as fallbacks, we can further simplify by stuffing everything in the if-statement below:

```ts
  if (hostMatchesNoProxy(host, noProxy || getEnvNoProxyList())) {
    return undefined;
  }
```

test/upstreamProxy.test.ts
jescalada, line 1: Let's aim for >90% coverage 😃

jescalada (changes requested):
@Andreybest Thanks for the contribution! 🚀 

I left some comments for code refactoring. It'd be very helpful if you could upload images or a video demonstrating that this feature works as intended.

@coopernetes Wondering if this implements your original requirements (#759) and works on your end 🤔

src/proxy/routes/index.ts
```diff
@@ -144,7 +147,100 @@ const getRequestPathResolver: (prefix: string) => ProxyOptions['proxyReqPathReso
...
+  if (!noProxy) {
+    return [];
+  }
+  return noProxy
+    .split(',')
+    .map((entry) => entry.trim())
+    .filter((entry) => entry.length > 0);
+};
+
+const hostMatchesNoProxy = (host: string | null | undefined, noProxyList: string[]): boolean => {
```
dcoric, line 167: `hostMatchesNoProxy` does not handle two standard `NO_PROXY` conventions that are universally used in corporate environments - exactly the target audience for this feature:

1. **Wildcard `*`** - means "bypass proxy for all hosts." Currently `*` would fall through to the `.endsWith` check and never match.
2. **Leading-dot patterns like `.example.com`** - the current `` .endsWith(`.${trimmed}`) `` logic would look for `..example.com`, which never matches anything.

These conventions are supported by curl, wget, Go stdlib, Python `requests`, and Node's `global-agent`. Since this feature targets corporate proxy environments, both patterns will appear in real `NO_PROXY` values.

Suggested fix:
```ts
const trimmed = pattern.trim().replace(/^\./, ''); // strip leading dot
if (trimmed === '*') return true; // wildcard - bypass all
```

src/proxy/routes/index.ts
```diff
@@ -144,7 +147,100 @@ const getRequestPathResolver: (prefix: string) => ProxyOptions['proxyReqPathReso
...
+
+  const configNoProxy = upstreamProxyConfig.noProxy ? upstreamProxyConfig.noProxy : [];
+  const envNoProxy = getEnvNoProxyList();
+  const combinedNoProxy = [...configNoProxy, ...envNoProxy];
+
+  if (hostMatchesNoProxy(host, combinedNoProxy)) {
+    return undefined;
+  }
+
+  return new HttpsProxyAgent(proxyUrl);
```
dcoric: The `proxyUrl` (from config or env vars) is passed directly to `new HttpsProxyAgent(proxyUrl)` without validation. If the URL is malformed (typo, missing protocol, empty string after trimming), this throws a cryptic `Invalid URL` error that surfaces through `express-http-proxy`'s error handler with no indication that the upstream proxy config is the problem.

Consider wrapping this in a try-catch or validating first:
```ts
try {
  new URL(proxyUrl); // validate before constructing agent
} catch {
  throw new Error(`Invalid upstream proxy URL: check your upstreamProxy.url config or HTTPS_PROXY env var`);
}
return new HttpsProxyAgent(proxyUrl);
```

website/docs/architecture/architecture.md
```diff
@@ -222,6 +222,26 @@ Currently supports the following out-of-the-box:
 - ActiveDirectory auth configuration for querying via a REST API rather than LDAP
 - Gitleaks configuration
 
+#### `upstreamProxy`
+
+Configures routing of outbound requests from the GitProxy server to upstream Git hosts (e.g. GitHub, GitLab) via an HTTP(S) proxy. Use this when the server runs in an environment where direct Internet access is not allowed and all traffic must go through a corporate web proxy ("proxying the proxy").
+
+- **`enabled`** (boolean): When `true`, outbound connections to upstream Git hosts use the configured proxy. When `false`, the proxy is not used even if `url` or environment variables are set.
+- **`url`** (string): The HTTP(S) proxy URL (e.g. `http://proxy.corp.local:8080` or `http://user:pass@proxy.corp.local:8080`). If omitted, GitProxy falls back to the `HTTPS_PROXY`, `https_proxy`, `HTTP_PROXY` or `http_proxy` environment variables (first defined wins).
```
dcoric, line 230: The docs show `http://user:pass@proxy.corp.local:8080` as a valid proxy URL format, which is correct - but the code has no redaction layer for this value. The `proxyUrl` containing credentials flows through `buildUpstreamProxyAgent` and into the request options. If `express-http-proxy` debug logging is enabled, or if `logConfiguration` prints the full config at startup, credentials will appear in plaintext in logs.

At minimum, consider:
- Adding a warning comment in the code that `proxyUrl` may contain credentials and must not be logged
- If any logging is added in the future around proxy setup, use a helper that redacts the userinfo portion of the URL (everything between `://` and `@`)

src/proxy/routes/index.ts
```diff
@@ -144,7 +147,100 @@ const getRequestPathResolver: (prefix: string) => ProxyOptions['proxyReqPathReso
...
+};
+
+const proxyReqOptDecorator: ProxyOptions['proxyReqOptDecorator'] = (proxyReqOpts, _srcReq) => {
+  const agent = buildUpstreamProxyAgent(proxyReqOpts);
```
dcoric, line 256: `buildUpstreamProxyAgent` is called from `proxyReqOptDecorator` on every proxied request. Each invocation creates a `new HttpsProxyAgent(proxyUrl)` and also calls `getUpstreamProxyConfig()` which runs `loadFullConfiguration()`. For high-throughput operations like large clones with many pack requests, this creates unnecessary object allocation and GC pressure.

Consider caching the agent instance and only recreating it when the proxy URL or configuration changes - for example, using a simple memoization keyed on `proxyUrl`.

dcoric (changes requested):
Thanks for tackling #759. I've left a few comments focused on correctness and robustness, especially around `NO_PROXY` pattern handling which is important to get right for the target use case. The main items are the missing `*`/leading-dot `NO_PROXY` support and proxy URL validation.

config.schema.json
Andreybest: Added

src/proxy/routes/index.ts
Andreybest: Refactored

src/proxy/routes/index.ts
Andreybest: Added

src/proxy/routes/index.ts
Andreybest: Removed

src/proxy/routes/index.ts
Andreybest: Removed

src/proxy/routes/index.ts
Andreybest: Added

src/proxy/routes/index.ts
Andreybest: Removed

src/proxy/routes/index.ts
Andreybest: Removed

src/proxy/routes/index.ts
Andreybest: I think we need to combine, selected first approach :)

src/proxy/routes/index.ts
Andreybest: Added

src/proxy/routes/index.ts
Andreybest: Added

website/docs/architecture/architecture.md
Andreybest: Added warnings in comments. Also added function for reduction if it will be used in future

src/proxy/routes/index.ts
Andreybest: Considering that loadFullConfiguration() returns cached object:

```ts
function loadFullConfiguration(): GitProxyConfig {
  if (_currentConfig) {
    return _currentConfig;
  }

```

I don't think we don't have to worry about it, for new HttpsProxyAgent(proxyUrl) - cache is implemented

src/proxy/routes/index.ts
Andreybest: Resolved in comments below

src/config/index.ts
```diff
@@ -149,6 +149,23 @@
...
 
+/**
+ * Redacts the userinfo (credentials) from a proxy URL for safe logging.
+ * e.g. http://user:pass@proxy.corp.local:8080 → http://<redacted>@proxy.corp.local:8080
+ *
+ * WARNING: proxyUrl may contain plaintext credentials in the userinfo portion.
+ * Never log a raw proxy URL — always pass it through this helper first.
+ */
+export const redactProxyUrl = (url: string): string => {
+  return url.replace(/(https?:\/\/)[^@]+@/, '$1<redacted>@');
```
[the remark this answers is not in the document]
Andreybest: Should be resolved

proxy.config.json
```diff
@@ -156,6 +156,11 @@
       }
     }
   ],
+  "upstreamProxy": {
+    "enabled": true,
```
coopernetes: This should be disabled by default. This configuration affects testing in CI as well as it is the default set of configuration settings shipped by both the NPM package of GitProxy as well as the Docker image. It has to work "out of the box" without requiring _another_ process or service (the external proxy in this case) to be running as well.

Also, as of this review comment, this being set to true by default is likely the cause of all the CI failures which have to be resolved before merging.

```suggestion
    "enabled": false,
```

Andreybest: That's makes sense. Thanks @coopernetes ! Applied your change.

test/upstreamProxy.test.ts
Andreybest: 89.87%, hopefully this passes :)

jescalada: Great, thanks!

src/proxy/routes/index.ts
```diff
@@ -144,7 +147,104 @@ const getRequestPathResolver: (prefix: string) => ProxyOptions['proxyReqPathReso
...
+};
+
+// WARNING: proxyUrl may contain plaintext credentials in the userinfo portion
+// (e.g. http://user:pass@proxy.corp.local:8080). Never log it directly — use
+// redactProxyUrl() from config for any log statements involving this value.
+let _cachedProxyAgent: { proxyUrl: string; agent: HttpsProxyAgent<string> } | null = null;
+
+const getOrCreateProxyAgent = (proxyUrl: string): HttpsProxyAgent<string> => {
+  if (!_cachedProxyAgent || _cachedProxyAgent.proxyUrl !== proxyUrl) {
+    _cachedProxyAgent = { proxyUrl, agent: new HttpsProxyAgent(proxyUrl) };
```
jescalada, line 214: Just repeating [Denis earlier comment](https://github.com/finos/git-proxy/pull/1458#discussion_r2972143535) - this should be properly validated and/or throw a descriptive error

src/proxy/routes/index.ts
jescalada: This comment might still need fixing! 🙂

src/proxy/routes/index.ts
jescalada: Looked at this in depth and as Andrey mentioned both `loadFullConfiguration()` and `buildUpstreamProxyAgent()` are cached - I don't think there are any concerns unless performance issues come up when the feature is enabled.

jescalada (commented):
LGTM after fixing Denis' comment about validating/error handling the `HttpsProxyAgent` creation.

@coopernetes Does everything work as you expected? 🤔

src/proxy/routes/index.ts
Andreybest: Don't know how I've missed it, added, thanks @jescalada and @dcoric !

src/proxy/routes/index.ts
Andreybest: Added, thanks for remind :)

src/proxy/routes/index.ts
```diff
@@ -144,7 +147,111 @@ const getRequestPathResolver: (prefix: string) => ProxyOptions['proxyReqPathReso
...
+const buildUpstreamProxyAgent = (
+  proxyReqOpts: Omit<RequestOptions, 'headers'> & {
+    headers: OutgoingHttpHeaders;
+  },
+) => {
+  const { enabled, url, noProxy } = getUpstreamProxyConfig();
+
+  const proxyUrl = url || getEnvProxyUrl();
+
+  if (enabled === false || !proxyUrl) {
```
dcoric: I think the env-var-only path is still broken. `proxy.config.json` is imported as the default config, and it now contains `upstreamProxy.enabled: false`. When a user omits `upstreamProxy` from their config and relies only on `HTTPS_PROXY`/`HTTP_PROXY`, the merged config still has `enabled === false`, so `buildUpstreamProxyAgent()` returns before using the env var. This contradicts the docs saying env vars enable proxying when `upstreamProxy` is not configured. The current test mocks `getUpstreamProxyConfig()` as `{}`, so it does not cover the real default-config path. Can we either remove `upstreamProxy` from the shipped default config, distinguish “unset” from explicitly disabled, or change the documented behavior and add a test for the default merge?

src/proxy/routes/index.ts
```diff
@@ -144,7 +147,111 @@ const getRequestPathResolver: (prefix: string) => ProxyOptions['proxyReqPathReso
...
+
+// WARNING: proxyUrl may contain plaintext credentials in the userinfo portion
+// (e.g. http://user:pass@proxy.corp.local:8080). Never log it directly — use
+// redactProxyUrl() from config for any log statements involving this value.
+let _cachedProxyAgent: { proxyUrl: string; agent: HttpsProxyAgent<string> } | null = null;
+
+const getOrCreateProxyAgent = (proxyUrl: string): HttpsProxyAgent<string> => {
+  if (!_cachedProxyAgent || _cachedProxyAgent.proxyUrl !== proxyUrl) {
+    try {
+      new URL(proxyUrl);
```
dcoric: This validation still accepts values that are syntactically parseable URLs but not valid HTTP(S) proxy URLs. For example, `new URL("socks5://proxy.example.com:1080")`, `new URL("ftp://proxy.example.com:21")`, and even `new URL("localhost:8081")` all succeed. `HttpsProxyAgent` also constructs for those values, then treats non-`https:` schemes as plain TCP proxy connections, so the user gets a later connectivity failure instead of a descriptive config error. Can we parse once, require `protocol` to be `http:` or `https:`, require a non-empty `hostname`, and add tests for unsupported schemes and missing protocol?

coopernetes (commented):
LGTM. The only rub is the lack of direct authentication support in the proxy agent. Many client and server libraries do not implement auth directly so it's fine to leave it out for now. Worth making a follow up issue to add HTTP Basic and NTLM (Windows) auth support which are the two most common methods (at least that I'm aware of).

kriswest:
> Worth making a follow up issue to add HTTP Basic and NTLM (Windows) auth support which are the two most common methods (at least that I'm aware of).

I'd suggest making that HTTP Basic, NTLM and Negotiate (the negotiate protocol is used to detect support for NTLM or Kerberos auth and then kick off which ever is supported).

src/proxy/routes/index.ts
Andreybest: Added checks to accept only https and http, and for hostname to exist.

src/proxy/routes/index.ts
Andreybest: I don't know how I've missed it, but now it's fixed. Unset enable now is treated as false. Thanks!

Andreybest:
@dcoric Thank you again for the review! Can you please check again? :)

kriswest:
@Andreybest could you run teh schema doc generation - you did typescript types gen but not but not config schema docs generation. See https://github.com/finos/git-proxy/blob/main/CONTRIBUTING.md#configuration-schema

Very keen to merge this as it will unblock some testing for us on windows machines. We'll report back with testing in a bit.

kriswest:
> @Andreybest could you run teh schema doc generation - you did typescript types gen but not but not config schema docs generation. See https://github.com/finos/git-proxy/blob/main/CONTRIBUTING.md#configuration-schema

@goldensyntax can probably PR into your branch for this...

goldensyntax:
> > @Andreybest could you run teh schema doc generation - you did typescript types gen but not but not config schema docs generation. See https://github.com/finos/git-proxy/blob/main/CONTRIBUTING.md#configuration-schema
> 
> @goldensyntax can probably PR into your branch for this...

Done that with:
https://github.com/Andreybest/git-proxy/pull/1

Andreybest:
> @Andreybest could you run teh schema doc generation - you did typescript types gen but not but not config schema docs generation. See https://github.com/finos/git-proxy/blob/main/CONTRIBUTING.md#configuration-schema
> 
> Very keen to merge this as it will unblock some testing for us on windows machines. We'll report back with testing in a bit.

Merged @goldensyntax PR
Would be happy to see it in main too :)

Can we merge now?

kriswest (approved):
LGTM

kriswest:
@jescalada will need to approve and merge due to previously requested change.

re-vlad (commented):
While the code supports upstreamProxy config, the user-facing doc does not mention this new feature. New users (especially in corporate environments) will not know how to enable it. Please add a configuration example under the "configuration" section in readme.md: 
```
"upstreamProxy": {
  "enabled": true,
  "url": "http://proxy.corp.local:8080",
  "noProxy": [
    "localhost",
    "127.0.0.1",
    "github.com"
  ]
}```

re-vlad:
> Adds upstream proxy support. Uses values from both configuration file in `upstreamProxy` fields and environment variables (`HTTPS_PROXY/HTTP_PROXY`, `NO_PROXY`)

I'd say 'both configuration file in `upstreamProxy` fields OR environment variables' (according to the src/proxy/routes/index.ts, lines 237–238 `const proxyUrl = url || getEnvProxyUrl();`)

Andreybest:
Thank you for the review @re-vlad!

1. This is already described in website/docs/architecture/architecture.md
2. Changed PR description as you suggested

vzateychuk (approved):

re-vlad (approved):

kriswest (commented):
We should be able to complete a retest of this tomorrow, which we'd like to do before this merges.

jescalada (approved):
@Andreybest LGTM 👍🏼

Do you mind adding an issue for direct authentication support (HTTP Basic, NTLM and Negotiate) as mentioned earlier and referencing to this PR?

kriswest (approved):
Retest completed here and all looks good to us too.

We may need to have a chat about the authentication spec - but we're able to auth with proxy credentials added to the URL currently.

kriswest:
One CLI test failed on windows - re-running that as I'm not convinced it was specific to this PR

Andreybest:
> Retest completed here and all looks good to us too.
> 
> We may need to have a chat about the authentication spec - but we're able to auth with proxy credentials added to the URL currently.

Glad to hear that!

On auth issue - raised #1541. @kriswest @jescalada

PR #1464, opened 2026-03-20
feat: add server-side pagination to API endpoints

### LINKED ISSUE #1297: Improve pagination defaults and layout in admin UI
The built-in admin UI allows you to paginate and filter the repositories, users and dashboard views. However, they default to showing just 5 records in a fairly expansive layout that isn't well suited to managing an enterprises repositories, users and contributions. The layout could be more compact and either a greater default number of records shown or a user control provided to allow more records to show.

Further all pagination is currently client-side. This won't scale. The database clients and adaptors are able to support server-side pagination which we could switch to at the same time.

The pagination setup should be suited to needs of a large enterprise and be able to comfortably support administration of hundreds of repositories, thousands of users and 10s of thousands of contributions.

### LINKED ISSUE #1420: Sort entries in the admin UI and apply proper pagination
Entries in the dashboard and users tabs of the admin UI are not sorted (the repositories tab does have a sorting control, but it doesn't seem to affect the sort order currently). The lack of sorting makes the dashboard, in particular, difficult to use once you have accumulated a number of entries.

(server-side) Pagination should also be applied, we currently retrieve all entries and paginate client-side, which is not sustainable.

The repository, dashboard and user views should all sort the selection of a sort order and page size. These (and the current page number) should be applied to data queries. Additionally, the currently selected status of pushes (all / pending / rejected / cancelled / error) should be applied as a filter to push queries (or we can't figure out the right page size until we receive the data).

- The dashboard should default to sorting by timestamp (most recent first).
- The repository tab should default to alphabetical sorting on project name.
- The users tab should default to alphabetical sorting on username.

Continuing without sorting or server-side pagination is not sustainable. Hence, the alternative is to defer fixing the admin UI is replaced in full.

**Additional context**
The DB clients and API service already support passing parameters to queries (with the exception of the `limit` and `skip` parameters in the API service for some reason). That facility should be used to apply server-side pagination and sorting.

fabiovincenzi:
Fixes https://github.com/finos/git-proxy/issues/1297
Implements server side pagination for `/push`, `/repo` and `/users` endpoints.

 **Note:** The list endpoints (`/push`, `/repo`, `/users`) now return `{ data, total }` instead of a plain array to support pagination metadata.


src/service/routes/utils.ts
```diff
@@ -14,13 +14,31 @@
...
+  const rawLimit = parseInt(req.query['limit'] as string, 10);
+  const rawPage = parseInt(req.query['page'] as string, 10);
+
+  const limit = Math.min(100, Math.max(1, isNaN(rawLimit) ? defaultLimit : rawLimit));
+  const page = Math.max(1, isNaN(rawPage) ? 1 : rawPage);
+  const skip = (page - 1) * limit;
+
+  const pagination: PaginationOptions = { skip, limit };
+  if (req.query['search']) pagination.search = req.query['search'] as string;
+  if (req.query['sortBy']) pagination.sortBy = req.query['sortBy'] as string;
```
dcoric: `sortBy` is taken directly from user input and passed straight to MongoDB's `.sort()` (and NeDB's `.sort()`) without any validation. An attacker could sort by sensitive fields (for example, `password` on the users endpoint) to infer information, or pass bogus field names to cause DB errors.

Please validate `sortBy` against a per-endpoint allowlist of sortable fields before passing it through.

src/service/routes/repo.ts
```diff
@@ -83,7 +82,7 @@ function repo(proxy: Proxy) {
     }
 
     const _id = req.params.id;
-    const username = req.body.username;
+    const username = req.body.username.toLowerCase();
```
dcoric, line 93: If `req.body.username` is `undefined` or `null`, calling `.toLowerCase()` on it will throw a `TypeError` and crash the request handler. The original code passed the value as-is to `db.findUser()`, which would just return null.

Please add a guard before calling `.toLowerCase()` - for example, check that `username` is a non-empty string first, or return a 400 if it is missing.

src/service/routes/push.ts
```diff
@@ -30,22 +31,20 @@ interface AuthoriseRequest {
...
     let parsedValue: boolean | undefined;
     if (rawValue === 'false') parsedValue = false;
     if (rawValue === 'true') parsedValue = true;
-    query[key] = parsedValue ?? rawValue?.toString();
+    query[key] = parsedValue ?? rawValue;
   }
 
-  res.send(await db.getPushes(query));
+  const result = await db.getPushes(query, pagination);
+  res.send(result);
```
dcoric: Worth noting that all three list endpoints (`/push`, `/repo`, `/users`) now return `{ data, total }` instead of plain arrays. This is technically a breaking change in the API. It probably does not matter much since nobody is likely consuming the API directly, but it is worth being aware of and documenting in the release notes.

src/ui/views/RepoDetails/Components/AddUser.tsx
```diff
@@ -90,7 +89,17 @@ const AddUserDialog: React.FC<AddUserDialogProps> = ({
   };
 
   useEffect(() => {
-    getUsers(setIsLoading, setUsers, setAuth, setErrorMessage);
+    const load = async () => {
+      setIsLoading(true);
+      const result = await getUsers();
```
dcoric: `getUsers()` is called here without any pagination params, but the server now applies a default `limit=10` via `parsePaginationParams`. This means the user autocomplete dropdown for adding users to a repo will only show the first 10 users. Previously it loaded all of them.

Either pass a large limit here (or `limit=0` to fetch all), or switch this to a search-as-you-type approach that queries the server as the user types.

src/db/mongo/helper.ts
```diff
@@ -56,6 +57,22 @@ export const connect = async (collectionName: string): Promise<Collection> => {
...
 
+export const paginatedFind = async <T>(
+  collection: Collection,
+  filter: Filter<Document>,
+  sort: Sort,
+  skip: number,
+  limit: number,
+  projection?: Document,
+): Promise<PaginatedResult<T>> => {
+  const total = await collection.countDocuments(filter);
```
dcoric: Nitpick: Internal callers like `checkUserPushPermission` and `proxy/index.ts` only need the data and never use `total`, but every call now runs an additional `countDocuments` query. That is an extra index scan on each internal `getUsers`/`getRepos` call.

Consider making the count conditional - for example, only run `countDocuments` when pagination params are actually provided, and default `total` to `-1` or `data.length` otherwise.

src/ui/views/PushRequests/components/PushesTable.tsx
```diff
@@ -78,43 +83,37 @@ const PushesTable: React.FC<PushesTableProps> = (props) => {
...
-    setFilteredData(filtered);
+  }, [
+    props.blocked,
+    props.canceled,
+    props.authorised,
+    props.rejected,
+    props.errored,
+    currentPage,
+    itemsPerPage,
+    searchTerm,
```
dcoric, line 94: `searchTerm` is in the `useEffect` dependency array, so every keystroke triggers a new server-side API request. The old client-side filtering was instant, but with server-side search you will want to add debounce (300-500ms) to avoid flooding the backend with requests while the user is still typing.

Same issue applies to the search term in `UserList.tsx` and `Repositories.tsx`.

dcoric (changes requested):
Thanks for the solid work on server-side pagination. I have a few concerns - mostly around security and a couple of behavioral issues that should be addressed before merging.

roskosm1 (commented):
docs API breaking changes?

src/db/mongo/helper.ts
re-vlad: can force odd/non-indexed sort fields, inconsistent adapter behavior, and slower queries. Please allowlist `sortBy` per endpoint + cap max `page/skip` (or reject excessive offsets).

re-vlad (changes requested):

src/db/file/helper.ts
```diff
@@ -15,8 +15,40 @@
...
+      .skip(skip)
+      .limit(limit)
+      .exec((err: Error | null, docs: any[]) => {
+        /* istanbul ignore if */
+        if (err) reject(err);
+        else resolve(docs);
+      });
+  });
+
+  return Promise.all([dataPromise, countPromise]).then(([data, total]) => ({ data, total }));
```
jescalada: I'm wondering if we should change the `data` variable to something more descriptive... Right now, we end up calling `response.data.data` to fetch the repos/users/etc. which can trip up people/agents working with the code.

I'd suggest having the name of the entity `{ users, total }`, etc. although we'd need to pass in the entity name to `paginatedFind` (unless it's already obtainable in the existing function parameters).

packages/git-proxy-cli/index.ts
```diff
@@ -106,12 +106,12 @@ async function getGitPushes(filters: Partial<PushQuery>) {
 
   try {
     const cookies = JSON.parse(fs.readFileSync(GIT_PROXY_COOKIE_FILE, 'utf8'));
-    const { data } = await axios.get<Action[]>(`${baseUrl}/api/v1/push/`, {
+    const response = await axios.get<{ data: Action[]; total: number }>(`${baseUrl}/api/v1/push/`, {
```
jescalada: Can use `PaginatedResult` type here

src/db/mongo/pushes.ts
```diff
@@ -30,34 +30,51 @@ const defaultPushQuery: Partial<PushQuery> = {
   type: 'push',
 };
 
+const pushProjection = {
```
jescalada, line 33: Perhaps this could be a constant (and potentially placed somewhere else)? Wondering if there are any other constants from the `db` module that we could wrap into a `/db/constants.ts` file.

On another note, I'm wondering why we have been doing manual projection at all - if we want to exclude certain fields, we could do post-processing just like we do with users (`toPublicUser`).

jescalada: Fixing this should also help simplify the tests that rely on projection 👍🏼

src/service/routes/utils.ts
```diff
@@ -14,13 +14,31 @@
...
+export const parsePaginationParams = (req: Request, defaultLimit = 10): PaginationOptions => {
+  const rawLimit = parseInt(req.query['limit'] as string, 10);
+  const rawPage = parseInt(req.query['page'] as string, 10);
+
+  const limit = Math.min(100, Math.max(1, isNaN(rawLimit) ? defaultLimit : rawLimit));
+  const page = Math.max(1, isNaN(rawPage) ? 1 : rawPage);
+  const skip = (page - 1) * limit;
+
+  const pagination: PaginationOptions = { skip, limit };
+  if (req.query['search']) pagination.search = req.query['search'] as string;
```
jescalada: Wondering if we could have stronger typing for `req` here - so we could do the following:

```ts
const { search, sortBy, sortOrder } = req.query;
```

Extending the `Request` type can be a huge pain, but it's been done before with the `user` property for the passport strategies. This might be more appropriate for a separate PR if it requires too many extra changes!

jescalada (changes requested):
Oddly, the repo list sorting doesn't seem to be working right on my end - whenever I change the sorting order, the dropdown selector just resets (and only the default alphabetical order is shown regardless of which one I choose).

Related to that, when I move to another page the selected filter gets reset, which (might) be resetting the filter to alphabetical ascending order.

I don't get any error output on console, BTW 🤷🏼

src/service/routes/push.ts
fabiovincenzi: Good point. I noted the change in the PR description. Since we probably want this PR post-v2, we'll call it out manually in the release notes when the time comes. We don't have a strong convention for flagging breaking changes, the only automatic mechanism is the `break:` prefix in the PR title, but that would trigger a major version bump in release-drafter, which may not be desirable here.

src/db/file/helper.ts
fabiovincenzi: Good point, the HTTP routes now return entity-specific keys (pushes, repos, users) instead of data. The data variable inside paginatedFind remains as a generic internal name but not exposed directly.

packages/git-proxy-cli/index.ts
fabiovincenzi: the `/push `endpoint now returns pushes instead of data.

src/db/mongo/pushes.ts
fabiovincenzi: I'm not sure about the reasons for projections here, filtering at the DB level avoids transferring data that is not needed, such as steps which can be quite large. Post-processing like toPublicUser, would work but at the cost of fetching full documents first. Not sure though if this is the reason                                                     
                                                                                                                               
Would it be worth tracking in a separate issue/PR to keep this one focused?

src/service/routes/utils.ts
fabiovincenzi: Done, introduced a `PaginationQuery` interface and cast `req.query` once at the top, allowing destructuring without repeated `as string` casts. I didn't make the global Request type since query params vary per route.

src/db/mongo/pushes.ts
jescalada: It's a minor thing which will probably stay very long in our backlog - and I'm inclined to ignore it so this PR doesn't keep growing... 🙂

kriswest:
According to @andypols #1507 doesn't introduce server-side pagination - so this will still be needed, but perhaps should merge after #1507, and some small changes to use it will be needed.

I haven't checked but I suspect the counts in the UI in #1507 may need to be calculated a different way if the entire dataset is not returned. Perhaps we add a repo stats API call or calculate the stats and add to the repo API itself - just speculating.

jescalada (commented):
I just did an in-depth review of @andypols' UI refactor #1507, and also took another look at this PR. The UI refactor revamps the entire sorting UI, but as far as I can see, doesn't implement server-side pagination.

The changes in #1507 are massive (and have been in the works for longer than this PR), so I'd suggest merging that in first and then modifying this PR to get pagination to work on the new UI. Wonder if there are any alternatives? 🤔

fabiovincenzi:
> I just did an in-depth review of @andypols' UI refactor #1507, and also took another look at this PR. The UI refactor revamps the entire sorting UI, but as far as I can see, doesn't implement server-side pagination.
> 
> The changes in #1507 are massive (and have been in the works for longer than this PR), so I'd suggest merging that in first and then modifying this PR to get pagination to work on the new UI. Wonder if there are any alternatives? 🤔

@jescalada I think it's easier as you mentioned, merging https://github.com/finos/git-proxy/pull/1507 first, then adjusting this over those changes

kriswest:
Sounds good to me - and presumably would be fine with @andypols. Any thoughts on the counts displayed in the new UI's repository view? Will we need an additional API to calculate & retrieve those?

andypols:
@fabiovincenzi @kriswest @fabiovincenzi - I agree it makes sense to merge the ui updates first.  We deliberately avoided changing the db as the data is small enough to manipulate in memory.  

> Any thoughts on the counts displayed in the new UI's repository view? Will we need an additional API to calculate & retrieve those?

I don't think it needs to be an additional API.  Most databases can return a page of rows plus the total count in a single query.

Issue #1210, opened 2025-09-22
Alternative email support in v2


06kellyjac:
**Is your feature request related to a problem? Please describe.**

As part of V2 we've moved to committer email being the determining factor for tying changes to a user.
In order to support cases where a user's email used to log in doesn't match the email used to make commits changes will be necessary.

**Describe the solution you'd like**

There are a few approaches we might want to implement:

* Users manually add alternative emails themselves
  * They're initially unverified & there's a process to verify ownership
    * could email a code that they enter
    * could email a link they click
* Provide a API for registering alternative emails on behalf of users
* Scrape an API for alternative emails

This should also probably be configurable:

* Fully disable alternative emails
* Allow specific methods of registration
* Probably rules on what emails can be added
  * Maybe unnecessary as enforced by commit config already?
  * Maybe apply the exact same rules when alt emails are submitted?
* Potentially optionally allowing unverified emails (default would be verified only)
  * could be a stopgap and easiest implementation, wouldn't be worse than the GH username used previously.

Basic data:

* alternativeEmails: list
  * email: string
  * verified: boolean

Could also have a timestamp for verification and have a re-verification mechanism but overkill IMO

**Describe alternatives you've considered**

Changing how user's auth is a bit tricky. Other than that there's not much of an alternative beyond the implementation options above.

**Additional context**

PR #1496, opened 2026-04-07
feat: add generic LDAP authentication strategy using ldapts

1saac-k:
Closes #1488

Add a new `ldap` authentication type that integrates with any standards-compliant LDAP server. The existing
`activedirectory` type relies on the `activedirectory2` library which is AD-specific and incompatible with lightweight
LDAP servers (e.g. [lldap](https://github.com/lldap/lldap)). This PR introduces a generic LDAP strategy built on `ldapts`
and `passport-custom`, lowering the barrier to entry for teams that want simple, manageable user authentication without
setting up a full Active Directory.

This implementation is not `lldap`-specific — it supports any standards-compliant LDAP server, which means it should also work with Active Directory. As a result, it may be possible in the future to replace the deprecated `activedirectory2` and `ldapjs` dependencies with `ldapts`.

Testing

- 11 unit tests added covering authentication success/failure scenarios.
- Manually verified end-to-end against a running [lldap](https://github.com/lldap/lldap) instance.


1saac-k:
Since this work was done outside of my company, I changed the commit email to my personal email and added GPG signatures to sign the CLA as an individual contributor.

dcoric (changes requested):
Requesting changes for the two required fixes in the inline comments:

- LDAP group lookup errors are currently swallowed as “not a member”.
- Missing LDAP email falls back to an empty string, which can break user upserts with the file DB unique email index.

The remaining inline comments are nits.

config.schema.json
```diff
@@ -505,6 +505,90 @@
           },
           "required": ["type", "enabled", "oidcConfig"]
         },
+        {
+          "title": "LDAP Auth Config",
```
dcoric, line 679: **nit, non-blocking** Since the config schema changed, please consider regenerating `website/docs/configuration/reference.mdx` as described in `CONTRIBUTING.md`. The generated website schema reference otherwise will not include the new LDAP auth config.

src/service/passport/ldap.ts
```diff
@@ -0,0 +1,277 @@
...
+    // Step 3: Search for the user entry
+    const entry = await searchUser(client, ldapConfig, username);
+    if (!entry) {
+      return null;
+    }
+
+    const userDN = entry.dn as string;
+
+    // Step 4: Check user group membership
+    const isMember = await isUserInGroup(client, ldapConfig, userDN, ldapConfig.userGroupDN);
```
dcoric: **nit, non-blocking** The implementation checks user/admin group membership before verifying the submitted password. I would normally verify the user bind immediately after resolving the user DN, then do group authorization only for authenticated credentials. That avoids extra LDAP work for bad-password attempts and reduces timing/log side effects.

src/service/routes/auth.ts
```diff
@@ -52,7 +53,7 @@ router.get('/', (_req: Request, res: Response) => {
 });
 
 // login strategies that will work with /login e.g. take username and password
-const appropriateLoginStrategies = [passportLocal.type, passportAD.type];
+const appropriateLoginStrategies = [passportLocal.type, passportAD.type, passportLdap.type];
```
dcoric, line 80: **nit, non-blocking** I do not think the CLI should implement LDAP directly: it already posts username/password to `/api/auth/login`, so it should work with LDAP when LDAP is the selected server-side username/password strategy. It may be useful to mention in docs or tests that `git-proxy-cli login` works with configured username/password auth methods, while `git-proxy-cli create-user` is still local DB user provisioning and does not create LDAP users.

src/service/passport/ldap.ts
```diff
@@ -0,0 +1,277 @@
...
+  groupDN: string,
+): Promise<boolean> => {
+  const groupFilter = (ldapConfig.groupSearchFilter || '(member={{dn}})').replaceAll(
+    '{{dn}}',
+    escapeFilterValue(userDN),
+  );
+
+  const searchBase = ldapConfig.groupSearchBase || groupDN;
+
+  try {
```
dcoric: `isUserInGroup()` catches every LDAP search error and returns `false`, so LDAP outages, invalid filters, bad search bases, or insufficient bind permissions all look the same as "user is not in the required group." For the required `userGroupDN` check, that should fail authentication as an LDAP/config error rather than silently denying the user. This would also make the admin-group `try/catch` below meaningful: let the helper throw, treat user-group lookup errors as authentication errors, and only downgrade admin-group lookup errors to `admin=false` if that is the intended best-effort behavior.

src/service/passport/ldap.ts
```diff
@@ -0,0 +1,277 @@
...
+    if (!lc[field]) {
+      throw new Error(`LDAP configuration field "${field}" is required but empty`);
+    }
+  }
+
+  return lc;
+};
+
+const createClient = (ldapConfig: LdapConfig): Client => {
+  return new Client({
```
dcoric, line 73: **nit, non-blocking** It may be worth setting or exposing LDAP operation bounds. `ldapts` supports client timeouts and search options such as `timeLimit`/`sizeLimit`; without them, a slow or misconfigured LDAP server can leave login requests waiting longer than expected or return unbounded matches.

src/service/passport/ldap.ts
```diff
@@ -0,0 +1,277 @@
...
+
+    const passwordValid = await verifyPassword(ldapConfig, userDN, password);
+    if (!passwordValid) {
+      return null;
+    }
+
+    // Step 7: Extract profile attributes and sync to database
+    const userObj = {
+      username: String(entry[usernameAttr] || username).toLowerCase(),
+      email: String(entry[emailAttr] || '').toLowerCase(),
```
dcoric: `email` falls back to an empty string when the configured LDAP email attribute is missing. The file DB has a unique email index, so two LDAP users without `mail` can both authenticate successfully against LDAP but the second one can fail during `db.updateUser()` because it attempts to upsert another user with `email: ""`. Could we either require the configured email attribute before updating the user, omit `email` when it is absent, or derive a stable fallback that remains unique per user?

src/service/passport/ldap.ts
```diff
@@ -0,0 +1,277 @@
...
+ * Check if a user is a member of a specific group by searching for a group
+ * entry that references the user's DN.
+ */
+export const isUserInGroup = async (
+  client: Client,
+  ldapConfig: LdapConfig,
+  userDN: string,
+  groupDN: string,
+): Promise<boolean> => {
+  const groupFilter = (ldapConfig.groupSearchFilter || '(member={{dn}})').replaceAll(
```
dcoric: **nit, non-blocking** For broader "generic LDAP" support, `groupSearchFilter` could also support a `{{username}}` placeholder in addition to `{{dn}}`. That would cover schemas such as `posixGroup` / `memberUid` without forcing group membership to reference the user DN.

1saac-k:
@dcoric Thank you for the detailed review. I'm a bit tied up right now, but I'll take a look within the next few days!

config.schema.json
1saac-k: Updated. I regenerated `website/docs/configuration/reference.mdx`.

src/service/passport/ldap.ts
1saac-k: Updated. The user bind now happens immediately after resolving the user DN, before user/admin group authorization checks.

src/service/passport/ldap.ts
1saac-k: Updated. isUserInGroup() now lets LDAP search errors propagate. Required user-group lookup failures fail authentication as LDAP/config errors, while admin-group lookup errors are still downgraded to `admin=false`.

src/service/passport/ldap.ts
1saac-k: Updated. Added configurable LDAP client timeout/connectTimeout and search timeLimit/sizeLimit options.

src/service/passport/ldap.ts
1saac-k: Updated. LDAP authentication now fails if the configured email attribute is missing, avoiding upserts with email: `""`.

src/service/passport/ldap.ts
1saac-k: Updated. `groupSearchFilter` now supports both `{{dn}}` and `{{username}}`.

src/service/routes/auth.ts
1saac-k: Updated. I clarified the CLI docs/help (new commit: fb0c588)

1saac-k:
I addressed all review comments and rebased onto the latest main branch. Most of the changes were squashed into the existing commits, and one additional commit (fb0c588) was added.

src/service/passport/ldap.ts
```diff
@@ -0,0 +1,298 @@
...
+ */
+
+import { Client } from 'ldapts';
+import { Strategy as CustomStrategy } from 'passport-custom';
+import type { PassportStatic } from 'passport';
+import type { Request } from 'express';
+
+import * as db from '../../db';
+import { getAuthMethods } from '../../config';
+import { LdapConfig } from '../../config/generated/config';
```
dcoric: Thanks for the fixes, the earlier LDAP auth concerns look addressed and the shape looks good overall. One thing caught my eye: `config.ts` now exports `LDAPConfig`, not `LdapConfig`, and I confirmed `npm run check-types:server` fails with TS2724 on this import. Could you update this import/type usage to `LDAPConfig` so the server typecheck passes?

1saac-k: Oh, my bad. I re-ran the items under CONTRIBUTION.md → Submitting a Pull Request → Verify locally.

- `npm run check-types:server` — the error you mentioned occurred, and I fixed it. (Squashed into 4522f33c.)
- `npm test` — 6 tests fail. However, the same is true on the main branch. (Not caused by my changes.)
- `npm run lint` — OK
- `npm run format:check` — OK

Logs are below (after fix a check-types error).

```bash
(0db) ~/ws/git-proxy $ npm run check-types:server

> @finos/git-proxy@2.0.0-rc.6 check-types:server
> tsc --project tsconfig.publish.json --noEmit

(0db) ~/ws/git-proxy $ npm test

> @finos/git-proxy@2.0.0-rc.6 test
> cross-env NODE_ENV=test vitest --run --dir ./test
.. snip ..
 Test Files  1 failed | 59 passed | 3 skipped (63)
      Tests  6 failed | 774 passed | 45 skipped (825)
   Start at  14:34:59
   Duration  21.28s (transform 806ms, setup 0ms, collect 2.31s, tests 18.18s, environment 0ms, prepare 61ms)

(0db) ~/ws/git-proxy $ npm run lint

> @finos/git-proxy@2.0.0-rc.6 lint
> eslint

(0db) ~/ws/git-proxy $ npm run format:check

> @finos/git-proxy@2.0.0-rc.6 format:check
> prettier --check "**/*.{js,jsx,ts,tsx,json,md,yml,yaml,css,scss}" --ignore-path .gitignore --config ./.prettierrc

Checking formatting...
All matched files use Prettier code style!
```

1saac-k:
This PR is still alive. I'm not trying to rush the maintainer; I'm just letting you know that this PR is not stale.

src/service/passport/ldap.ts
```diff
@@ -0,0 +1,298 @@
...
+    );
+    if (!isMember) {
+      console.log(`ldap: user ${username} is not a member of ${ldapConfig.userGroupDN}`);
+      return null;
+    }
+
+    // Step 6: Check optional admin group membership
+    let isAdmin = false;
+    try {
+      isAdmin = await isUserInGroup(client, ldapConfig, userDN, ldapConfig.adminGroupDN, username);
```
jescalada, line 229: Is the admin group membership check optional? If so, we wouldn't want to make `adminGroupDN` a required field 🤔

src/service/passport/ldap.ts
```diff
@@ -0,0 +1,298 @@
...
+
+    const email = entry[emailAttr];
+    if (!email) {
+      throw new Error(`LDAP user "${username}" is missing required email attribute "${emailAttr}"`);
+    }
+
+    // Step 7: Extract profile attributes and sync to database
+    const userObj = {
+      username: String(entry[usernameAttr] || username).toLowerCase(),
+      email: String(email).toLowerCase(),
```
jescalada, line 242: Should we be handling potential array values in here (and other fields, if relevant)? I don't know how emails are typically stored in LDAP, but if array values are allowed to coexist with strings, this would add invalid emails on calling `db.updateUser` later.

Note that GitProxy doesn't currently support multiple emails (#1210), you'd have to either implement support for this, or just select the first entry in case `entry.mail` is an array of strings.

jescalada: Another thing to note: if you're aware of other fields being multi-valued, we ought to have a helper function that preprocesses them so we can write valid database entries 👍🏼 

It could be as simple as:

```ts
Array.isArray(value) ? value[0] : value
```

Although this brings up the question of whether arrays in LDAP are ordered or not and what the default ordering might be...

src/service/passport/ldap.ts
```diff
@@ -0,0 +1,298 @@
...
+      throw new Error(`LDAP configuration field "${field}" is required but empty`);
+    }
+  }
+
+  return lc;
+};
+
+const createClient = (ldapConfig: LDAPConfig): Client => {
+  return new Client({
+    url: ldapConfig.url,
```
jescalada, line 74: Just a comment after scanning the PR with AI (please verify for accuracy before implementing):

Should we throw an error here if `starttls` is disabled but the URL starts with `ldap://`? If `starttls` is disabled, the `verifyPassword` function below would send the user/password in plaintext when binding.

src/service/passport/ldap.ts
```diff
@@ -0,0 +1,298 @@
...
+ * entry that references the user's DN.
+ */
+export const isUserInGroup = async (
+  client: Client,
+  ldapConfig: LDAPConfig,
+  userDN: string,
+  groupDN: string,
+  username: string,
+): Promise<boolean> => {
+  const groupFilter = (ldapConfig.groupSearchFilter || '(member={{dn}})')
```
jescalada, line 135: Another AI-driven review comment:

Check if `groupSearchFilter` is wrapped in parentheses, and add them if not so. (Otherwise we should document the requirement so we don't end up with malformed filters)

src/service/passport/ldap.ts
```diff
@@ -0,0 +1,298 @@
...
+      if (!username || !password) {
+        return done(null, false);
+      }
+
+      try {
+        const user = await authenticateUser(ldapConfig, username, password);
+        return done(null, user || false);
+      } catch (error: unknown) {
+        const message = handleErrorAndLog(error, 'LDAP authentication error');
+        return done(message);
```
jescalada, line 279: `passport` has its own way of handling errors, we should check if calling `done(error)` instead of `done(message)` handles this better or more efficiently (worst case being that it silences the error for some reason).

I'm aware that we're already doing this in `src/service/passport/activeDirectory.ts` - however if passing the `error` is the right choice, we should also fix the calls in that file.

jescalada (changes requested):
@1saac-k Thanks for the contribution! 🚀 

Would you mind adding a short demo or some screenshots showing a working LDAP auth flow, and the resulting database entries when creating users? We want to make sure that database entries are not malformed: this might happen if LDAP gives an array of strings for the `user.email` among other fields.

1saac-k:
> @1saac-k Thanks for the contribution! 🚀
> 
> Would you mind adding a short demo or some screenshots showing a working LDAP auth flow, and the resulting database entries when creating users? We want to make sure that database entries are not malformed: this might happen if LDAP gives an array of strings for the `user.email` among other fields.

Thank you for taking the time to review this.

1. I will share configs, screenshots, LDAP-related logs, and DB values showing git-proxy working with [lldap](https://github.com/lldap/lldap) (as one example LDAP server).
2. I will test whether git-proxy handles cases where array values or invalid values exist in LDAP.

I will review and address the items you pointed out as soon as possible!

kriswest:
Theres a few small conflicts to resolve on this

jescalada:
Hi @1saac-k, are you still working on this? It'd be great to resolve conflicts and fix up my open comments to get this merged. Also, test coverage ought to be raised to >90%.

We're happy to help clear these action items if you're unavailable. Thanks!

kriswest:
Conflicts resolved, with 2.1.0 releasing shortly we should be ready to merge this very soon, once @jescalada's and @dcoric comments are dealt with. 

@dcoric Please note that you'll have to re-visit reviews you've commented on now that you've got write access ass they will block merges until resolved ;-)

1saac-k:
@jescalada @kriswest 
Sorry for the late reply. Unfortunately, right after @jescalada kindly reviewed it, things got so busy that I haven't had any time to work on it. But I'll definitely finish it someday — and I hope "someday" turns out to be "soon" (I want that too!).

PR #1501, opened 2026-04-10
feat: add tsoa for API type control and documentation

### LINKED ISSUE #1430: Improve API types and documentation
Right now, our API routes (`/src/service/routes/*`) are undocumented and don't enforce strict request/response types causing confusion on how to use certain endpoints, especially for those that return `Action` objects or allow filtering responses.

This also prevents #1343 from being completed, since generating user-friendly documentation (in the form of a MDX doc) requires some preexisting API spec.

I thought about the best way to do this, and I like [tsoa](https://github.com/lukeautry/tsoa) for strengthening the types and also generating an OpenAPI reference which can be easily converted to MDX via another tool.

An alternative is to manually add the OpenAPI reference, but handling massive YAML docstrings in the endpoint code doesn't seem like such a good idea.

Any thoughts? @finos/git-proxy-maintainers 

**Additional context**
#1343

### LINKED ISSUE #1510: Render generated `swagger.json` for API documentation
#1501, not yet merged as of creating this issue, refactors the endpoints to make them type-safe and documented. It generates a `swagger.json` (OpenAPI spec) that can be easily parsed into MDX and added to our website documentation.

Automatically parse the generated `swagger.json` into MDX that's added to the website docs, similar to how the [Schema Reference](https://git-proxy.finos.org/docs/configuration/reference/) is autogenerated in `scripts/doc-schema.js`:

```js
const { execFileSync } = require('child_process');
const { writeFileSync, readFileSync, mkdtempSync } = require('fs');
const { tmpdir } = require('os');
const { sep } = require('path');
const JSFH_CONFIG = './jsfh.config.json';
const SCHEMA_FILE = './config.schema.json';
const OUTPUT_PATH = './website/docs/configuration/reference.mdx';

try {
  const osTempdir = tmpdir();
  const tempdir = mkdtempSync(`${osTempdir}${sep}`);

  const genDocOutput = execFileSync('generate-schema-doc', [
    '--config-file',
    JSFH_CONFIG,
    SCHEMA_FILE,
    `${tempdir}${sep}schema.md`,
  ]).toString('utf-8');
  console.log(genDocOutput);

  const schemaDoc = readFileSync(`${tempdir}${sep}schema.md`, 'utf-8')
    .replace(/\s\s\n\n<\/summary>/g, '\n</summary>')
    .replace(/# GitProxy configuration file/g, '# Schema Reference'); // https://github.com/finos/git-proxy/pull/327#discussion_r1377343213
  const docString = `---
title: Schema Reference
description: JSON schema reference documentation for GitProxy
---

${schemaDoc}
`;
  writeFileSync(OUTPUT_PATH, docString);
  console.log(`Wrote schema reference to ${OUTPUT_PATH}`);
} catch (err) {
  console.error(err);
}
```

An alternative is to use Swagger UI which is feature rich, allows testing endpoints, etc., although this would add a runtime dependency.

**Additional context**
Original issue: #1343 (Closed since the requirements have changed considerably since we needed to first improve the API code and docstrings)

Andreybest:
Resolves #1430.

Add tsoa for endpoint type checking and generation of OpenAPI documentation. Generates `swagger.json` to /dist folder.
All tests for endpoints that were present, passes (with minor changes), all logic is preserved.

Caveats:
1. tsoa checks for field presence, but does not check if string in field is empty, thus should be checked manually (or I've missed something in configuration). For example: `POST api/v1/push/{id}/reject` (rejectPush), if empty or whitespaces provided, will not check for it, thus old logic for theck with `.trim()` is left.
2. tsoa has a configuration field `noImplicitAdditionalProperties`, that allows to change behavior of endpoints for rejecting of excess fields. Currently set to `ignore` due to `POST api/v1/repo` (createRepo), this endpoint passes body to a `db.createRepo(body)`, and inner type expects additionalProperties that are passed to a row in a DB. Question: is this expected or not? If not, maybe we should enable more strict mode on `noImplicitAdditionalProperties`? :)


package.json
```diff
@@ -49,8 +49,9 @@
     "server": "ALLOWED_ORIGINS=* tsx index.ts",
     "start": "concurrently \"npm run server\" \"npm run client\"",
     "build": "npm run generate-config-types && npm run build-ui && npm run build-ts",
-    "build-ts": "tsc --project tsconfig.publish.json && node scripts/fix-shebang.js",
+    "build-ts": "npm run build-tsoa && tsc --project tsconfig.publish.json && node scripts/fix-shebang.js",
```
jescalada, line 52: Wonder if this can be simplified or done elsewhere? I feel the `build-ts` script is starting to get too cluttered. Perhaps we could rename the `build-tsoa` script from below to something more descriptive like `build-api-spec`, and then call `build-api-spec && build-ts` when needed?

Feel free to suggest alternatives 🙂

src/app.ts
jescalada, line 1: It seems Git mistakenly interpreted this as a rename... Don't think much can be done here since the diff looks valid 👍🏼

src/service/authentication.ts
jescalada, line 1: If I understand correctly, this is a rewrite of `/src/service/passport/jwtAuthHandler.ts`? Is this necessary, and if so, should we delete that version instead?

We should double check that all usages of `jwtAuthHandler` still work as intended.

src/service/controllers/AuthController.ts
```diff
@@ -0,0 +1,342 @@
...
+        console.log('Logged in successfully. User:', user);
+        return res.redirect(`${uiHost}:${uiPort}/dashboard/profile`);
+      });
+    },
+  )(req, res, next);
+}
+
+// ---------- Response types ----------
+
+interface AuthResources {
```
jescalada, line 116: If possible, these types should be stored in a separate file and imported here (and in other controllers) if necessary

src/service/controllers/AuthController.ts
```diff
@@ -0,0 +1,342 @@
...
+      usernamePasswordMethod,
+      // other enabled auth methods
+      otherMethods: getAuthMethods()
+        .map((am) => am.type.toLowerCase())
+        .filter((authType) => authType !== usernamePasswordMethod),
+    };
+  }
+
+  // TODO: provide separate auth endpoints for each auth strategy or chain compatibile auth strategies
+  // TODO: if providing separate auth methods, inform the frontend so it has relevant UI elements and appropriate client-side behavior
```
jescalada, line 191: I'm aware these comments were here before, but I believe the `/auth/config` endpoint is being used in the UI right now to fetch the auth methods. I think we can safely delete the "seperate auth methods" one.

src/service/controllers/AuthController.ts
```diff
@@ -0,0 +1,342 @@
...
+
+  /**
+   * Updates the Git account (username) of a user.
+   * Admins may update any user; non-admins may only update their own account.
+   */
+  @Post('/gitAccount')
+  public async updateGitAccount(
+    @Body() body: GitAccountBody,
+    @Request() req: ExpressRequest,
+    @Res() unauthorisedResponse: TsoaResponse<401, { message: string }>,
```
jescalada, line 275: Wonder if we could look at standardizing/improving the shape of the API responses too, since not all API endpoints are using this `message: string` version

src/service/controllers/AuthController.ts
```diff
@@ -0,0 +1,342 @@
...
+    @Res() notFoundResponse: TsoaResponse<404, { message: string }>,
+    @Res() validationErrorResponse: TsoaResponse<400, { message: string }>,
+    @Res() forbiddenResponse: TsoaResponse<403, { message: string }>,
+    @Res() internalServerErrorResponse: TsoaResponse<500, { message: string }>,
+  ): Promise<void> {
```
jescalada, line 280: Shouldn't the success response's shape also be typed? Same for all the other endpoint functions

jescalada: Looked into this into more detail, and it might be alright. Just wondering if the return value should be updated from `void` to something else to indicate success, or perhaps the absence of errors is enough?

src/service/controllers/PushController.ts
```diff
@@ -0,0 +1,285 @@
...
+import { AttestationAnswer, Rejection } from '../../proxy/processors/types';
+import { MessageResponse } from '../models';
+
+interface RejectBody {
+  /** The reason for rejecting the push request. */
+  reason: string;
+}
+
+/** Attestation answer as sent by the UI (may include display-only fields). */
+interface AttestationAnswerRequest {
```
jescalada, line 45: `AttestationAnswer` already exists at `src/proxy/processors/types.ts` (ignore if intentional)

src/service/controllers/PushController.ts
```diff
@@ -0,0 +1,285 @@
...
+    @Path() id: string,
+    @Body() body: RejectBody,
+    @Request() req: ExpressRequest,
+    @Res() unauthorisedResponse: TsoaResponse<401, { message: string }>,
+    @Res() validationErrorResponse: TsoaResponse<400, { message: string }>,
+    @Res() notFoundResponse: TsoaResponse<404, { message: string }>,
+    @Res() forbiddenResponse: TsoaResponse<403, { message: string }>,
+  ): Promise<MessageResponse> {
+    if (!req.user) {
+      return unauthorisedResponse(401, { message: 'Not logged in' });
```
jescalada, line 114: Why do we pass in the error code when calling the response functions if they already have it set in the definition above? `unauthorizedResponse: TsoaResponse<401, { message: string }>`

src/service/controllers/RepoController.ts
```diff
@@ -0,0 +1,299 @@
...
+  name: string;
+  project: string;
+}
+
+interface RepoWithProxy extends Repo {
+  proxyURL: string;
+}
+
+/**
+ * Repository management (JWT-protected).
```
jescalada, line 56: Wondering if we could remove the "JWT-protected" comments throughout the code since JWT API auth is optional 🤔

src/service/controllers/RepoController.ts
```diff
@@ -0,0 +1,299 @@
...
+  /**
+   * Grants a user push permission on a repository.
+   */
+  @Patch('/{id}/user/push')
+  public async addPushUser(
+    @Path() id: string,
+    @Body() body: UsernameBody,
+    @Request() req: ExpressRequest,
+    @Res() unauthorisedResponse: TsoaResponse<401, { message: string }>,
+    @Res() userNotFoundResponse: TsoaResponse<400, { error: string }>,
```
jescalada, line 174: Would be great to standardize these to use `{ message: string }` on both, as well as all the endpoints below. This may require tweaking the usages in the UI

src/service/controllers/RepoController.ts
```diff
@@ -0,0 +1,299 @@
...
+        `API request to proxy repository ${repoUrl} is for a new origin: ${newOrigin},\n\texisting origin list was: ${JSON.stringify(existingHosts)}`,
+      );
+
+      const repoDetails = await db.createRepo(body);
+      const proxyURL = getProxyURL(req);
+
+      // restart the proxy if we're proxying a new domain
+      if (newOrigin) {
+        console.log('Restarting the proxy to handle an additional host');
+        const proxy = getProxy();
```
jescalada, line 153: Pinging @kriswest in case this is relevant - not sure if using a store affects the proxy restart process. (Related: [`src/controllers/proxyStore.ts`](https://github.com/finos/git-proxy/pull/1501/files#diff-8dac2901f86a3ede94a7e77bd4af037211159e71c7536c271162e093bf7fd751))

src/service/controllers/RepoController.ts
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    // determine if we need to restart the proxy
+    const previousHosts = await db.getAllProxiedHosts();
+    await db.deleteRepo(id);
+    const currentHosts = await db.getAllProxiedHosts();
+
+    if (currentHosts.length < previousHosts.length) {
+      console.log('Restarting the proxy to remove a host');
+      const proxy = getProxy();
```
jescalada, line 292: Another bit to double-check for proper restart behaviour

src/service/models.ts
jescalada, line 1: Maybe the request/response interfaces could be stored here?

test/services/routes/auth.test.ts
```diff
@@ -16,22 +16,53 @@
...
 import * as db from '../../../src/db';
+import { ValidateError } from 'tsoa';
 
-const newApp = (username?: string): Express => {
+/**
+ * Builds a minimal Express app with tsoa routes registered.
+ * When `username` is supplied the request is pre-authenticated via middleware,
+ * simulating a session-authenticated user.
+ */
+const newApp = (username?: string, isAdmin = false): Express => {
```
jescalada, line 29: Same as the comment on `users.test.ts`, I don't think setting `isAuthenticated` is necessary as it's exclusively used for the `jwtAuthHandler` which is disabled by default.

test/services/routes/users.test.ts
```diff
@@ -26,7 +26,17 @@ describe('Users API', () => {
   beforeEach(() => {
     app = express();
     app.use(express.json());
-    app.use('/users', usersRouter);
+    // Pre-authenticate so JWT security allows through in tests.
```
jescalada, line 29: This comment is a bit misleading - JWT auth isn't required by default. The `isAuthenticated` flag is used only for JWT anyways and shouldn't be necessary here AFAIK 🤔

tsoa.json
```diff
@@ -0,0 +1,26 @@
+{
+  "entryFile": "src/app.ts",
+  "noImplicitAdditionalProperties": "ignore",
+  "controllerPathGlobs": ["src/service/controllers/**/*Controller.ts"],
+  "spec": {
+    "outputDirectory": "dist",
+    "specVersion": 3,
+    "securityDefinitions": {
```
jescalada, line 8: Might be worth mentioning that JWT API auth is optional - and might be subject to change as it's a feature only used by our org. It's a bit confusing and poorly documented so we'll be polishing things once v2 is out...

jescalada (changes requested):
Thanks for the contribution, @Andreybest! 🚀

Things are looking good so far, just a few things to mention other than the code comments:

- It'd be great to have screenshots of the generated OpenAPI spec to see if things are working as we hoped
- I noticed a duplicate interface (`AttestationAnswer`) and am wondering if there aren't other interfaces that can be reused or combined into one
- Test coverage seems to go down somewhat (-1.5%), we should improve coverage on any code additions such as `proxyStore.ts`, etc.

We might need another review since the changes are quite large @finos/git-proxy-maintainers

kriswest:
@jescalada and @Andreybest There is a docusaurus plugin for rendering OpenAPI docs at:

https://docusaurus-openapi.tryingpan.dev/
https://github.com/PaloAltoNetworks/docusaurus-openapi-docs

Whereas in the FDC3 website we use a very simple static page and the Redoc lib to render the same: https://github.com/finos/FDC3/blob/main/website/src/pages/schemas/next/app-directory.html

Note that you have to link to this as an external link or docusaurus will return a 404 when navigating within the site:

https://github.com/finos/FDC3/blob/ffea3be0735bdd7a72ffba5473addfe49535833e/website/docs/app-directory/spec.md?plain=1#L7-L13

package.json
Andreybest: `build-tsoa` builds `swagger.json` **and** `generatedRoutes.ts` which is required for `index.ts`, without it - build fails, so its is essential part of a build (unfortunately). Considering this thing, `build-tsoa` naming makes more sense than `build-api-spec` because we build spec and routes (tsoa spec-and-**routes**).

src/service/authentication.ts
Andreybest: Unfortunatelly - yes. TSOA expect an expressAuthentication with a different set of rules. Deleted `/src/service/passport/jwtAuthHandler.ts`. Yes, it is the rewrite. 
Checked and fixed tests for jwtAuth

src/service/controllers/AuthController.ts
Andreybest: Deleted!

src/service/controllers/AuthController.ts
Andreybest: Moved to `/interfaces/{controller_name}.interfaces.ts`, please let me know if that is alright :)

src/service/controllers/AuthController.ts
Andreybest: Done, moved all to `/decorators/response.types.ts`. Only thing that differs from other errors: `UserNotFoundResponse = TsoaResponse<400, { error: string }>;`

It uses `error: string` instead of `message: string`

src/service/controllers/AuthController.ts
Andreybest: Yes, in return type you only show the returned type and it will be parsed to OpenAPI. That is alright

src/service/controllers/PushController.ts
Andreybest: Removed, thanks for noticing!

src/service/controllers/PushController.ts
Andreybest: In case of parameter decorators in TSOA, it does not take type and use it in runtime. So we can't remove this repetition, only way to change that, is to fork TSOA...

src/service/controllers/RepoController.ts
Andreybest: Removed :)

src/service/controllers/RepoController.ts
Andreybest: Yes, and I dont really want to change this in this PR to not break even a small thing. Considering the future huge UI update... But if we really can make this, let me know :)

src/service/interfaces/common.interfaces.ts
Andreybest: Moved to common.interfaces.ts. To follow naming with other responses/requests

test/services/routes/auth.test.ts
Andreybest: Removed

test/services/routes/users.test.ts
Andreybest: Removed

tsoa.json
Andreybest: In this case, it is only used for 2 things:

1. For a name in the decorator to be used as `Security('jwt')`
2. As a documentation for OpenAPI

It does not do any functionally, but it is required anyway if we want to use `Security` decorator properly

Andreybest:
Thank you for the review @jescalada !
Can you please review it again? Applied changes to most of your comments :)

1. Screenshot of swagger 
image
2. Duplicate removed, quickly checked, should be no more repetiotions (but not sure)
3. Don't see a message from code cov bot on this matter, can you please show me where I can find a code coverage stats?

Andreybest:
Checked all commits on main branch for new changes to logic for endpoints. None was changed, checked up to commit `6981427c50a5e5ccd3e91a2d4229b27135f8dcc3`

package.json
```diff
@@ -54,8 +54,9 @@
...
     "build-ui": "vite build",
+    "build-tsoa": "tsoa spec-and-routes",
     "check-types": "tsc",
     "check-types:server": "tsc --project tsconfig.publish.json --noEmit",
```
fabiovincenzi: CI jobs fail because the `check-types:server` step runs `tsc --noEmit` before `build-ts`, on a fresh checkout generatedRoutes.ts doesn't exist yet when tsc runs.
```suggestion
    "check-types:server": "npm run build-tsoa && tsc --project tsconfig.publish.json --noEmit",
```

Andreybest:
Thanks for change @fabiovincenzi ! Added your suggested change.

Andreybest:
Checked all commits on main branch for new changes to logic for endpoints. None was changed, checked up to commit: `ad58c5a47f64b6380da1ce1d7f6556c17e7e9efc`

tsoa.json
```diff
@@ -0,0 +1,26 @@
+{
+  "entryFile": "src/app.ts",
+  "noImplicitAdditionalProperties": "ignore",
```
re-vlad: This allows arbitrary additional properties in request bodies. The current implementation in `createRepo()` passes the entire body to `db.createRepo(body)`, but this should be validated against a strict schema. The db schema should be the source of truth, not the API layer accepting arbitrary data. So, change "noImplicitAdditionalProperties": "ignore" to "noImplicitAdditionalProperties": "throw"

re-vlad (commented):
just small non critical change requested

src/service/index.ts
```diff
@@ -26,8 +26,11 @@ import * as config from '../config';
 import * as db from '../db';
 import { serverConfig } from '../config/env';
 import { Proxy } from '../proxy';
-import routes from './routes';
+import { RegisterRoutes } from './generatedRoutes';
```
kriswest, line 30: is this file checked in @Andreybest ?

src/service/controllers/RepoController.ts
kriswest: As long as it triggers the registration of routes (which looks like its depending on a missing file currently).

Probably missing a call to RegisterRoutes(app); - which is from the missing source file

src/service/controllers/RepoController.ts
kriswest: again probably missing a call to `RegisterRoutes(app);`

src/service/index.ts
Andreybest: This one is generated during build

Andreybest: I do not commit it, you suggest it to be included in repo?

kriswest: A perennial question! I've tended to go with committing it as reviewing the generated code can often help you spot issues in what its generated from, and it helps someone trying to navigate through the code to understand something. Tooling such as SAST scanners an code coverage calculators may need it to be there to work as they should.

If committing generated code, then its worth either adding a commit hook (best to check for modifications in this) or PR check box to confirm its been done ;-)

Andreybest:
Checked all commits on main branch for new changes to logic for endpoints. Only change with addition of https server, which required to change test files for https, would appreciate check on this.

Checked up to commit: `4c6e8d47d9189ee016d6f6a59adbd2f013c0666f`

src/service/index.ts
Andreybest: Thanks for the concern!

Added `generatedRoutes.ts` to branch, and added pre-commit hook to lint-staged that will automatically stage changes to `generatedRoutes.ts`.

tsoa.json
Andreybest: Hey @re-vlad, thanks for the concern. That is actually what I pointed out in my PR description no. 2 point. And I agree with you.

@kriswest and @jescalada would appreciate your comment on this matter.

src/service/controllers/RepoController.ts
jescalada: Makes sense, I'm just nitpicking - although if we ever do large-scale type improvements again we might want to fix those up 👍🏼

src/service/controllers/RepoController.ts
jescalada: Pinging @Andreybest in case you missed this comment 🙂

src/service/controllers/RepoController.ts
jescalada: @Andreybest

tsoa.json
jescalada: I agree with setting `noImplicitAdditionalProperties` to `throw` 👍🏼 

Can this be done by adding proper types to the request body? You can see what the body looks like in `src/ui/services/repo.ts`:

```ts
const addRepo = async (repo: RepoView): Promise<ServiceResult<RepoView>> => {
  const url = new URL(`${apiV1Base}/repo`);

  try {
    const response = await axios.post<RepoView>(url.toString(), repo, getAxiosConfig());
    return successResult(response.data);
  } catch (error: unknown) {
    return errorResult(error, 'Failed to add repository');
  }
};
```

Note that `repo` is of `RepoView` type, which is just an extended version of the database `Repo`:

```ts
export interface RepoView extends Repo {
  proxyURL: string;
  lastModified?: string;
  dateCreated?: string;
}
```

jescalada (commented):
@Andreybest Looks good so far, just wondering if you've dealt with Kris' comments above for the missing `RegisterRoutes(app)`

As for the `"noImplicitAdditionalProperties": "ignore"` setting, we might want to fix our typing so that it's possible to set that flag to `"throw"`! 👍🏼

jescalada:
Also, don't forget to run `npm run format` at the end to fix the CI 🙂

src/service/controllers/RepoController.ts
Andreybest: Thanks for pointing out @jescalada, somehow missed it :)
`generatedRoutes.ts` now in repo, so we should not be missing the registration of routes.

src/service/controllers/RepoController.ts
Andreybest: Answered in comment above

tsoa.json
Andreybest: Changed `"noImplicitAdditionalProperties"` to  `"throw-on-extras"`, and fixed tests to accommodate this change.

jescalada (approved):
@Andreybest LGTM, just needs fixes for the failing tests!

It seems this commit (https://github.com/finos/git-proxy/pull/1501/commits/a23b7ebfd02f716c5c29fb964aa4c179dedb6509) breaks them - I'm wondering if it's due to the `throw-on-error` option we added, or the tweaks to the repo tests themselves 🤔

Andreybest:
Thank you @jescalada! Sorry for not replying on previous review.
On tests - fixed, reason - no idea, but the change is minimal - users for creation of new repo is now optional to be aligned with tests. In the code for interaction with db it is replaced with empty arrays anyway, so should work as expected :)

Andreybest:
Checked all commits on main branch for new changes to logic for endpoints. None was changed, checked up to commit: `ffd6937419b1d3de718c60ec26604f9f7541fffa`

jescalada:
Just pushed fixes for the E2E (Docker) and Cypress tests. The `throw` flag is indeed causing the test failures due to how strict it is with unused/extra params.

Meanwhile the Dependency error is coming from critical vulnerabilities in our `devDependencies` (`vitest`, `nyc`). Upgrading breaks our tests, so I'll find fixes for these and report back 👍🏼

jescalada:
@kriswest I think this can be reviewed again since Dependency Review isn't a required check. I'll do the Vitest upgrade in a separate PR!

Andreybest:
Hey @jescalada! Thanks for the fixes! Merged latest main to PR and added suggestion from @fabiovincenzi on pre-commit generation of web docs (just like generatedRoutes being automatically generated on commit).
Please let me know if any changes are needed.

Andreybest:
Checked all commits on main branch for new changes to logic for endpoints. On main new feature for getting params from config for server configuration was added to `auth.ts` route of old express api, moved logic to Auth controller, checked up to commit: `6e657167273404f4c2d668f9fd82f28045453058`

package.json
```diff
@@ -59,10 +59,11 @@
     "server": "cross-env ALLOWED_ORIGINS=* tsx index.ts",
     "start": "concurrently \"npm run server\" \"npm run client\"",
     "build": "npm run generate-config-types && npm run build-ui && npm run build-ts",
-    "build-ts": "tsc --project tsconfig.publish.json && node scripts/fix-shebang.js",
+    "build-ts": "npm run build-tsoa && tsc --project tsconfig.publish.json && node scripts/fix-shebang.js",
```
kriswest, line 62: you possibly need to add a run of the format script here as the generated code fails linting, but can be autofixed - which would stop this PR (and future ones) failing the linting check.

Alternatively see how generate-config-types runs prettier on a specific file

kriswest:
@Andreybest can you resolve conflicts and deal with the application of prettier to the generated files (to stop the linting check failing - see https://github.com/finos/git-proxy/pull/1501#discussion_r3387950683). After that this should be mergable!

package.json
Andreybest: I don't think that we need to format generated files, I've actually added generatedRoutes.ts to .prettierIgnore, but it was not used, now it is fixed.

Andreybest:
Checked all commits on main branch for new changes to logic for endpoints. On main new ssh routes were added, and code for routes and tests were aligned with TSOA, checked up to commit: `656d3f5395c855834490e30b0db8b78411c27465`

Andreybest:
@kriswest, conflicts resolved, new features alligned with this PR. On your comment - made a solution, please approve it, and if this works - lets go!

package.json
kriswest: I'm not bothered either way tbh, as long as it passes the check. Other projects I work on apply automated fixes to generated files, and it occasionally squeezes out variations between different versions of the code generation dependency - but then can also introduce differences in PRs if the formatter wasn't run for some reason.

scripts/doc-swagger.js
jescalada, line 1: @fabiovincenzi Just wondering if this can be imported as a library and configured, rather than adding the entire script locally - just to save us the trouble of maintaining it (for example, modifying our API by introducing a response with a new status code, would also require registering the status code in this file).

jescalada, line 1: If I understand correctly, this is being executed automatically when calling `lint-staged` in our pre-commit hook?

jescalada (commented):
@fabiovincenzi OpenAPI spec is looking good to me 👍🏼

image

---

It'd be nice to document each field description in the code, and *when* each response happens, so developers don't need to check the source code themselves. We ought to add this in a separate PR to finish up the API documentation! 😃

scripts/doc-swagger.js
Andreybest: Yes it is @jescalada :)

website/docs/architecture/architecture.md
```diff
@@ -452,7 +452,7 @@ Allows defining ways to authenticate to the API. This is useful for securing cus
 
 If `apiAuthentication` is left empty, API endpoints will be publicly accesible.
 
-Currently, only JWT auth is supported. This is implemented via the [`jwtAuthHandler` middleware](https://github.com/finos/git-proxy/blob/main/src/service/passport/jwtAuthHandler.ts). Aside of validating incoming access tokens, it can also assign roles based on the token payload.
+Currently, only JWT auth is supported. This is implemented via the [`@Security('jwt')` decorator](https://github.com/finos/git-proxy/blob/main/src/service/authentication.ts). Aside of validating incoming access tokens, it can also assign roles based on the token payload.
```
kriswest, line 455: typo (Aside _of_ ...), Can improve wording slightly:

```suggestion
Currently, only JWT auth is supported. This is implemented via the [`@Security('jwt')` decorator](https://github.com/finos/git-proxy/blob/main/src/service/authentication.ts). In addition to validating incoming access tokens, the decorator assigns roles based on the token payload.
```

scripts/doc-swagger.js
```diff
@@ -0,0 +1,362 @@
...
+
+  lines.push('---');
+  lines.push('title: API Reference');
+  lines.push('description: REST API reference documentation for GitProxy');
+  lines.push('---');
+  lines.push('');
+  lines.push('# API Reference\n');
+
+  if (info.description) {
+    lines.push(`${info.description}\n`);
```
kriswest, line 280: This ends up being the description from the package.json file, its a very limited intro on the docs page:

Image

Perhaps we could swap this out for a nicer introduction to the docs page.

kriswest (commented):
One really minor comment on the docs page (which is a great addition btw) which could have a better intro - still comparing the revised controllers to the previous routes, and some testing to do before finalising review. Hope to get somewhere on that tomorrow

package.json
```diff
@@ -148,6 +150,7 @@
     "react-router-dom": "6.30.3",
     "simple-git": "^3.30.0",
     "ssh2": "~1.17.0",
+    "tsoa": "^7.0.0-alpha.0",
```
kriswest, line 153: Question: How comfortable are we on basing this on an alpha release of an intermittently released project thats 7 months old? Theres been no activity over the last 6 months... Questions about the release timeline for vv7 have gone unanswered: https://github.com/lukeautry/tsoa/issues/1859

The OSFF scorecard result for this project is low at 3.9 - at the least we should probably raise an issue or PR in the project with some suggestions to improve that score (e.g. enabling branch protection, adding a SAST workflow, fixing its GitHub workflow tokens with excessive permissions). The project is in desperate need of contributions/maintainers.

There exists a tsoa-next project (https://tsoa-next.dev/ https://github.com/tsoa-next/tsoa-next https://www.npmjs.com/package/tsoa-next) which has some slightly more recent activity - but no scorecard result available...

Other options include:

1. NestJS + @nestjs/swagger
  - probably the closest mainstream alternative
  - large active ecosystem and widely adopted
  - but is a much larger framework and would mean migration off Express (prob not viable)
2. Zod + zod-openapi or @asteasolutions/zod-to-openapi
  - Excellent typescript support
  - Not controller based
3. ts-rest
  - Shared client/server contracts (will be good on teh UI end)
  - will require some conceptual migration
4. Flip things around and use the OpenAPI spec generated through this work as the contract that the code has to implement. 
  - Generate interfaces and server contracts/stubs from it that the new controllers implement/use.
  - Have openapi as the long-term dependency rather than tsoa
  - generate the docs from the OpenAPI spec

kriswest (changes requested):
I've hit a problem with the tsoa project which has me concerned, we need to consider if we want to take it on as a long-term dependency. This work is undoubtedly valuable - but we might need to go further and perhaps promote teh OpenAPI spec (gnerated through this work) to being the source of truth that we then implement, rather than the other way round.

WDYT @finos/git-proxy-maintainers

package.json
kriswest: If we get everything else reviewed and merged before this we might consider descoping from the 2.1.0 release while we figure this out.

jescalada: TSOA unfortunately seemed like a good choice when I made the issue, having recent activity at the time :/ 

I like NestJS as a long-term solution - it provides type safety and a lot of idiomatic constructs like Pipes and Guards, and would definitely help clean up our db and auth code as well. But, it's going to result in *another* massive refactoring PR.

As for the other ones, I feel `ts-rest` fits the bill: it lets you [generate OpenAPI specs](https://ts-rest.com/openapi) and extending contracts is very straightforward. `tsoa-next` would let us keep our changes here - but it doesn't seem actively maintained either (it's more like a one-off fork of the original).

Since this would provide contracts for the UI, `ts-rest` could be implemented after merging the UI refactors. Not sure if others are okay with this or if we should go for direct OpenAPI generation instead.

Thoughts? @kriswest @Andreybest @fabiovincenzi @andypols

jescalada:
Ahead of our community call today, I'm okay with descoping this from 2.1 until we sort out the library swap. I think the best choice is to merge in @andypols' UI refactors first, and then make another PR strengthening types via `ts-rest`. Feel free to suggest alternatives!

package.json
Andreybest: Totally get the point of TSOA "alpha" stage. In my understanding, TSOA would be the closest to express jump, but considering this circumstances, this will not pass.

On my options:

NestJS - was good, but not sure if it is still the choice for the long term, considering that TS dropped support for experimental decorators couple of years ago

ts-rest - looks ok on a quick glance

generate openapi ourselves - i think that is a very bad choice, no type checks on server, small mistakes on dev side which will pour over to clients, and "dev-experience" is not good...

encore - very simple type safe (TS interface based) framework, but has couple of pitfalls

AdonisJs - good modern framework, with simple mechanisms for type safety and client lib generation. And has full database integration (which will require refactoring the DB mechanism too, but if im not wrong, its optional)

kriswest:
> generate openapi ourselves - i think that is a very bad choice, no type checks on server, small mistakes on dev side which will pour over to clients, and "dev-experience" is not good...

Not what I meant. Rather we adopt the openAPI spec that you've ended up creating through this work, but make it the new source of truth and generate types from it for type checking. I.e. you do exactly what OpenAPI was designed for, define the API contract, then implement it and validate against it. I was just referring to reuse of the one you've created (through tsoa and the annotations it uses) so far as a starting point - you would then need to check-in the actual spec document and add code generation for the types and stubs.

Just clarifying what I meant rather than advocating for it. ts-rest is probably the simplest (acceptable) option to translate to - although I doubt a coding assistant would have much trouble switching things around to use an OpenAPI spec and code generation. That way you could also (probably generate) the UI services that work with the API 🤔

Issue #1506, opened 2026-04-20
20th April 2026 - GitProxy Meeting Minutes


jescalada:
## Date

20260420 - 11am EDT / 4pm BST

## Meeting info

- [Meeting link](https://zoom-lfx.platform.linuxfoundation.org/meeting/95849833904?password=99413314-d03a-4b1c-b682-1ede2c399595)

- [Register for future meetings](https://zoom-lfx.platform.linuxfoundation.org/meeting/95849833904?password=99413314-d03a-4b1c-b682-1ede2c399595&invite=true)

## Meeting notices

- FINOS **Project leads** are responsible for observing the FINOS guidelines for [running project meetings](https://community.finos.org/docs/governance/meeting-procedures/). Project maintainers can find additional resources in the [FINOS Maintainers Cheatsheet](https://community.finos.org/docs/finos-maintainers-cheatsheet).

- **All participants** in FINOS project meetings are subject to the [LF Antitrust Policy](https://www.linuxfoundation.org/antitrust-policy/), the [FINOS Community Code of Conduct](https://community.finos.org/docs/governance/code-of-conduct) and all other [FINOS policies](https://community.finos.org/docs/governance/#policies).

- FINOS meetings involve participation by industry competitors, and it is the intention of FINOS and the Linux Foundation to conduct all of its activities in accordance with applicable antitrust and competition laws. It is therefore extremely important that attendees adhere to meeting agendas, and be aware of, and not participate in, any activities that are prohibited under applicable US state, federal or foreign antitrust and competition laws. Please contact legal@finos.org with any questions.

- FINOS project meetings may be recorded for use solely by the FINOS team for administration purposes. In very limited instances, and with explicit approval, recordings may be made more widely available.

## Agenda

- [x] Convene & roll call (5mins)
- [x] Display [FINOS Antitrust Policy summary slide](https://community.finos.org/Compliance-Slides/Antitrust-Compliance-Slide.pdf) and review Meeting Notices (see above)
- [x] Approve past meeting minutes
  - #1471 
- [ ] Release branching strategy + release-drafter fix
  - #1483
  - is #557
 needed?
- [x] V2.0.0-rc.6
- [x] Revisit round-table discussion on objectives for Git-Proxy
- [ ] Issue prioritisation for roadmap
  - Renovate PRs (10 of 27 open PRs): https://github.com/finos/git-proxy/pulls/app%2Frenovate
  - Issues with a PR (13 of 27 open PRs) :
    - [x] #27 
[...]

Issue #1363, opened 2026-01-26
Ensure Git Proxy user interfaces meet WCAG 2.2 Level AA accessibility requirements


kriswest:
**Is your feature request related to a problem? Please describe.**

UK and EU private sector firms are encouraged to meet WCAG 2.2 Level AA for accessibility compliance. This standard ensures a comprehensive level of accessibility for a wider range of users with disabilities. To my knowledge, Git Proxy has not been reviewed for accessibility yet and may not meet the desired level of accessibility.

**Describe the solution you'd like**

- Conduct a review of Git Proxy’s accessibility using the identified support/tooling.
- Build an issue list for remediation based on the review findings.
- Implement the relevant accessibility features.

**Additional context**

We intend to contribute to this work in March 2026.

PR #1507, opened 2026-04-20
feat: Refreshed UI

### LINKED ISSUE #1106: Replace MUI v4 styling with Tailwind
I looked into upgrading the existing MUI v4 styles to v5, which requires changing them into the newer `styled` API. There are currently 33 style files, which contain lots of styles not all of which are being used. (#1023) On top of that, the TS refactors make it harder to refactor the styles without getting type errors.

For the same amount of effort, we could remove our custom MUI styles entirely and use Tailwind instead, which will make our codebase cleaner and more manageable (at the cost of slightly more cluttered component code). 

Removing the MUI dependency (we could keep the icons), and recreate the styles with Tailwind. Since the app is not at all UI-heavy, it shouldn't take a lot of work other than getting the spacing and fonts right.

Upgrading MUI to v5 has been a pain, and it would not get rid of the extensive, and potentially unused styles in the app. Removing the unused styles would require going through each entry in each style file which is too much work. 

**Additional context**
Related: #1023

### LINKED ISSUE #1123: Supress tooltip errors for private projects
**Describe the bug**
If you add a private GitHub project to git proxy, users without access to it or not logged into GitHub will be presented with tooltip errors whenever they visit the repository list or details page for that project.

As the additional data returned by the API call is non-essential (project metadata and images), we should suppress the error / reduce it down to a warning message in the console. The data will definitely be optional after #1043 as we can only retrieve it for GiutHub and GitLab and not other SCM providers/bare git repos (without them supporting it or doing further work). 

**To Reproduce**
Steps to reproduce the behaviour:
1. add a private project, e.g. https://github.com/finos/flowave-bpm-platform.git (which does exist, but is currently private)
2. Confirm you can't see the project (visit the URL)
3. Visit the repository listing page anbd observe the tooltip error.

**Expected behaviour**
The existence of a private project doesn't degrade the UI experience for users that don't have access to it.

**Screenshots**

Image

### LINKED ISSUE #1176: Resolve users to GitHub and Gitlab user profiles in push views
Follow ip to #1175 
GitProxy used to assume that git clients are configured such that the `user.name` config is set to the GitHub username and would use that to render liks to the GitHub user profile. However, this is not a correct assumption as the user.name is usually set to a human readable name (see: https://docs.github.com/en/get-started/git-basics/setting-your-username-in-git#setting-your-git-username-for-every-repository-on-your-computer).

#1175 seeks to remove the estimated profile links, ot replace them with email links, as the email should always be set on a commit (meaning it will be present even for a bare git server). However, if we can identify a supported git SCM provider the users details should be resolved to profile url and link rendered to it.
 
GitProxy should:
- Be able to render links to user profiles (in push detail views and user profiles) for multiple platforms (at least GitHub and GitLab)
- Be able to _resolve_ users to user profile links rather than guessing at them, either:
  - via API calls
  - via its own user database 

Looking around, its not easy to get the GitHub id from the email address - at best you can search against the single, optional public/primary email address with the API. E.g. This works: https://api.github.com/search/users?q=solankiarpit1997@gmail.com but this doesn't https://api.github.com/search/users?q=kristopher.west@natwest.com

This lib offers an alternative based on searching for commits with the email address: https://github.com/sindresorhus/github-username
I've not tried it, but it might provide a solution

For Gitlab there is a user API for lookup, but it will require an API access token to return results:
https://docs.gitlab.com/api/users/#as-a-regular-user

The final alternative is to support multiple fields for user names at SCM providers and to use those to resolve emails to usernames/profile links. However, we only currently require commiters (those pushing the changes) to have accounts in GitProxy and the permission to push, so we can expect some (otherwise legal) authors to not resolve by this method.

### LINKED ISSUE #1258: Remove/rename all usages of `gitAccount`
As per #973, we no longer associate commits by `gitAccount` (which is actually the GH username). #1042 further extends support for other Git hosts, which means that storing GH usernames specifically doesn't make sense either.

We'd like to remove or rename all usages of `gitAccount` in a way that makes sense considering that repos from different hosts (GitHub, GitLab, etc.) can be present. We should also take care of not breaking anything that used to rely on the gitAccount - though most of the extant usages and logic are in the UI. 

Finally, we want to adjust the UI (especially `src/ui/views/User`) so that it suits the current implementation.

**Additional context**
Related PRs #973, #1042, #1257

### LINKED ISSUE #1297: Improve pagination defaults and layout in admin UI

### LINKED ISSUE #1420: Sort entries in the admin UI and apply proper pagination

### LINKED ISSUE #1495: Move GitHub project metadata requests from client to backend to avoid unauthenticated rate limits
# Summary

The GitProxy UI currently calls the GitHub API directly from the client to retrieve project information. These requests are unauthenticated and frequently hit GitHub’s primary rate limit.

## Problem

Because these requests come from the client, they use a shared IP address across the organisation. When GitHub applies rate limiting, it affects everyone using the UI.

According to the GitHub docs, the primary rate limit for unauthenticated requests is **60 requests per hour**.
https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2026-03-10

## Proposed change

Move this GitHub API call from the client to the backend so the request can use an authenticated token.

Authenticated requests have a much higher rate limit, which should significantly reduce failures. Per the GitHub docs, this can scale up to 12,500 requests per hour.

andypols:
# Summary

This PR modernises the git-proxy UI while keeping backend changes to a minimum.

resolves #1106
resolves #1123
resolves #1176
resolves #1258
resolves #1297
resolves #1420
resolves #1495 

## UI L&F 

* Upgraded to the latest React version
* Replaced legacy @material-ui with GitHub’s component library ([@primer/react](https://primer.style/product/getting-started/react/)) and Tailwind CSS (fixes #1106).
* Replaced legacy moment with luxon for UI date/time handling.
* Removed the left-hand navigation and replaced it with a responsive top navigation bar.
* Updated pages to use query parameters to preserve state, so browser back/forward navigation works as expected.

### Standard top navigation:

CleanShot 2026-04-20 at 10 47 40@2x

### Mobile (responsive) navigation:

CleanShot 2026-04-20 at 10 52 03@2x

## Repositories

* Fixed repository sorting, filtering, and pagination (#1420)
* Increased the default page size (#1297)
* Only show pagination when there is more than one page
* Display the total number of repositories
* Added an activity summary for each repository, including approved, cancelled, rejected, and errored pushes, with links to filtered activity views
* Added sorting by name, activity, pending, and relevance (where relevance means you are a contributor or reviewer)
* Default sort is relevance, as this is the most useful view for most users
* Added 24-hour metadata caching for GitHub rate-limited API calls (#1495).  More below.
* Updated the new repository modal to auto-populate the organisation and repository name fields when pasting a URL, if those fields are blank

### Repository list view

CleanShot 2026-04-20 at 11 12 58@2x

## Repository

* Added activity summary for the repository
* Replaced the add-user modal with proper autocomplete.  User lookup now matches on typed characters instead of showing all users. It also supports pasting a username directly

### Repository view

CleanShot 2026-04-20 at 11 51 35@2x

## Activities (previously Dashboard)

* Renamed Dashboard to Activities, as “dashboard” was confusing for users.
* “Activities” also gives us room to expand in future to support PRs, issues, comments, and more.
* Fixed activity (push request) sorting, filtering, and pagination (#1420).
* Increased the default page size (#1297)
* Simplified the push request table so it only contains internal git-proxy links (#1176).
* Long commit messages are truncated in the table, with the full text shown on hover.
* Tabs now show totals.

### Activity list view

CleanShot 2026-04-20 at 11 32 52@2x

## Users

* Fixed user sorting, filtering, and pagination (#1420).
* Increased the default page size (#1297)
* Removed gitAccount from the UI (#1258), although it remains in the database.
* Display the total number of users.
* Added an activity summary for each user.

### User list view

CleanShot 2026-04-20 at 11 40 41@2x

## User 

* Added activity summary for the user

### User view
CleanShot 2026-04-20 at 11 43 41@2x

## Backend APIs & Services 

This branch adds the backend APIs and service-layer support needed for the new UI:

* A new GET `/api/v1/repo/:id/scm-metadata` endpoint fetches and caches GitHub/GitLab repository metadata using new provider-specific logic in `src/service/gitProviders`. The github api endpoint is currently unauthenticated, but the implementation is structured so future PRs can add bearer-token auth and provider-specific operations such as forking or syncing.
* A new GET `/api/v1/user/:id/activity` endpoint returns all pushes attributed to a user.
* The existing GET `/api/v1/user response` now includes activity tab counts.
* The existing GET `/api/v1/repo` response now includes activity tab counts.


kriswest:
Hey @andypols, we discussed this PR at #1506 and are (as a group) excited to see this get done. We debated whether to put this PR in scope for [2.1.0](https://github.com/finos/git-proxy/milestone/10) or whether to go for [2.2.0](https://github.com/finos/git-proxy/milestone/11)... There are at least 2 PRs also scoped for 2.1.0 that will need UI additions:

- #1332
  - User administration of SSH support 
- #1457
  - Password change dialog

SSH has been waiting for a long time so consensus at the meeting was (reluctantly) to do that first and then ask you to update #1507 to add the additional UI panels (or to collab with you on doing so). Hence, we've marked this for 2.2.0 (and as the only issue in that milestone currently). Does that work for you?

 Also, please see the below notes from the meeting agenda:
 
- there was a question about some additional issues you may or may not be taking care of with the PR (no worries if not, just needed to check and add them to the resolves list if they are getting done):
- We need to know if #1464 is redundant and can be closed (sounds like a yes, but wanted to be 100% sure)

Agenda except from #1506

>    - [ ] #1106 & #1176 & #1258 & #1297 & #1420 & #1495 
>      - #1507 
>      - Are these also solved/solvable? 
>        - #1123
>        - #1363
>        - #624
>      - Is #1464 still needed or does #1507 deal with all the same issues (and more)  
>        - See comment from @andypols: https://github.com/finos/git-proxy/issues/1420#issuecomment-4169017529- [ ] #1303
>      - What is the impact on renovate PRs?

On the final question, we're expecting it to close a few renovate PRs as we drop some dependencies in favour of new ones?

andypols:
Hey @kriswest 

> SSH has been waiting for a long time so consensus at the meeting was (reluctantly) to do that first and then ask you to update #1507 to add the additional UI panels (or to collab with you on doing so). Hence, we've marked this for 2.2.0 (and as the only issue in that milestone currently). **Does that work for you?**

I think the bigger issue is that planning has started to block releases rather than guide them. We used to release regularly; now worthwhile contributions sit for long periods and get pushed back further and further. SSH is a good example, and I’m concerned the UI work may end up the same way. 

I’d prefer a simpler approach: merge bug fixes and useful enhancements when they’re ready. Then make release review a standing agenda item at each meeting by asking, “Is there any reason not to release the latest changes?”

> * We need to know if [feat: add server-side pagination to API endpoints #1464](https://github.com/finos/git-proxy/pull/1464) is redundant and can be closed (sounds like a yes, but wanted to be 100% sure)

I did not implement the paging in the database - we tried to leave the db layer alone and just focus on the UI.  #1464 is not redundant, although its not really needed yet and can wait until the ui changes have been made.

> * Supress tooltip errors for private projects #1123

Yes, the caching has addressed this.  It will not include any project metadata

> * Ensure Git Proxy user interfaces meet WCAG 2.2 Level AA accessibility requirements #1363

It has probably improved but will still need to be done.

> * Extend Cypress tests for dashboard pages and components #624

No this still needs doing.  Cypress is hard for us as its blocked internally for security reasons.  The existing cypress tests are working locally, no idea why they are failing in the pr tests.

kriswest:
> I think the bigger issue is that planning has started to block releases rather than guide them. We used to release regularly; now worthwhile contributions sit for long periods and get pushed back further and further. SSH is a good example, and I’m concerned the UI work may end up the same way.

The delay on the SSH feature was a decision we made collectively because the wholesale refactor of the code conflicted with everything making it very hard to merge. For 2.1.0 we've picked that and a number of other (pretty much) dev complete PRs to get merged to try and get to release relatively quickly to start clearing our backlog, with the goal of getting back to quicker releases as soon as we've got that done. 

Nothing we've picked needs much new work apart from resolving conflicts and review/testing - we just had to deal with the expected conflict between your UI rebuild and a couple of the PRs that will add to the UI. There's not an easy solution to that other than one or the other going first. As per the meeting minutes, I offered to check-in with you as the meeting was keen to get both the UI and SSH etc. in ASAP. 

We also have some issues with Cypress testing here - but I've got a machine I can run them on. Perhaps we can get some collab going on resolving those so that the coverage doesn't dip. It might be worth merging this into a FINOS repo branch to do that?

kriswest:
> I think the bigger issue is that planning has started to block releases rather than guide them. 

I think we switched from making smaller changes to great big ones which caused this. SSH and a new UI are not small, but not quite as massive as the whole codebase refactor. Hopefully after these we'll have a run of smaller changes.

kriswest:
I think the cypress failure is down to this id: 
https://github.com/qube-rt/git-proxy/blob/d07c2f0aea8eb239d7c788f2ca8080c8886fa35a/src/ui/views/PushDetails/components/Reject.tsx#L132

Which the test expects to be `#reason` when its `#push-reject-reason`
https://github.com/qube-rt/git-proxy/blob/d07c2f0aea8eb239d7c788f2ca8080c8886fa35a/cypress/e2e/docker/pushActions.cy.js#L129

kriswest:
The dependency review check may be having trouble because its a PR from a fork?? Perhaps setting 
`comment-summary-in-pr: on-failure` at:

https://github.com/finos/git-proxy/blob/6981427c50a5e5ccd3e91a2d4229b27135f8dcc3/.github/workflows/dependency-review.yml#L22

will help?

src/db/file/pushes.ts
```diff
@@ -41,6 +49,99 @@ try {
...
+  }
+  let row = m.get(canonicalKey);
+  if (!row) {
+    row = emptyRepoActivityTabCounts();
+    m.set(canonicalKey, row);
+  }
+  row[tab] += 1;
+}
+
+function bumpMaxTimestampMs(
```
jescalada, line 68: It'd be great to add a docstring to explain what `bumpMaxTimestampMs` and `bumpCount` are doing; it isn't obvious at first glance 🙂

jescalada: Otherwise, more descriptive function names could also work as with `bumpUserActivityCount` below

src/db/index.ts
```diff
@@ -229,4 +262,68 @@ export const getAllProxiedHosts = async (): Promise<string[]> => {
...
+): void {
+  let row = counts.get(username);
+  if (!row) {
+    row = emptyRepoActivityTabCounts();
+    counts.set(username, row);
+  }
+  row[tab] += 1;
+}
+
+export const getUserActivityTabCountsByUsername = async (): Promise<
```
jescalada, line 292: Is there a reason why these functions (`bumpUserActivityCount`, `getUserActivityCountsByUsername`) were extracted to `src/db/index.ts`? Meanwhile the equivalent ones for repos are split into each of the database implementations (`src/db/file/pushes.ts`, `src/db/mongo/pushes.ts`).

I'm wondering since we'll add PostgreSQL support soon, perhaps we want to make this more consistent (by either extracting or splitting all helpers) or otherwise explain why not all functions were extracted.

jescalada: I just verified that the `bumpMaxTimestampMs` function is identical in both `file` and `mongo` implementations - seems like a prime candidate for extraction 😃

src/db/userProfilePushQuery.ts
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export function escapeRegex(s: string): string {
+  return s.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
```
jescalada, line 18: Should we make a string helpers file just for processing strings, or perhaps import an existing library for this? We're messing with strings in various ways elsewhere and might benefit from having it centralized.

Some other examples are:
- `normalizeRepoPathForMatch` in `src/activity/canonicalRemoteUrl.ts`
- `trimTrailingDotGit` in `src/db/helper.ts`

Feel free to ignore if too much hassle, this seems more appropriate for a separate PR anyways 🙂

src/ui/components/ActivityBadgeGroup/ActivityBadgeGroup.tsx
```diff
@@ -0,0 +1,103 @@
...
+  className: string;
+  label: (count: number) => string;
+}
+
+const badgeConfigs: BadgeConfig[] = [
+  {
+    status: 'pending',
+    icon: EyeIcon,
+    className:
+      'inline-flex items-center gap-1 rounded-full border border-[var(--borderColor-attention-muted)] bg-[var(--bgColor-attention-muted)] px-2 py-0.5 text-xs font-medium text-[var(--fgColor-attention)] !no-underline hover:border-[var(--borderColor-attention-emphasis)]',
```
jescalada, line 43: Would it make sense to extract the common elements in these `className` strings into a separate file for Tailwind classes? I think it might prevent issues with inconsistent or unmaintainable styles down the road.

src/ui/components/CustomButtons/CodeActionButton.tsx
```diff
@@ -14,133 +14,71 @@
...
-} from '@primer/octicons-react';
 import React, { useState } from 'react';
-import { PopperPlacementType } from '@material-ui/core/Popper';
-import Button from './Button';
+import { ActionMenu, IconButton, Text } from '@primer/react';
+import { CheckIcon, CodeIcon, CopyIcon, TerminalIcon } from '@primer/octicons-react';
+
+/** Git-style green (matches common "Code" affordance). */
+const codeButtonGreenClassName =
+  '!border-0 !shadow-none !bg-[#1a7f37] !text-white hover:!bg-[#136c2e] hover:!text-white active:!bg-[#115f2a]';
```
jescalada, line 24: Another candidate for class string extraction 🙂

src/ui/components/GitProxyUnderlineTabs/gitProxyUnderlineTabStripClass.ts
jescalada, line 1: Might be a good candidate for class string extraction to a shared file, though I'm not sure if it's more appropriate to have these split into style files for each individual component as we did here (we would sacrifice consistency for the sake of quick lookups vs. putting everything in a single `src/ui/tailwindClasses.ts`).

src/ui/components/Navbars/DashboardUserMenu.tsx
```diff
@@ -0,0 +1,92 @@
...
+    }
+  };
+
+  if (!user) {
+    return (
+      <button
+        type='button'
+        data-testid='dashboard-header-sign-in'
+        onClick={goLogin}
+        className='rounded-md border border-solid border-[rgba(240,246,252,0.35)] bg-black px-3 py-1 text-sm font-medium leading-normal text-[rgba(240,246,252,0.92)] antialiased [font-family:var(--fontStack-sansSerif)] hover:border-[rgba(240,246,252,0.5)] hover:bg-neutral-950 hover:text-white focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-[var(--header-accent)] focus-visible:ring-offset-2 focus-visible:ring-offset-[var(--header-bgColor)]'
```
jescalada, line 59: Perhaps button styles and colors could be extracted here?

src/ui/services/repo.ts
```diff
@@ -20,6 +20,94 @@ import { Repo } from '../../db/types';
...
+    if (aMs !== bMs) {
+      return aMs - bMs;
+    }
+  }
+  return compareRepoName(a, b, 'asc');
+};
+
+export const sortRepoViews = (
+  repos: RepoView[],
+  sort: RepoSortField,
```
jescalada, line 84: I think the import for `RepoSortField` is missing in this file 🤔

src/ui/views/PushDetails/components/StepsTimeline.tsx
```diff
@@ -14,287 +14,187 @@
...
+                <div className='min-w-0'>
+                  <Text as='p' className='m-0 mb-1 text-xs font-semibold text-(--fgColor-muted)'>
+                    Logs ({step.logs.length})
+                  </Text>
+                  <div className='min-w-0 space-y-1'>
                     {step.logs.map((log: string, logIndex: number) => (
-                      <div key={logIndex} className={classes.logItem}>
+                      <div
+                        key={logIndex}
+                        className='min-w-0 rounded-sm border-l-2 border-(--borderColor-muted) bg-(--bgColor-muted) px-2 py-1 font-mono text-xs break-words text-(--fgColor-default)'
```
jescalada, line 153: These style strings look like they could be extracted

jescalada (approved):
@andypols Fantastic contributions! Tailwind makes the UI code a lot more readable, and the UI is looking much, much better. 😃 

I'm wondering if we could/should do some refactoring for the `className` strings, I think we might benefit from having these centralized (especially for certain elements such as buttons, form inputs, etc.).

Other than a few minor points, this is looking great!

src/ui/components/ActivityBadgeGroup/ActivityBadgeGroup.tsx
andypols: I’d avoid extracting Tailwind class strings into a separate file just for reuse. Tailwind’s philosophy is utility-first and markup-local: you generally keep the Tailwind classes on the element they style. A core benefit of Tailwind is that styles stay colocated with the markup, which makes components easier to understand and reduces the chance of orphaned styles.

If the same group of classes is repeated in several places, I’d treat that as a signal that we may want to extract a component, or possibly a variant helper, rather than creating shared class constants.

src/db/file/pushes.ts
andypols: Thats fair - not the clearest naming.  Although they are placeholders and will be probably get replaced when we implement #1464

src/service/scmMetadata.ts
```diff
@@ -0,0 +1,164 @@
...
+import { trimTrailingDotGit } from '../db/helper';
+import { getGitProvider } from './gitProviders';
+import type { SCMRepositoryMetadata } from './gitProviders';
+
+export type { SCMRepositoryMetadata };
+
+/** Repo card enrichment (description, language, license, fork, avatars) changes slowly. */
+const SUCCESS_TTL_MS = 24 * 60 * 60 * 1000;
+/** Cooldown after a failed fetch before treating cache as stale again. */
+const FAILURE_TTL_MS = 15 * 60 * 1000;
```
fabiovincenzi: Consider exposing as env vars with current values as defaults. 24h is hard to test against, and different deploys may want different values

src/db/index.ts
```diff
@@ -229,4 +262,68 @@ export const getAllProxiedHosts = async (): Promise<string[]> => {
...
+  Map<string, RepoActivityTabCounts>
+> => {
+  const sink = start();
+  const [users, pushes] = await Promise.all([sink.getUsers(), sink.getPushes({ type: 'push' })]);
```
fabiovincenzi, line 296: getUserActivityTabCountsByUsername` loads **every** push document into memory on every `GET /api/v1/user`. Cost scales with the full push history and grows forever.

Suggest a DB-side `$group` aggregation so Mongo returns the counts directly. Same issue in `getRepoPushRollupsByCanonicalUrl`.

src/db/index.ts
```diff
@@ -178,6 +190,15 @@ export const canUserCancelPush = async (id: string, user: string) => {
 
 export const getSessionStore = (): MongoDBStore | undefined => start().getSessionStore();
 export const getPushes = (query: Partial<PushQuery>): Promise<Action[]> => start().getPushes(query);
+export const getPushesForUserProfile = async (username: string): Promise<Action[]> => {
```
fabiovincenzi: The route handler in `routes/users.ts` already calls `db.findUser(username)` for the 404 check, then this wrapper calls `findUser` again. Also the two paths behave differently when the user doesn't exist (route → 404, wrapper → `[]`).
Could the wrapper accept the already loaded `User`?

fabiovincenzi (commented):
Looks good!
Can you open a follow-up issue for adding SCM token support so the remaining work stays tracked?

Also, the `Linting` job fails 
Root cause is in https://github.com/finos/git-proxy/blob/d07c2f0aea8eb239d7c788f2ca8080c8886fa35a/.github/workflows/lint.yml#L32 — `npm install --workspaces` does not install root devDependencies (where `eslint` lives).

Suggested fix: replace with `npm ci`.

Extend `allow-dependencies-licenses` in `.github/workflows/dependency-review.yml` with the three packages flagged by the failing workflow: `bowser`, `focus-visible`, and `spdx-exceptions`.

src/ui/components/ActivityBadgeGroup/ActivityBadgeGroup.tsx
jescalada: Insightful answer, thanks 😃

kriswest:
@andypols we've got some accessibility fixes to contribute to this, might be easier to get you merged and then raise that as a new PR... We'll get the issue raised in the meantime.

Are you able to sort the dependency issues breaking the CI build so this can move forward?

kriswest:
Not sure what we're waiting for on the 2.1.0 release now? Just @jescalada being on holiday and something to do with the release drafter?

jescalada:
Seems that the Dependency Review error is coming from a Vitest 3 vulnerability, which is being fixed in #1564. Perhaps we could merge that one first?

@kriswest The only thing left for the 2.1 release is #1627, just waiting on a final approval from @andypols

andypols:
> Seems that the Dependency Review error is coming from a Vitest 3 vulnerability, which is being fixed in #1564. Perhaps we could merge that one first?

Sounds sensible
> 
> @kriswest The only thing left for the 2.1 release is #1627, just waiting on a final approval from @andypols

Done.  Chocks away...

src/db/index.ts
andypols: Its deliberate.  It should definitely move to the db, but I wanted to limit the blast radius of this PR. To be honest I don't believe performance is an issue for any of the users of git-proxy. I vote we do this when we need it.

src/service/scmMetadata.ts
andypols: Fixed

src/db/index.ts
andypols: Good catch - thanks - fixed

jescalada (approved):
LGTM! 🚀 @fabiovincenzi Thoughts before merging?

fabiovincenzi (approved):
Looks good!

PR #1513, opened 2026-04-24
test: extend Cypress coverage for dashboard pages

### LINKED ISSUE #624: Extend cypress tests for  each page of the UI
Write cypress test to cover all functionality and UI elements on each page to minimize UI-related bugs and regression to ensure consistent functionality across all UI pages.

dcoric:
Closes #624

## Summary
- add Cypress coverage for dashboard pages, navigation, profile, settings, user list, repo list/details, push list/details, and action flows
- add test-only cleanup routes and Cypress commands for repeatable data cleanup
- add stable UI selectors and small routing/rendering fixes needed by the tests

## Verification
- docker compose -f docker-compose.yml -f docker-compose.ci.yml up -d --build --wait
- CYPRESS_BASE_URL=http://localhost:8081 CYPRESS_API_BASE_URL=http://localhost:8081 CYPRESS_GIT_PROXY_URL=http://localhost:8000 CYPRESS_GIT_SERVER_TARGET=git-server:8443 npm run cypress:run:docker
- Full Docker Cypress suite passed: 71/71


dcoric:
I understand that a UI redesign is currently underway, so some of the changes in this PR may already be superseded or become obsolete soon. This work had been sitting on my side for a while, and I thought it was better to clean it up and share it rather than leave it abandoned. Even if it does not land as-is, parts of it may still be useful or reusable.

CYPRESS_PLAN.md
jescalada, line 1: We might want to remove this file 🙂

cypress/e2e/docker/pushActions.cy.js
```diff
@@ -217,6 +230,9 @@ describe('Push Actions (Approve, Reject, Cancel)', () => {
 
   describe('Negative: unauthorized reject', () => {
     beforeEach(() => {
+      // Rate-limit guard: wait to avoid 429 from rapid push creations
+      // eslint-disable-next-line cypress/no-unnecessary-waiting
+      cy.wait(2000);
```
jescalada: I wonder if there's a way to get around the 429 error without calling `cy.wait` - since using lots of these will slow down overall E2E speeds significantly. Note that on `main`, the E2E workflow runs in about 3 mins, meanwhile it currently takes about 6 minutes on this PR.

cypress/e2e/error-pages.cy.js
```diff
@@ -0,0 +1,69 @@
...
+  beforeEach(() => {
+    cy.login('admin', 'admin');
+  });
+
+  afterEach(() => {
+    cy.logout();
+  });
+
+  // --- 9.1 Unknown route shows 404 ---
+  it('9.1 — Unknown route shows 404 page', () => {
```
jescalada: We may want to remove these identifiers (`9.1`, etc.) since they only make sense within the agent planning context

cypress/e2e/push-details.cy.js
jescalada, line 1: Lots of comments here that could be condensed or cleaned up, same for other files 🙏🏼

cypress/e2e/push-details.cy.js
```diff
@@ -0,0 +1,324 @@
...
+      .request({
+        method: 'GET',
+        url: `${Cypress.config('baseUrl')}/api/v1/push/${pushId}`,
+        failOnStatusCode: false,
+        timeout: 30000,
+      })
+      .then((res) => {
+        if (res.status !== 200) {
+          // eslint-disable-next-line cypress/no-unnecessary-waiting
+          cy.wait(3000);
```
jescalada: This `cy.wait` along with the timeout above (30s) and the 30 retries might be a massive source of E2E execution delays if we ever have any trouble with the API (error 403, 500, etc.). I'd suggest checking for `res.status` failure codes *other* than "push not yet found" and "success" (404, 200) and re-throwing such errors so the test fails early.

jescalada (changes requested):
Looks good so far! 🚀  Just a few things we ought to fix up:
- Cleaning up agent-related artifacts
- Figuring out if we can reduce E2E test execution somewhat (originally 3 mins.)
  - Potentially we could remove or reduce the `cy.wait` calls?
- Fixing up the 200 check, which isn't the only guaranteed result after fetching if the push was created

cypress/e2e/docker/pushActions.cy.js
```diff
@@ -49,10 +48,18 @@ describe('Push Actions (Approve, Reject, Cancel)', () => {
     cy.logout();
   });
 
-  afterEach(() => {
+  afterEach(function () {
```
jescalada, line 51: For consistency:

```suggestion
  afterEach(() => {
```

cypress/e2e/docker/pushActions.cy.js
jescalada, line 1: We may want to leave the TODO comments in this file - unless we're sure the action items have already been dealt with

cypress/e2e/repo-details.cy.js
jescalada, line 1: Nit: would be nice to add spacing between each `it` block in this and other test files 🙂

jescalada (commented):
Looks good, just a few more nitpicks from me 🙂 

I noticed that the E2E tests have slowed down significantly (12-16 mins. in the last run). However, on closer inspection, it seems that the Docker setup step *before* the tests is unusually slow. I'm playing around with fixes for the `Dockerfile` in #1521 - we can safely ignore the slowdown in this PR as it's not due to the new tests.

PR #1514, opened 2026-04-29
feat(config): require merged top-level defaults

### LINKED ISSUE #1287: Require top-level config values while allowing to override config file
Currently, top-level config values are not required, which means that the generated config types are optional. This forces us to use fallbacks, check if values are present or cast them:

```ts
app.use(
  session({
    store: db.getSessionStore(),
    secret: config.getCookieSecret() as string,
    resave: false,
    saveUninitialized: false,
    cookie: {
      secure: 'auto',
      httpOnly: true,
      maxAge: (config.getSessionMaxAgeHours() || DEFAULT_SESSION_MAX_AGE_HOURS) * 60 * 60 * 1000,
    },
  }),
);
```

Adding the parameters to the "required" list will break tests that override the config entirely since it errors out if the entries aren't present. The best solution seems to be to post-process the generated config types so that the top-level properties are wrapped in `Required` types. This allows stronger typing as well as overriding.

We should also consider how we can set default values for the config - see the related discussion: https://github.com/finos/git-proxy/pull/1202#discussion_r2550326695

**Additional context**
https://github.com/finos/git-proxy/pull/1202#discussion_r2511095369
https://github.com/finos/git-proxy/pull/1202#discussion_r2550326695

dcoric:
## Summary

Fixes #1287.

This updates config loading so user override files can keep using the generated optional `GitProxyConfig` type, while the merged runtime configuration is narrowed to a stricter `FullGitProxyConfig` after defaults are applied and verified.

## Changes

- Add a `FullGitProxyConfig` type for the merged runtime configuration.
- Validate that default-backed top-level config values are present after merging defaults and user overrides.
- Keep deprecated compatibility fields optional where the default config does not set them.
- Remove fallback values from config accessors that now read from the validated merged config.
- Remove the duplicate session max-age fallback in service session setup.
- Add tests covering defaulted top-level values and partial user overrides.

## Validation

- `npm run check-types:server`
- `npm test -- test/testConfig.test.ts`
- Commit hook ran prettier, eslint fixes, and `npm run lint`


src/config/index.ts
```diff
@@ -25,8 +25,55 @@ import { getConfigFile } from './file';
 import { validateConfig } from './validators';
 import { handleErrorAndLog, handleErrorAndThrow } from '../utils/errors';
 
+// Deprecated compatibility fields are still optional because the defaults do not set them.
+type OptionalTopLevelConfigKey = 'proxyUrl' | 'sslCertPemPath' | 'sslKeyPemPath';
+type RequiredTopLevelConfigKey = Exclude<keyof GitProxyConfig, OptionalTopLevelConfigKey>;
```
jescalada, line 30: Wondering if we should go ahead and deprecate the compatibility fields to simplify the implementation here 🤔

src/config/index.ts
```diff
@@ -25,8 +25,55 @@ import { getConfigFile } from './file';
...
+  'uiRouteAuth',
+  'urlShortener',
+] as const satisfies readonly RequiredTopLevelConfigKey[];
+
+type MissingRequiredTopLevelConfigKeys = Exclude<
+  RequiredTopLevelConfigKey,
+  (typeof REQUIRED_TOP_LEVEL_CONFIG_KEYS)[number]
+>;
+type AssertNever<T extends never> = T;
+type _RequiredTopLevelConfigKeysAreExhaustive = AssertNever<MissingRequiredTopLevelConfigKeys>;
```
jescalada, line 64: After bringing in the latest `main`, a type issue popped up here. It can be reproduced when running `npm run check-types:server`:

image

jescalada (changes requested):
LGTM after fixing CI 🚀 Just wondering if we should go ahead and deprecate the old fields to simplify the implementation a bit. It would also help clean up our config files and docs.

src/config/index.ts
```diff
@@ -25,8 +25,55 @@ import { getConfigFile } from './file';
...
+  'domains',
+  'plugins',
+  'privateOrganizations',
+  'rateLimit',
+  'sessionMaxAgeHours',
+  'sink',
+  'tempPassword',
+  'tls',
+  'uiRouteAuth',
+  'urlShortener',
```
jescalada, line 56: Nevermind, I just realized that you designed this to catch config keys with missing values 🙂 This one got brought in from `main` and just needs updating:

```suggestion
  'upstreamProxy',
  'urlShortener',
```

dcoric:
Thanks for the review @jescalada! Pushed two commits:

1. **`fix(config): add upstreamProxy to required top-level keys`** — addresses the CI failure you flagged after my main merge brought in `upstreamProxy`.
2. **`refactor(config): remove deprecated proxyUrl, sslKeyPemPath, sslCertPemPath fields`** — takes you up on the deprecation suggestion. Drops the deprecated fields from the schema, regenerates the TS types and reference docs, removes the legacy fallback logic, and simplifies `FullGitProxyConfig` to plain `Required<GitProxyConfig>` (so the `OptionalTopLevelConfigKey` indirection is gone).

This is a breaking change for any users still relying on the deprecated fields, but since they were already marked `deprecated: true` in the schema and `tls.key` / `tls.cert` have been the documented path for a while, hopefully that's acceptable. Happy to split the deprecation removal into a separate PR if you'd prefer to merge the CI fix on its own first.

jescalada (commented):
After the recent changes, using deprecated configs (those that still contain the `proxyUrl`/`sslKeyPemPath`/`sslCertPemPath`) or using a config with a non-existent key (such as `invalidKey`) throws the following error when starting the proxy:

```bash
[0] Error: Failed to load configuration: Invalid value for key "invalidKey" on GitProxyConfig. Expected boolean but got false
[0]     at handleErrorAndThrow (/home/juan/Desktop/Juan/Projects/git-proxy/src/utils/errors.ts:32:9)
[0]     at defaultSettings (/home/juan/Desktop/Juan/Projects/git-proxy/src/config/index.ts:426:3)
[0]     at Object.<anonymous> (/home/juan/Desktop/Juan/Projects/git-proxy/src/config/index.ts:428:1)
[0]     at Module._compile (node:internal/modules/cjs/loader:1562:14)
[0]     at Object.transformer (/home/juan/Desktop/Juan/Projects/git-proxy/node_modules/tsx/dist/register-D46fvsV_.cjs:3:1104)
[0]     at Module.load (node:internal/modules/cjs/loader:1313:32)
[0]     at Function._load (node:internal/modules/cjs/loader:1123:12)
[0]     at TracingChannel.traceSync (node:diagnostics_channel:322:14)
[0]     at wrapModuleLoad (node:internal/modules/cjs/loader:217:24)
[0]     at Module.require (node:internal/modules/cjs/loader:1335:12)
[0] 
[0] Node.js v22.13.1
```

Note the "`expected boolean but got X`" bit - it still fails when setting a boolean value like this:

`"invalidKey": false`

If removing those fields, we would want a more descriptive/accurate error so people know those fields are missing or no longer available, rather than the type being mistaken as the current error message implies.

---

I'm thinking the deprecation removal would be best implemented in a separate PR even if the config default requirement implementation is a bit messy on this PR. I'm happy to merge the PR in the previous state with the CI fixes added in.

Pinging @kriswest for thoughts on removing deprecated `proxyUrl`/`sslKeyPemPath`/`sslCertPemPath` soon - I don't think anybody in v2.0 is still using these at all...

dcoric:
Reverted the `refactor(config): remove deprecated proxyUrl, sslKeyPemPath, sslCertPemPath fields` commit ([4234c16](https://github.com/finos/git-proxy/commit/4234c16fbddca7abc10f08fc4bd2285e0a58ec44)) — agreed it's better to keep the deprecation removal out of this PR. The deprecated fields stay in the schema (still flagged `deprecated: true`) and the legacy fallback logic is restored, so this avoids the breaking change for 2.x users. Happy to follow up with a separate PR targeting 3.0 for the actual removal, with the more descriptive error message you mentioned for missing/removed fields.

This PR is now back to just the config defaults work (the original goal + the `upstreamProxy` CI fix). Ready for another look @jescalada.

dcoric:
Tracked the eventual removal as #1545 (3.0). That issue also captures your suggestion about a clearer error message for removed/unknown fields.

jescalada (approved):
LGTM

PR #1516, opened 2026-05-01
docs: add release branching documentation

jescalada:
Closes #1483.

The updated workflows have been added in #1520, all documentation here reflects the changes included in that PR.

We should decide which versions will be supported (with occasional bugfixes). I thought about supporting `1.20` and `2.x`, and then dropping support for `1.20` when `3.0` comes out, but `v1` has drifted so much from `2.0` that we'd be better off dropping support entirely to avoid all the extra work. Either way, this doc should be updated to inform users/maintainers about which GitProxy versions have active support.

### Notes
- I wrote this with the assumption that we'll be doing cascade-based merges rather than cherry-picking, as @coopernetes suggested in #1483
  - For reference, some [pros and cons of cascade merges](https://softwareengineering.stackexchange.com/questions/460758/is-cascade-merging-forward-porting-riskier-than-backporting)
- We might want to add a section to explain which versions will be supported, in order to limit the amount of extra work and merge conflicts when cascading bugfixes. For example, only the latest of each major version (currently `1.19` and `2.0`)
- I mentioned that releases are supposed to follow a cadence, but I never specified what that cadence should be. We should decide on an appropriate cadence (say, every three months) or remove that part entirely.

Would love to hear your thoughts! @finos/git-proxy-maintainers


docs/Releases.md
```diff
@@ -0,0 +1,128 @@
...
+Version bumps are computed automatically by our release drafter workflow, based on this [configuration file](https://github.com/finos/git-proxy/blob/main/.github/release-drafter-config.yml). Thus, we encourage contributors to label their PRs and commits appropriately as described below.
+
+## Pull Request Conventions
+
+Both release notes and the release drafter workflow rely on [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/)-style PR titles for automatic tagging.
+
+- `feat:` labels the PR as a feature (minor bump)
+- `fix:` labels the PR as a bugfix (patch bump)
+- `break:` or `feat!:` (exclamation mark suffix) labels the PR as a breaking change (major bump)
+- Other prefixes map to the maintenance categories defined in `release-drafter.yml`
```
kriswest, line 43: What about the ! syntax for marking breaking changes (prompts a major version bump)

docs/Releases.md
```diff
@@ -0,0 +1,128 @@
...
+
+1. Reviews the draft for the relevant `release/*` branch
+2. Edits the release notes if necessary (for instance, adding a migration guide for breaking changes)
+3. Clicks Publish.
+
+Publishing creates a git tag, and triggers the `npm.yml` workflow described below.
+
+### Publish to NPM
+
+This workflow runs whenever you publish a GitHub release.
```
kriswest: Does the workflow include any automated chacks on the version number in package.json matching the branch name? If not we should add that as it can save headaces/mistakes

docs/Releases.md
```diff
@@ -0,0 +1,128 @@
...
+
+#### Troubleshooting
+
+##### My draft release got overwritten!
+
+This usually happens when forgetting to publish the draft for an older release branch (during cascade merges).
+
+Just re-run the `release-drafter.yml` workflow from the related release branch. Example: If your `1.2.x` draft got overwritten, go to the [Release Drafter action page](https://github.com/finos/git-proxy/actions/workflows/release-drafter.yml), click on the run matching the latest `release/1.2` branch, and then click "Re-run all jobs".
+
+##### I published the draft release and forgot to bump the `package.json`s!
```
kriswest, line 123: The release flow could automatically detect this happening (see previous comment)

kriswest (commented):
LGTM - one thought on automatically checking alignment between package.json versions and branch name.

This could/should all be added to AI steering files - if an AI can pick up that a release is being prepared it can comment on the steps to be taken, merges to be performed and potentially help execute those steps correctly.

@re-vlad @jescalada

docs/Releases.md
jescalada: Made an issue for this, as it might require another round of NPM deployment testing: #1533

docs/Releases.md
jescalada: Included this in #1533 and linked for context!

docs/Releases.md
jescalada: Updated the wording here and added the "!" bit as an extra entry.

jescalada:
@kriswest Ready for another look 👍🏼

kriswest (approved):
LGTM

Issue #1121, opened 2025-07-27
Email notifications for reviews


kriswest:
**Is your feature request related to a problem? Please describe.**
When a user pushes and a review is needed, they currently need to manually notify the reviewers (if any are set-up). When reviews are submitted, the reviewer needs to manually notify the contributor that they have done so. In SCM services (GitHub/GitLab) email notifications are generated for both reviews requested and submitted which reduces the toil in the whole process.

**Describe the solution you'd like**
Generate email notifications when:

- pushes are submitted for review
- pushes change state (approved, rejected, cancelled).

Support templating or template variables through configuration for emails generated.

**Describe alternatives you've considered**
Teams integration for notifications.

**Additional context**
There is some very basic email functionality implemented at: src\service\emailSender.js however it appears to be unused.

PR #1519, opened 2026-05-05
feat(event-handlers): introduce event handler system for GitProxy lifecycle events

### LINKED ISSUE #1449: Extensibility through Event Hooks
# Summary

Organizational deployments of Git Proxy often need to run organization-specific integrations before and after Git Proxy actions are performed, such as clone, pull, and push.

Examples include:

- integration with internal request or approval systems
- fine-grained entitlement system integrations
- chat or notification integrations
- organization-specific scanners or processing hooks

## Proposal

Introduce a simple, lightweight, and well-abstracted event system that external integrations can hook into.

These interfaces should be abstracted from Git Proxy internals and should expose only the logical operations Git Proxy is performing. This would allow organizations to extend behavior without coupling directly to internal workflow or chain implementation details.

## Example

A simple example of an email handler:

```ts
proxy
  .registerEventHandler()
  .onPull()
  .onCompleted(myEmailHandler);

async function myEmailHandler(repo: RepositoryContext, user: UserContext): Promise<void> {
  // Check whether this is the user's first pull of the repository
  if (await isFirstPull(repo, user)) {
    const emailBody = `
Welcome to Git Proxy.

Here is everything you need to get started before submitting your first push request...
`;

    await email.send(user.emailAddress, "Getting started with Git Proxy", emailBody);
  }
}
```

## Proposed API

```ts
interface IProxyEventRegistry {
  onPull(): IActionEventHandler;
  onClone(): IActionEventHandler;
  onPush(): IActionEventHandler;
}
```

```ts
interface IActionEventHandler {
  onStarted(handler: ActionEventCallback): IActionEventHandler;
  onCompleted(handler: ActionEventCallback): IActionEventHandler;
  onError(handler: ActionErrorEventCallback): IActionEventHandler;
  onPermissionDenied(handler: ActionEventCallback): IActionEventHandler;
}
```

```ts
type ActionEventCallback = (
  repo: RepositoryContext,
  user: UserContext
) => void | Promise<void>;

type ActionErrorEventCallback = (
  repo: RepositoryContext,
  user: UserContext,
  error: Error
) => void | Promise<void>;
```

## Notes

While the existing chain could theoretically be modified to insert additional actions, that approach is cumbersome and invasive because it requires integrations to hook directly into the core execution flow of Git Proxy.

That would make future workflow refactoring significantly harder, since external users would then become dependent on internal chain handler interfaces and execution semantics.

A dedicated event model would provide a cleaner extension point, with lower coupling and a clearer contract for organizational integrations.

Andreybest:
resolves #1449.

Introduces a lightweight, async, observer-only event system so external integrations (notifications, audit feeds) can react to push/pull lifecycle events without coupling to chain internals.

Event hooks can be added the same way as the `plugins` in `config.proxy.json`

Supports 2 operations: `pull` and `push`
With 5 phases:
- `started`
- `completed`
- `error`
- `permissionDenied`
- `pendingReview`

This is not well documented, and not sure where to add info on that, would appreciate suggestions on that matter :)


andypols:
@Andreybest Could I vote for an additional phase?

We have several custom actions that run when a push is first received and the proxy blocks it for review. This state is not really `permissionDenied` or `error`; it is closer to `pendingReview`.

We also have custom actions that would work really well as plugin events. For example, we:

- run an AI-based security audit of the changes
- create an AI summary of the changes to help the reviewer
- notify reviewers when there is a push to review

It would be great if plugins could also add data that can be rendered in the UI.

This would be a hugely helpful change, as it would allow us to separate our custom code from the base git-proxy. Keeping the two in sync is painful at the moment.

Andreybest:
Hey @andypols ! Thanks for the suggestion, this seems like a good use cases for events. Will try to add mechanisms that would be able to support your use cases :)
Will ping you on that later.

re-vlad (commented):
Just a wish: while the code is well-documented, there's no external docs (README, docs, etc.) explaining how to use the event system. This should be added.

proxy.config.json
```diff
@@ -140,6 +140,7 @@
   "contactEmail": "",
   "csrfProtection": true,
   "plugins": [],
+  "eventHandlers": [],
```
re-vlad, line 152: The config schema shows the eventHandlers array, but there are no examples of how to configure event handlers in proxy.config.json. Consider to document example

src/eventHandlers/dispatcher.ts
```diff
@@ -0,0 +1,141 @@
...
+  url: action.url,
+  project: action.project,
+  name: action.repoName,
+});
+
+const buildUserContext = (action: Action): UserContext | undefined => {
+  if (!action.user && !action.userEmail) {
+    return undefined;
+  }
+  return { username: action.user, email: action.userEmail };
```
re-vlad, line 38: EventDetails.user and repository must be immutable copies (using String()) of Action fields — not references. Otherwise, async handlers may see stale or modified data if Action is mutated later in the chain.  This rather critical I believe, but easy to fix:
``` 
return { 
  username = action.user ? String(action.user) : undefined;
  email = action.userEmail ? String(action.userEmail) : undefined;
}```.

**Review by re-vlad — CHANGES_REQUESTED** [review]

File: src/eventHandlers/dispatcher.ts [hunk 3]
```diff
@@ -0,0 +1,141 @@
...
+  ActionEventCallback,
+  ActionPhase,
+  EventDetails,
+  ProxyOperation,
+  RepositoryContext,
+  UserContext,
+} from './types';
+
+const buildRepositoryContext = (action: Action): RepositoryContext => ({
+  url: action.url,
```
re-vlad, line 29: The same problem like line 38, don't return references, return values instead:
```
name: action.repoName? String(action.repoName) : undefined,
url: String(action.url),
```

re-vlad (changes requested):

proxy.config.json
Andreybest: Added documentation for that in architecture.md. Thanks!

src/eventHandlers/dispatcher.ts
Andreybest: Both values are strings, and strings are copied by value, not by reference, thus why this is unnecessary.

src/eventHandlers/dispatcher.ts
Andreybest: Responded on comment above

Andreybest:
> @Andreybest Could I vote for an additional phase?
> 
> We have several custom actions that run when a push is first received and the proxy blocks it for review. This state is not really `permissionDenied` or `error`; it is closer to `pendingReview`.
> 
> We also have custom actions that would work really well as plugin events. For example, we:
> 
> * run an AI-based security audit of the changes
> * create an AI summary of the changes to help the reviewer
> * notify reviewers when there is a push to review
> 
> It would be great if plugins could also add data that can be rendered in the UI.
> 
> This would be a hugely helpful change, as it would allow us to separate our custom code from the base git-proxy. Keeping the two in sync is painful at the moment.

Hey @andypols! Added a new phase - `pendingReview`. Which indicates that the Action is `blocked` and not `autoApproved` and not `autoRejected`

On "add data that can be rendered in the UI", not sure if this should be the part of this PR, since the current architecture of event hooks is to inform a plugin that something happened in a manner of "fire-and-forget". And as I understand this will also require changes to database schema itself.

Please let me know if I understood what you have meant. :)

andypols:
> Hey @andypols! Added a new phase - `pendingReview`. Which indicates that the Action is `blocked` and not `autoApproved` and not `autoRejected`
> 
> On "add data that can be rendered in the UI", not sure if this should be the part of this PR, since the current architecture of event hooks is to inform a plugin that something happened in a manner of "fire-and-forget". And as I understand this will also require changes to database schema itself.
> 
> Please let me know if I understood what you have meant. :)

@Andreybest Thats sounds perfect.  It makes sense that the UI is in a different PR, although not sure about the need for a scheme change (we have this and did not need one)

src/eventHandlers/builtin/consoleLogger.ts
jescalada, line 1: This is a great example file to use for the Event Handler documentation 🙂

src/eventHandlers/types.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+/**
+ * Registry through which integrations subscribe to git-proxy lifecycle events.
+ *
+ * Handlers are observers only - they cannot block, modify, or reject git
+ * operations. For blocking/policy logic, write a chain plugin instead.
+ *
+ * Handlers run asynchronously, fire-and-forget. Errors thrown by a handler
+ * are caught and logged but never propagate to the git response.
```
jescalada: These details would be best explained in a separate document - things like when to choose an event handler over a plugin as they're superficially similar 👍🏼

src/proxy/actions/Action.ts
```diff
@@ -35,6 +35,7 @@ class Action {
   errorMessage?: string | null;
   blocked: boolean = false;
   blockedMessage?: string | null;
+  permissionDenied: boolean = false;
```
jescalada, line 56: I understand `permissionDenied` is set exclusively when `checkUserPushPermission` fails. I wonder if there's a more elegant solution - perhaps storing the name of the step where the `Action` failed. We could also infer this from the `lastStep` so we wouldn't need the extra field.

jescalada: I see the value of handling the specific case of  `permissionDenied`. I'm just wondering if we could prevent future bloat from happening as people try to add event handling to other, arbitrary push actions.

OTOH, I'd say it's okay to keep it if it makes the event handler implementation more obvious.

src/eventHandlers/builtin/consoleLogger.ts
```diff
@@ -0,0 +1,67 @@
...
+  registry: IProxyEventRegistry,
+) {
+  for (const op of operations) {
+    const builder = op === 'push' ? registry.onPush() : registry.onPull();
+    for (const phase of phases) {
+      if (phase === 'started') builder.onStarted(log);
+      else if (phase === 'completed') builder.onCompleted(log);
+      else if (phase === 'pendingReview') builder.onPendingReview(log);
+      else if (phase === 'error') builder.onError(log);
+      else if (phase === 'permissionDenied') builder.onPermissionDenied(log);
```
jescalada, line 64: Might want to have an `else` to handle things if someone added an additional `ActionPhase` but forgot to update the switch here.

src/eventHandlers/dispatcher.ts
jescalada: I think Vlad's point here is that there is no object-level protection, in other words the fields can be overwritten unless we enforced the contract via something like `Object.freeze(details.user)` after defining `details` below.

For example, the following event handler would affect any subsequent handlers that use the `EventDetails`:

```ts
export default new EventHandlerPlugin(() => {
  const redact = (details: EventDetails) => {
    if (details.user) {
      details.user.email = '<redacted>';
    }
    sendTo3rdPartyApp(details);
  };
  registry.onPush().onCompleted(redact);
});
```

jescalada (changes requested):
@Andreybest Thanks for the contribution! It's looking great so far.

It'd be great to have a simple doc on how to configure a preexisting handler (such as the `consoleLogger`), and how to develop a new one. A short step-by-step guide will make it easier for everyone to adopt.

It'd be fantastic to have an extra builtin handler for sending emails as mentioned in #1121, although we could add it in a separate PR.

src/eventHandlers/dispatcher.ts
Andreybest: You are right, haven't thought about that, but I don't think that we should freeze object. This is proper, im not against it, but probably plugin code or dependencies will change object (which I do not agree they should do that), so I made a cloning of EventDetails object for each handler invocation.

src/proxy/actions/Action.ts
Andreybest: Lets keep it like this for now :)

src/eventHandlers/builtin/consoleLogger.ts
Andreybest: Makes sense, thanks for pointing out! Added!

src/eventHandlers/types.ts
Andreybest: Thanks! Added information on this example as a separate page in website docs.

src/eventHandlers/builtin/consoleLogger.ts
Andreybest: Thanks! Added information on this example as a separate page in website docs.

Andreybest:
Thanks for the review @jescalada! Added changes to your comments. On matter of built in handler for email notifications - I suggest put it as a separate PR, since the specifications for events are not 100% final and could be changed later in this PR :)

src/eventHandlers/dispatcher.ts
```diff
@@ -38,6 +38,21 @@ const buildUserContext = (action: Action): UserContext | undefined => {
...
+/**
+ * Produces a per-handler copy of the event details, including fresh nested
+ * objects. Handlers run fire-and-forget and would otherwise share a single
+ * `details` reference, so a handler mutating `details.user` or
+ * `details.repository` would corrupt what later handlers observe. Cloning
+ * isolates each handler without forcing plugin authors to deal with frozen
+ * objects. `error` is shared by reference — Error instances do not clone
+ * cleanly, and it is the same underlying failure for every handler.
+ */
+const cloneDetails = (details: EventDetails): EventDetails => ({
```
jescalada, line 50: This looks good to me, but we should add a regression test for when the `EventDetails` get changed (since adding a new field to `EventDetails` wouldn't inform contributors to check this function too). We should add a test where we mutate a nested field inside handler 1, and then assert handler 2 sees clean data rather than handler 1's mutation.

Also, the test should clearly point to `cloneDetails` so contributors know where to fix.

src/eventHandlers/dispatcher.ts
jescalada: LGTM, though check my previous comment on `cloneDetails` 👍🏼

jescalada (commented):
@Andreybest Thanks for adding documentation! I have one more comment about the `EventDetails` cloning bit. I'd also want an extra check from @finos/git-proxy-maintainers to make sure things are good to merge.

src/eventHandlers/dispatcher.ts
Andreybest: Hey @jescalada! Added test to cover this :)

jescalada (approved):
LGTM! 🚀 Thoughts? @finos/git-proxy-maintainers

jescalada (approved):
LGTM, just needs a second approval for the latest formatting changes

andypols:
This is something we really (getting requests all the time).  It was a common request at the roadmap and don't want to implement our own version, but we do this this functionality soon.  I'm happy  what do the other @finos/git-proxy-maintainers think?

kriswest:
We also really want to see this in. We thought it was missing some key fields in the events however in order to generate good notification cards. 

@goldensyntax can you dig out what we thought was missing and add details or work to raise a PR into this one that adds them?

goldensyntax:
Hi @Andreybest, thanks for this. We think these fields could be added too. They're useful for building richer notifications, and most are already available on the `Action` object (`src/proxy/actions/Action.ts`) at dispatch time. Here's what we think would help:

| # | Field | Type | Usage / what it gives the notification |
|---|-------|------|----------------------------------------|
| 1 | `branch` | `string` | The branch the push targets (e.g. `main`, `feature/foo`). Shows what was pushed to — a push to `main` reads very differently from a feature branch. |
| 2 | `commitFrom` | `string` | Starting commit SHA of the push range. Enables linking straight to the diff. |
| 3 | `commitTo` | `string` | Ending commit SHA of the push range. Pairs with `commitFrom` so a reviewer can jump to the exact changes. |
| 4 | `stepResults` | `StepResult[]` | Per-step pass/fail from the automated chain checks (`{ name, passed, message }`). Shows which check fired — e.g. gitleaks vs commit-message policy. The core of a compliance notification. |
| 5 | `resolution` | `Resolution` (`'auto' \| 'manual'`) | How the push was resolved — auto (policy) or manual (human approval). |
| 6 | `reviewer` | `UserContext` (`{ username, email }`) | Who approved or rejected the push. |
| 7 | `rejectionReason` | `string` | The reason a human reviewer gave when rejecting. A push can pass all the automated checks and still be rejected by a reviewer, so this is separate from `stepResults`. |

FYI @kriswest

PR #1520, opened 2026-05-06
ci: refactor workflows for release branching

### LINKED ISSUE #1517: Tweak workflow files to work with GitLabFlow branching
In #1516, we're documenting how release branches work. Our current `release-drafter.yml` and `npm.yml` workflows were written assuming that we'd use a GitHubFlow (single `main` branch for latest release). In order to draft releases and publish to NPM correctly, we need to adapt these two workflows so that they can handle multiple drafts and updating old releases for patching.

`release-drafter.yml` should be updated to make multiple drafts, one for each updated release branch rather than a single draft from `main`. The workflow must allow to immediately publish the drafts if everything looks fine.

`npm.yml` should be updated to republish older versions that got patched. Right now, it simply publishes to the `latest` tag which gets installed by default when running `npx @finos/git-proxy`, which means that re-publishing old releases could catastrophically override the default install.

```yml
      - name: Check if pre-release and publish to NPM
        run: |
          VERSION=$(node -p "require('./package.json').version")
# Only checks if prerelease to decide if latest or rc tag
          if [[ "$VERSION" == *"-"* ]]; then
            echo "Publishing pre-release: $VERSION"
            npm publish --access=public --tag rc
          else
            echo "Publishing stable release: $VERSION"
            npm publish --access=public
          fi
```

**Additional context**
#1516 for documentation reference, already updated to reflect the desired behaviour
#1483 for more context on GitLabFlow

jescalada:
Closes #1517.

This refactors the release drafting and NPM publisher automation to work with release branches, as described in #1516. I've already updated #1516 with a step-by-step guide and a troubleshooting section based on my own testing.

Although it's a bit of a mess, I've replicated the release process in my [git-proxy fork](https://github.com/jescalada/git-proxy):
- [Test releases made from drafts](https://github.com/jescalada/git-proxy/releases)
- [NPM test package built from those releases](https://www.npmjs.com/package/@jescalada/git-proxy-deployment-testing?activeTab=versions)
- [Related PRs showcasing the branching flow](https://github.com/jescalada/git-proxy/pulls?q=is%3Apr+is%3Aclosed)


jescalada:
@kriswest Would be great to merge this in as we prepare for the 2.1 release! I've already made the [`release/2.0` branch](https://github.com/finos/git-proxy/tree/release/2.0) from the `2.0.0` bump PR. Since these CI workflow updates haven't been merged yet, nothing has executed when making the `release/2.0` branch (which is fine, since we decided to do everything manually one last time).

kriswest:
@re-vlad could you look at this one

re-vlad (changes requested):
Renaming .github/release-drafter.yml to .github/release-drafter-config.yml breaks the release-drafter job in pr-lint.yml: the with: block (lines 58–59) does not set config-name, so the action still looks for the default .github/release-drafter.yml and will likely fail at runtime (with message: "config file not found").

re-vlad (changes requested):
The job also keeps updating a draft release on main (commitish: main), which conflicts with the GitLabFlow model where drafts are created per release/* branch by the new .github/workflows/release-drafter.yml workflow. Autolabeling is now handled by auto-label.yml (`.github/workflows/auto-label.yml`), so this job looks redundant.

Suggested fix: remove the release-drafter job from pr-lint.yml and add edited to auto-label.yml pull_request_target.types

jescalada:
@re-vlad Thanks for the review! I forgot to remove the old release-drafter step from `pr-lint.yml` in this PR (I had already moved it into `release-drafter.yml`).

The point of `pr-lint.yml` is to just lint the PR title; and the new `release-drafter.yml` does exactly what the filename implies: drafting a release, rather than just being a template. The actual template/configuration is now appropriately named `release-drafter-config.yml`.

I think both of your comments stem from that oversight of mine.

Note that in my release tests, the [updated file is being used](https://github.com/jescalada/git-proxy/actions/runs/25305276224/workflow), that's why I didn't see any related failures when making PRs or releases.

jescalada:
@kriswest Ready for a final check!

jescalada:
@finos/git-proxy-maintainers Just need this merged to kick off the 2.1 release 🙂

.github/workflows/release-drafter.yml
```diff
@@ -0,0 +1,33 @@
...
+        with:
+          egress-policy: audit
+
+      - uses: release-drafter/release-drafter@563bf132657a13ded0b01fcb723c5a58cdd824e2
+        with:
+          # Target the branch that triggered this run
+          # Ex: "release/2.1" becomes the commitish for the draft
+          commitish: ${{ github.ref_name }}
+          repository: ${{ github.repository }}
+          config-name: release-drafter-config.yml
```
andypols, line 31: `commitish` controls the target branch of the generated release, but it does not by itself isolate drafts belonging to concurrent release branches

It's probably worth adding a  `filter-by-commitish: true`.  When set to true, it restricts the included commits to those that are descendants of the previous release's target_commitish (the commit or branch associated with the last release), rather than including all commits since the beginning of the repository or a fixed date.

andypols (changes requested):
LGTM - but think it's a good idea to add the filter-by-commitish setting.

.github/workflows/release-drafter.yml
jescalada: You're right, I also missed adding this when porting the PR over - I originally added it to `release-drafter-config.yml`: https://github.com/jescalada/git-proxy/pull/17 which is why the release notes were generated without issue on my end 🙂

jescalada:
@andypols Pardon the late update on this - should be good to go now!

andypols (approved):
LGTM

PR #1521, opened 2026-05-07
fix: refactor Dockerfile to speed up run times

jescalada:
Fixes `Dockerfile` slow runtimes (5-10 mins each) to speed up E2E test execution speeds.

Note that Docker Compose step execution times may vary considerably depending on whether we've executed the CI "recently" (perhaps within a few hours). Another factor might be Docker/NPM traffic spikes that make fetching slow.

I've seen [worst-case executions of around 15 mins](https://github.com/finos/git-proxy/actions/runs/25104668732/job/73562702163):

image

### Before

~7m without caching
image

~2m 10s when cached
image

### After

~4m 50s without caching (~30% faster)
image

~1m 20s when cached (~35% faster)
image


coopernetes (approved):
Simple enough optimization, LGTM. 

Something to also consider is leveraging a combination of GitHub Actions & Docker caching so that `node_modules` are also cached in the builder step.

```yaml
      - name: Build and Publish Docker Image
        if: github.repository_owner == 'finos'
        uses: docker/build-push-action@f9f3042f7e2789586610d6e8b85c8f03e5195baf # v7
        with:
          context: .
          file: Dockerfile
          push: true
          tags: ${{ steps.tags.outputs.tags }}
          provenance: true
          cache-from: type=gha
          cache-to: type=gha,mode=max
```


```Dockerfile
COPY package*.json ./
COPY tsconfig.json tsconfig.publish.json proxy.config.json config.schema.json test-e2e.proxy.config.json vite.config.ts index.html index.ts ./

# Cache hit on no changes to /root/.npm, miss if dependencies change
RUN --mount=type=cache,target=/root/.npm npm pkg delete scripts.prepare && npm ci --include=dev
```

PR #1529, opened 2026-05-10
feat: dependency-check-plugin-revision

RichardUri:
## Description

Revises and addresses review feedback on #799, adding dependency vulnerability scanner plugin.

The plugin is intended to intercept pushes and scans changed dependency files against the OWASP Vuln database. If a vulnerability is found at or above a severity threshold, the push will be held for review.

Made changes according to reviewer comments on original PR.
General changes:
- Severity threshold configured via `DEPENDENCY_VULN_THRESHOLD` environment variable (default: `HIGH`) rather than a top-level config schema entry
- Fixed `git diff` argument order (Previously showed removals instead of additions)
- Fixed `git show` using `HEAD` instead of the pushed commit SHA
- Fixed missing parent directory creation for nested file paths (e.g. `src/lib/foo.json`)
- `deleteTempRepo` now deletes the entire `.tempRepo/` directory instead of only the current push's subdirectory 
- Added exit code check for `dependency-check` to prevent silent false negatives on failure
- Replaced `spawnSync` for the clone operation with async `spawn` to avoid blocking
  the Node.js event loop
- Removed dead variables and unneeded aliases.
- Added Basic auth credential forwarding on the clone URL, similar one used in `pullRemote` processor
- Blocked message now includes CVE ID, severity, and description
- 'temporaryRepo.js' removed and is now directly included into the main file.
- Added `exports` entry and README documentation for the new plugin

## Related Issue

Revises #799 

### General

### Documentation

### Configuration

### Tests

- [ ] Tests have been added/updated for new functionality


kriswest (commented):
It seems odd to have a plug-in to a complete extra clone of the repository and compute its own diff in order to run scanning tools on it, when when the built-in scanners already do both of these things... Can this not be written to use the data from the existing clone and diff operations?

RichardUri:
@kriswest Yes, I thought it was strange in the original PR too. I believe it was structured this way because plugins are inserted at position 1 in the processor chain after parsePush so when the plugin executes, neither the bare clone nor the diff step exist yet. To reuse that data, I believe the plugin insertion point would need to move to later in the chain, after pullRemote and getDiff have run.  I can take a look at changing that.

plugins/git-proxy-plugin-samples/checkDependencyVuln.js
```diff
@@ -0,0 +1,246 @@
...
+
+        if (cloneResult.exitCode !== 0) {
+          step.setError(`Failed to clone repository for dependency scan: ${cloneResult.stderr}`);
+          action.addStep(step);
+          return action;
+        }
+
+        // Apply the pushed pack data to the local bare clone.
+        // req.body is the raw pack buffer, set by proxyFilter before the chain runs.
+        spawnSync('git', ['receive-pack', action.repoName], {
```
jescalada, line 119: We might want to check the result of `spawnSync` or wrap it in a try-catch. If unpacking silently fails then the subsequent diff would be empty and the plugin/processor would allow the push regardless of the actual contents.

plugins/git-proxy-plugin-samples/README.md
```diff
@@ -10,6 +10,43 @@ These plugins are maintained by the core GitProxy team. As a future roadmap item
...
+**Prerequisites**
+
+- The `dependency-check` CLI must be installed and available in `PATH`.
+- Run `dependency-check --updateonly` at least once after installation to populate the NVD database.
+  Repeat periodically to keep vulnerability data current (the plugin uses `--noupdate` on each scan
+  to avoid the 20-30 minute refresh overhead).
+
+**Configuration**
+
+Set the `DEPENDENCY_VULN_THRESHOLD` environment variable to control which severity levels trigger a block.
```
jescalada, line 30: The code below mentions that we should set the `dependencyVulnThreshold` on the `proxy.config.json`, which I think is the cleaner solution. Should we perhaps replace the environment variable with the config entry (`plugin` array) for this plugin?

jescalada (commented):
LGTM - although as Kris mentioned, this seems more appropriate (and useful for the wider community) as a default processor rather than a plugin since we're pulling/scanning the diff. Is there a reason to keep it around as a plugin?

PR #1542, opened 2026-05-20
feat: expose service module through package exports (#1539)

### LINKED ISSUE #1539: Expose 'service' module as a public export for dashboard also slightly changed config/index.js configurations
Is your feature request related to a problem? Please describe.

The current package exports expose ./ui, but the dashboard bootstrap/runtime implementation appears to reside in src/service/index.js.

The Service module is responsible for:

Creating the Express application
Registering routes
Serving the compiled frontend build
Listening on GIT_PROXY_UI_PORT

Since this module is not exported as part of the public package API, consumers cannot programmatically start the dashboard service independently.

This limitation became more visible in v2 because the package now uses Node.js "exports" restrictions. Unlike v1, consumers can no longer directly access arbitrary internal files from node_modules.

For example, consumers previously relied on internal imports such as:

require('@finos/git-proxy/src/config/file')

However, these paths are now blocked unless explicitly exported.

Configuration access can be resolved cleanly by re-exporting the required APIs from:

src/config/index.ts

However, the dashboard/service runtime itself is still inaccessible because src/service/index.js is not part of the public exports.

Describe the solution you'd like

Expose the service module as part of the public package exports.
This would enable programmatic dashboard startup, for example:

const service = require('@finos/git-proxy/service');

await service.start();

Additionally, expose configuration helpers through the public ./config entrypoint instead of requiring consumers to depend on internal file paths.

Describe alternatives you've considered

Current workaround approaches rely on importing internal package paths, such as:

require('@finos/git-proxy/src/config/file')
require('@finos/git-proxy/src/config/env')

However, these paths are incompatible with Node.js package exports restrictions introduced in v2 and tightly couple consumers to the internal folder structure.

Re-exporting the required APIs through ./config provides a cleaner and more stable public interface.

Additional context

Based on the current source structure:

src/ui — frontend source assets/components
src/service/index.js — serves the compiled frontend build and starts the dashboard server
src/config/file.ts — configuration file helpers
src/config/env.ts — environment/server configuration

rajeevr-g:
## Description

Exposes the service module through package exports and updates configuration exports to support service module accessibility.

General changes:

* Added package exports entry for the service module
* Updated `src/config/index.ts` to expose the required configuration exports
* Ensured service module accessibility through package-level exports
* Verified project build and TypeScript publish build complete successfully
* Validated linting, formatting, and type checks pass successfully

## Related Issue

Resolves #1539

### General

* [x] I have read the [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines
* [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) format
* [ ] I have a FINOS CLA on file

### Documentation

* [ ] Documentation has been added/updated for any new features

### Configuration

* [ ] If configuration schema (`config.schema.json`) was modified:

  * [ ] TypeScript types regenerated (`npm run generate-config-types`)
  * [ ] Schema reference docs regenerated (`npm run gen-schema-doc`)

### Tests

* [x] Unit tests pass (`npm test`)
* [x] Linting and formatting pass (`npm run lint` and `npm run format:check`)
* [x] Type checks pass (`npm run check-types`)

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -27,7 +27,6 @@ import { validateConfig } from './validators';
 import { handleErrorAndLog, handleErrorAndThrow } from '../utils/errors';
 
 export { setConfigFile, getConfigFile, validate } from './file';
-export { serverConfig } from './env';
 
 // Deprecated compatibility fields are still optional because the defaults do not set them.
 type OptionalTopLevelConfigKey = 'proxyUrl' | 'sslCertPemPath' | 'sslKeyPemPath';
@@ -225,6 +224,11 @@ export const getUpstreamProxyConfig = () => {
 export const getAuthorisedList = () => {
   const config = loadFullConfiguration();
   return config.authorisedList;
+};
+
+// Get GIT_PROXY_UI_PORT
+export const getUIPort = (): number => {
+  return Number(serverConfig.GIT_PROXY_UI_PORT);
 };
 
 // Gets a list of authorised repositories
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -22,8 +22,12 @@ import { ConfigLoader } from './ConfigLoader';
 import { Configuration } from './types';
 import { serverConfig } from './env';
 import { getConfigFile } from './file';
+
 import { validateConfig } from './validators';
 import { handleErrorAndLog, handleErrorAndThrow } from '../utils/errors';
+
+export { setConfigFile, getConfigFile, validate } from './file';
+export { serverConfig } from './env';
 
 // Cache for current configuration
 let _currentConfig: GitProxyConfig | null = null;
--- a/package.json
+++ b/package.json
@@ -44,6 +44,11 @@
       "types": "./dist/src/utils/errors.d.ts",
       "import": "./dist/src/utils/errors.js",
       "require": "./dist/src/utils/errors.js"
+    },
+    "./service": {
+      "types": "./dist/src/service/index.d.ts",
+      "import": "./dist/src/service/index.js",
+      "require": "./dist/src/service/index.js"
     }
   },
   "scripts": {


re-vlad (approved):

re-vlad:
Hi @andypols @kriswest @jescalada, @Andreybest, dear @git-proxy-maintainers,

Could you please prioritize a review of this PR [#1542 “feat: expose service module through package exports (#1539)”](https://github.com/finos/git-proxy/pull/1542)?

This change is critical for us because we need the service module to be available through package exports in order to validate our migration path to the 2.0.0 release.

If anything needs adjustment from our side, we can respond quickly.

Thank you for your help and for maintaining the project,
Vlad (re-vlad)

Andreybest (approved):

jescalada (approved):
LGTM, thanks for the contribution @rajeevr-g! It'll need an extra review, however, as I fixed merge conflicts in the latest commit 👍🏼

rajeevr-g:
Hi @andypols @kriswest, dear @git-proxy-maintainers,

Could you please help prioritize the review of PR #1542:
https://github.com/finos/git-proxy/pull/1542

This change is important for us because we need the service module to be exposed through package exports to validate our migration path for the 2.0.0 release and ensure our integration works successfully.

If any adjustments or updates are needed from our side, we’ll be happy to address them promptly.

Thank you for your time and for maintaining the project.

Best regards,
Rajeev

kriswest (commented):
I don't have a problem with this being exposed, however, it could be handled in a way that's more consistent with the exposure of other parts of the config, via a get* function. I don't like the fact that this part of the config sets defaults differently and that we can't set many of these settings from a config file (with the exception of GIT_PROXY_COOKIE_SECRET which can be set in config and overridden by an ENV var).

Personally, I think this should be refactored and made consistent with the rest of the config handling and worry that directly exporting it will lock in the current, inconsistent design.

@jescalada could you take another look? Is there an advantage to us *only* setting some of these thrings through env vars or using the defaults? Why not have config variables and defaults (set via the same approach as all others) for some of these critical settings?

src/config/index.ts
```diff
@@ -22,9 +22,13 @@ import { ConfigLoader } from './ConfigLoader';
 import { Configuration } from './types';
 import { serverConfig } from './env';
 import { getConfigFile } from './file';
+
 import { validateConfig } from './validators';
 import { handleErrorAndLog, handleErrorAndThrow } from '../utils/errors';
 
+export { setConfigFile, getConfigFile, validate } from './file';
+export { serverConfig } from './env';
```
kriswest: As discussed at #1552 this export of env.ts should be removed from teh PR - if you need access to these variables then please add a get* function instead (see examples of other config get functions in this file) and consider taking on a quick refactor of those settings to allow them to be set via the config file as well/instead.

kriswest (changes requested):
To avoid locking in the technical debt relating to env.ts / serverConfig, it should not be exported but a get function added to allow access to those settings. Even better would be a quick refactor to allow those settings to come from config in addition to env vars and to move their defaults into proxy.config.json.

GIT_PROXY_COOKIE_SECRET is the one variable that is currently settable from both config file and environment variable and can be used as an example: 

see: https://github.com/finos/git-proxy/blob/ffd6937419b1d3de718c60ec26604f9f7541fffa/src/config/index.ts#L187-L190

and: https://github.com/finos/git-proxy/blob/ffd6937419b1d3de718c60ec26604f9f7541fffa/src/config/index.ts#L301-L309

I'm fine with the rest of this PR!

kriswest:
I've created issue #1553 to describe whats needed to fix the server config env variables properly.

kriswest (approved):
LGTM, as far as it goes.

PR #1543, opened 2026-05-21
feat(migrate): DB tooling for v1.19.2 -> v2.0.0 (URLs + emails + ACL audit)

### LINKED ISSUE #1535: Proposal: Automated MongoDB Migration Script from v1.19.2 to v2.0.0 upgrade
GitProxy v2.0.0 has breaking schema changes from v1.19.2 that require data migration, but no built-in migration script exists. Critical changes include:
     - Repository URLs must change from https://github.com/org/repo → https://github.com/org/repo.git
     - User permissions (canPush, canAuthorise) now require email addresses instead of usernames

To add a dedicated MongoDB migration script that:
     1. Takes a database backup first;
     2. Converts repository URLs to new format;
     3. Adds missing fields with default values;
     4. (Optional) Provides a CSV import option for username→email mapping

Manual database updates are error-prone and not recommended for production. Without a migration script, users will face data inconsistency issues when upgrading.

**Additional context**
no additional context

re-vlad:
## Description

Upgrade to Git-Proxy v2.0.0 requires database preparation for breaking behavior around exact repo URL lookup (need .git on repos.url) and v2 push authorization based on committer email + User.username ACL entries. 
This PR adds operator-run migration scripts with dry-run by default, explicit apply, structured reports, and separate backups so upgrades are safer and repeatable (aligned with [issue #1535](https://github.com/finos/git-proxy/issues/1535)).

Phase 1 — **repo URL normalization**: append .git to repos.url where missing (idempotent) via migrate-urls.js + backup-urls.js.
Phase 2 — **users & ACL normalization**:
- migrate-users.js always runs email audit + ACL orphan audit (entries in canPush / canAuthorise that don’t resolve to any User.username; no silent ACL rewrite).
- optional email updates via --apply --csv username,email (lib/csv.js, apply-user-emails.js with post-apply uniqueness simulation).
- backup-users.js writes backup-users-*.json (full snapshot, passwords excluded) and users-email-*.csv template for admin edits before apply.
- docs: The Migration-guide-v1-to-v2.md updated for both phases + csv constraints.

## Related Issue

Resolves [#1535](https://github.com/finos/git-proxy/issues/1535)

### General

### Documentation

### Configuration

### Tests

**Note**: Migration scripts are one-off operator tooling, not part of the main app bundle. Their unit tests live under scripts/migrate/test/ and are not included in the default npm test run (which targets ./test). Run them separately with:
```bash
npm run test:migrate
```


scripts/migrate/Migration-guide-v1-to-v2.md
```diff
@@ -0,0 +1,94 @@
...
+npm run migrate:users -- --apply --csv ./map.csv
+```
+
+---
+
+## Phase 1 — Repo URL normalization (append `.git` where missing)
+
+**Goal:** every `repos.url` that v2 will match must include the `.git` suffix where it is missing.
+
+**Why:** v2 resolves repos by **exact** `url`; v1 often relied on `name`. Clients typically send URLs with `.git`; rows without `.git` stop matching.
```
jescalada: If the "client" you're referring to is the UI, this is inaccurate: the UI enforces the `.git` ending when creating new repositories. I'd edit this for accuracy/clarity:

```suggestion
**Why:** v2 resolves repos by **exact** `url`; v1 often relied on `name`. Rows without `.git` no longer match when calling `getRepoByUrl`, which makes processors such as `checkRepoInAuthorisedList` behave unexpectedly.
```

scripts/migrate/lib/analyze-users.js
```diff
@@ -0,0 +1,136 @@
...
+    duplicate: issues.filter((x) => x.status === 'duplicate').length,
+    invalidFormat: issues.filter((x) => x.status === 'invalid-format').length,
+  };
+
+  const report = {
+    totalUsers: users.length,
+    counts,
+    duplicateGroups,
+    issues,
+    blockingIssueCount: counts.missing + counts.blank + counts.duplicate,
```
jescalada: Shouldn't we be counting emails with invalid formatting `counts.invalidFormat` here as well?

scripts/migrate/test/analyze-users.test.ts
```diff
@@ -0,0 +1,51 @@
...
+      { _id: 'u1', username: 'noEmail' }, // missing field
+      { _id: 'u2', username: 'blankEmail', email: '   ' },
+      { _id: 'u3', username: 'badEmail', email: 'not-an-email' },
+    ];
+    const { report } = buildUsersEmailReport(users);
+    const byId = new Map(report.issues.map((i: any) => [i.userId, i]));
+    expect(byId.get('u1').status).toBe('missing');
+    expect(byId.get('u2').status).toBe('blank');
+    expect(byId.get('u3').status).toBe('invalid-format');
+    expect(report.blockingIssueCount).toBe(2); // missing + blank
```
jescalada: Same as the previous comment: any specific reason why emails with invalid formatting are not blocking? The script would just proceed with the update even if invalid emails were present.

scripts/migrate/lib/analyze-urls.js
```diff
@@ -0,0 +1,83 @@
...
+  const report = {
+    totalRepos: repos.length,
+    reposNeedingUpdate: 0,
+    reposAlreadyFixed: 0,
+    changes: [],
+  };
+
+  for (const repo of repos) {
+    const currentUrl = (repo.url || '').trim();
+    const needsUpdate = !currentUrl.endsWith('.git');
```
jescalada, line 67: Wouldn't this create invalid URLs if the `url` field wasn't formatted properly before? Some examples:

- `url: ""` would turn into `url: ".git"`
- `url: "https://github.com/finos/git-proxy/` would turn into `url: "https://github.com/finos/git-proxy/.git`

At least we should report these, so admins can fix them manually.

scripts/migrate/migrate-urls.js
```diff
@@ -0,0 +1,124 @@
...
+        console.log(`     node scripts/migrate/migrate-urls.js --apply`);
+      }
+    } else {
+      console.log('\n=== APPLY PHASE ===');
+      let reposUpdated = 0;
+      let errors = 0;
+
+      for (const change of report.changes) {
+        try {
+          const success = await updateRepoUrl(
```
jescalada: Just a heads up: `updateRepoUrl` is creating a new `MongoClient` for each repo to update. We probably want to have a single client or do bulk writes instead.

scripts/migrate/lib/common.js
```diff
@@ -0,0 +1,68 @@
...
+    );
+
+    return result.modifiedCount === 1;
+  } finally {
+    await client.close();
+  }
+}
+
+async function countReposWithoutGit(mongoUri, dbName) {
+  const client = new MongoClient(mongoUri);
```
jescalada: This function might also benefit from having a shared `MongoClient`

scripts/migrate/lib/common.js
```diff
@@ -0,0 +1,68 @@
...
+
+  try {
+    await client.connect();
+    const db = client.db(dbName);
+    const reposCollection = db.collection('repos');
+
+    return await reposCollection.countDocuments({
+      url: { $not: /\.git$/ },
```
jescalada: Wouldn't this match empty strings/null URLs? These would get counted but not actually migrated since we're currently ignoring empty URLs in `buildUrlNormalizationReport`.

jescalada (changes requested):

scripts/migrate/lib/analyze-urls.js
re-vlad: Agreed, thank you @jescalada, the initial implementation could produce invalid values (e.g. "" → ".git" or trailing / → "/.git"). This PR now normalizes only safe cases (trim + strip trailing /) and reports blank/non-http(s) URLs as issues (with url-issues-*.csv) for manual admin fixes; it also exits non‑zero while issues remain so they can’t be missed.

scripts/migrate/migrate-urls.js
re-vlad: Agreed: now we reuse a single Mongo сlient when 'apply' to avoid reconnect per repo

Andreybest (commented):
Looks ok from my side, please check @jescalada comments.
Only 2 things from my side to make it perfect:
1. TypeScript in migration scripts would be appreciated.
2. Tests that are run on virtual mongo would show more reliability by using [mongodb-memory-server](https://github.com/typegoose/mongodb-memory-server) or it's derivatives.

scripts/migrate/lib/common.js
re-vlad: Absolutely, now the function accept already prepared collection and not opened the connection internally

scripts/migrate/lib/common.js
re-vlad: The criteria: { url: { $not: /\.git$/ } } also matches empty/null and other non-.git URLs.
That's intentional. remaining is a post-apply DB check (how many repos.url values still don't end with .git), not a count of failed migration updates. Blank/unsupported URLs are skipped by normalizeRepoUrl and go to 'issues', not `report.changes`, but they still belong in this total because they need manual fixes.
We already report them as URL 'issues' and fail the run when urlIssues > 0. I updated the verify warning to note that remaining includes blank/invalid URLs listed under URL issues.

re-vlad:
> 1. TypeScript in migration scripts would be appreciated.

Agreed. I'll track a follow-up to convert scripts/migrate to TS for consistency with the rest of the repo.

scripts/migrate/Migration-guide-v1-to-v2.md
re-vlad: Good point, this is definately my miswording - by "clients" I meant incoming git HTTP traffic (which includes .git in the path), not the admin UI, which already enforces .git on create. I've updated the 'Why..' section to describe the actual failure mode: exact `getRepoByUrl` lookup vs legacy rows without `.git`, and the impact on processors like `checkRepoInAuthorisedList`

scripts/migrate/lib/analyze-users.js
re-vlad: Good catch: invalid-format is already in issues and `counts.invalidFormat`, but it was omitted from `blockingIssueCount`. That meant `dry-run` could exit 0 while the DB still had malformed emails, which is inconsistent with CSV validation. Now the`counts.invalidFormat` included into the `blockingIssueCount`.

scripts/migrate/test/analyze-users.test.ts
re-vlad: Sure, changed: missing + blank + invalid-format

re-vlad:
@jescalada @andypols @dcoric  please re-review and resolve comments

scripts/migrate/lib/common.js
jescalada: Makes sense if these get fixed manually down the line 👍🏼

jescalada (approved):
@re-vlad LGTM, thanks for the contribution! 🚀

PR #1554, opened 2026-06-02
feat(config): allow server settings via config file with env precedence

### LINKED ISSUE #1553: Allow server details to be set through configuration in addition to environment variables

The key server settings below are not configurable via the git proxy config file, they a can be set via environment variables and defaults for them are set in source code, rather than in the same place as all other config settings:
https://github.com/finos/git-proxy/blob/ffd6937419b1d3de718c60ec26604f9f7541fffa/src/config/env.ts#L19-L27

These are likely the oldest settings for git proxy and the difference in how they are configured is a legacy/tech debt. The difference in how they are set is confusing and creates complexity for those running Git Proxy installations for no real benefit.

These settings should be added to the config schema and read from the default and user config files for consistency with the rest of the configuration.

`GIT_PROXY_COOKIE_SECRET` is the one variable that is currently settable from both config file and environment variable and can be used as an example: 

see: https://github.com/finos/git-proxy/blob/ffd6937419b1d3de718c60ec26604f9f7541fffa/src/config/index.ts#L187-L190

and: https://github.com/finos/git-proxy/blob/ffd6937419b1d3de718c60ec26604f9f7541fffa/src/config/index.ts#L301-L309

Usage of the config settings will need to shift to using `get*` functions rather than direct import of the variables from [env.ts](https://github.com/finos/git-proxy/blob/main/src/config/env.ts). I.e. replace:
https://github.com/finos/git-proxy/blob/ffd6937419b1d3de718c60ec26604f9f7541fffa/src/service/index.ts#L36

https://github.com/finos/git-proxy/blob/ffd6937419b1d3de718c60ec26604f9f7541fffa/src/proxy/index.ts#L35-L36

https://github.com/finos/git-proxy/blob/ffd6937419b1d3de718c60ec26604f9f7541fffa/src/service/urls.ts#L22

We  could deprecate the environment variables or maintain support for them with primacy over the config file (as is the case for GIT_PROXY_COOKIE_SECRET.

Continue on with most config going through the config file, but key server settings only being settable via environment variables.

**Additional context**

Identified as necessary when reviewing https://github.com/finos/git-proxy/pull/1542, which sought to export the server config and would have locked in this pattern by enabling external use.

akagifreeez:
## Description

The server port, HTTPS server port, UI host, UI port and HTTPS UI port were only
configurable via environment variables, with defaults hard-coded in `src/config/env.ts` —
unlike every other setting, which lives in the config schema/file. This legacy split adds
needless complexity for operators.

This moves the five settings into the config schema (`config.schema.json` /
`proxy.config.json`), mirroring how `GIT_PROXY_COOKIE_SECRET` already works:

- **Precedence:** environment variable (when set) → config file → built-in default.
- `src/config/env.ts` now exposes the raw environment values (`undefined` when unset) so the
  config file can supply the default.
- New getters `getServerPort` / `getHttpsServerPort` / `getUIHost` / `getHttpsUIPort` (and
  `getUIPort` now reads the merged config); the proxy, service, URL helpers and the OIDC auth
  route read these lazily instead of at module load.
- Also fixes the OIDC callback redirect, which previously fell back to UI port `3000` instead
  of the configured UI port.

Config-file properties added: `serverPort`, `httpsServerPort`, `uiHost`, `uiPort`, `httpsUiPort`.

## Related Issue

Resolves #1553

### General

- [x] I have a [FINOS CLA](https://finosfoundation.atlassian.net/wiki/spaces/FINOS/pages/75530375/Contribution+Compliance+Requirements) on file (will sign via EasyCLA on this PR)

### Documentation

- [x] Documentation has been added/updated (config overview note + regenerated schema reference)

### Configuration

### Tests

- [x] Tests have been added/updated for new functionality (default / config-file / env-precedence for the new settings)
- [x] Unit tests pass (`npm test`) — 808 passed


kriswest (approved):
LGTM

PR #1582, opened 2026-06-08
feat: data migration from mongo/fs to the PostgreSQL sink

### LINKED ISSUE #1557: PostgreSQL sink: data migration from mongo/fs to postgres
Follow-up to #1497 (PostgreSQL sink backend, added in #1532).

The postgres sink bootstraps an empty schema but provides no path to move existing data from the `fs` or `mongo` backends.

## Decision needed first

`users._id` and `repos._id` are `UUID` columns (`gen_random_uuid()`), but mongo `_id`s are 24-char ObjectId hex strings, which are not valid UUIDs. The migration must either:

- regenerate IDs on import (repos are looked up by URL, users by username, and pushes reference repos by URL, so this is mostly safe — but `/repo/:id` links would change), or
- change those columns to `TEXT` to preserve the originals.

## Scope

A documented, scripted migration (`fs` / `mongo` → `postgres`). The adapter's idempotent upserts (`writeAudit`, `createRepo`, `createUser` / `updateUser`) can be reused by the importer.

dcoric:
## Description

> Draft, stacked on #1532 (PostgreSQL sink). The diff below currently includes the #1532 changes. Once #1532 merges, this branch will be rebased onto `main` so only the migration changes remain, and the draft flag will be removed.

Adds a data-migration utility to copy `users`, `repos` and `pushes` from a `mongo` or `fs` (NeDB) backend into PostgreSQL.

- `src/db/postgres/migrate.ts` orchestrates the import. It is idempotent: users and repos that already exist (matched by username/email and URL) are skipped, and pushes are upserted by id, so the import is safe to re-run.
- Source readers `migrateMongoSource.ts` (MongoDB driver) and `migrateFileSource.ts` (NeDB) read from the legacy backend with their own connection while the importer writes through the active PostgreSQL sink, so the two connections never clash.
- Record `_id`s are not carried over; PostgreSQL assigns fresh UUIDs. Push ids (text) are preserved by the upsert.
- Invoked via `npm run migrate:postgres -- --from <mongo|fs> [--mongoUrl ... | --dataDir ...]`.
- Unit tests cover the orchestrator and both sources; a gated integration test covers the fs to postgres path end to end.

Note: the integration test exercises the fs to postgres path only, because the PostgreSQL CI lane does not run a Mongo service. The mongo source is unit-tested with a mocked driver. Wiring a Mongo service into that lane for full mongo-to-postgres integration coverage is a reasonable follow-up.

## Related Issue

Resolves #1557

### General

### Documentation

### Configuration

- [ ] If configuration schema (`config.schema.json`) was modified: (not modified in this PR)

### Tests


kriswest:
If you start working these PRs from FINOS repo branches (which you should be able to do as a member of the (trusted) contributors team) then you could target these PRs at the the other PR branch so we only see the diff for the additions... Once the the other PR merges and its branch is deleted github switches the target to main automatically.

If you are using the DB adaptor to perform the migration (only took a brief look so far), can this not be generalised to handle migration to/from any adaptor? That would be preferable to a postgres as destination specific migration tool (I'm not convinced thats the right direction :-p ).

scripts/migrate-to-postgres.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  const source = await buildSource();
+  try {
+    const summary = await migrate(source, postgres, { log: (message) => console.log(message) });
+    console.log('Migration complete:');
+    console.log(`  users:  ${summary.users.imported} imported, ${summary.users.skipped} skipped`);
+    console.log(`  repos:  ${summary.repos.imported} imported, ${summary.repos.skipped} skipped`);
+    console.log(`  pushes: ${summary.pushes.imported} imported`);
+  } finally {
+    await source.close();
```
fabiovincenzi, line 77: Worth closing the pool too or process.exit(0) after the summary.

fabiovincenzi:
Nice!

src/db/postgres/migrateMongoSource.ts
```diff
@@ -0,0 +1,52 @@
...
+export const createMongoSource = async (
+  connectionString: string,
+  options: MongoClientOptions = {},
+): Promise<MigrationSource> => {
+  const client = new MongoClient(connectionString, options);
+  await client.connect();
+  const db = client.db();
+
+  const readAll = async <T>(collection: string, proto: object): Promise<T[]> => {
+    const docs = await db.collection(collection).find().toArray();
```
jescalada, line 40: I'm wondering if there's a way to batch the migration process in case we have lots of pushes (from the `readAll` call for `getPushes`  below) - I figured this might be a problem when handling 10k+ rows which might be the case for production setups looking to migrate.

Maybe we could batch this in blocks of 100 or so, and add logging so the migration process is visible and easily recoverable for large batch numbers?

Just a suggestion, and I hope it doesn't add too much extra wiring to this PR!

src/db/postgres/migrateFileSource.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Build a read-only {@link MigrationSource} backed by the NeDB datastores the
+ * `fs` sink writes. `dataDir` defaults to the location the sink uses
+ * (`./.data/db`). Record `_id`s are ignored by the Postgres writers, which
+ * assign fresh UUIDs.
+ */
+export const createFileSource = (dataDir: string = DEFAULT_DATA_DIR): MigrationSource => {
+  const load = (file: string): Datastore =>
+    new Datastore({ filename: path.join(dataDir, file), autoload: true });
```
jescalada: Might want to check that the file is actually empty versus just invalid (or non-existent) file - and output to the user/fail fast 👍🏼

test/db/postgres/migrate.test.ts
```diff
@@ -0,0 +1,141 @@
...
+  close: vi.fn().mockResolvedValue(undefined),
+  ...over,
+});
+
+const makeDestination = (over: Partial<MigrationDestination> = {}): MigrationDestination => ({
+  findUser: vi.fn().mockResolvedValue(null),
+  findUserByEmail: vi.fn().mockResolvedValue(null),
+  createUser: vi.fn().mockResolvedValue(undefined),
+  getRepoByUrl: vi.fn().mockResolvedValue(null),
+  createRepo: vi.fn().mockResolvedValue(undefined),
```
jescalada, line 40: Wondering if we need `addUserCanPush` and `addUserCanAuthorise` here - does `createRepo` admit adding these values directly?

test/db/postgres/migrate.test.ts
```diff
@@ -0,0 +1,141 @@
...
+  getRepos: vi.fn().mockResolvedValue([]),
+  getPushes: vi.fn().mockResolvedValue([]),
+  close: vi.fn().mockResolvedValue(undefined),
+  ...over,
+});
+
+const makeDestination = (over: Partial<MigrationDestination> = {}): MigrationDestination => ({
+  findUser: vi.fn().mockResolvedValue(null),
+  findUserByEmail: vi.fn().mockResolvedValue(null),
+  createUser: vi.fn().mockResolvedValue(undefined),
```
jescalada, line 38: Similarly to the last comment, I'm wondering if passwords are handled correctly - AFAIK `createUser` hashes the password input already, so we would end up with invalid login credentials 🤔

jescalada (commented):
Thanks for this! Just a few comments on this one - and Fabio's comment to fix up 👍🏼

src/db/postgres/migrateMongoSource.ts
dcoric: Done - pushes now stream in batches (default 500, configurable via pushBatchSize). The mongo source iterates a cursor with batchSize() instead of one toArray(), so the table never sits in memory whole; the NeDB source chunks to the same contract since it is in-memory regardless. Covered with a batch-splitting test.

src/db/postgres/migrateFileSource.ts
dcoric: Added - the file source now fails fast with a clear error when the data directory does not exist or contains none of the fs sink datastores, and it loads explicitly (no autoload) so a corrupt datastore surfaces as 'Failed to load ...' instead of reading as empty. Tests cover all three cases.

test/db/postgres/migrate.test.ts
dcoric: createRepo persists the whole users permission map itself (it lands in the repos JSONB column), so no separate addUserCanPush / addUserCanAuthorise calls are needed. Added a test asserting the map arrives at createRepo intact.

scripts/migrate-to-postgres.ts
dcoric: Done - the CLI now closes the destination pool in the finally block (alongside source.close()), so the process exits cleanly after the summary instead of hanging on open handles.

test/db/postgres/migrate.test.ts
dcoric: Good instinct, but it checks out: hashing lives in the service-level wrapper (src/db/index.ts createUser calls bcrypt.hash), while the migration writes through the raw postgres adapter whose createUser stores the supplied value verbatim. Source passwords are already bcrypt hashes and are copied byte for byte, so logins survive. Added a test pinning that behaviour so a future refactor cannot silently reintroduce hashing on this path.

jescalada (approved):
LGTM, just needs conficts fixed 🚀

PR #1584, opened 2026-06-10
feat: configurable PostgreSQL connection (discrete fields, TLS, pool)

### LINKED ISSUE #1560: PostgreSQL sink: connection configuration (split env vars, pool & TLS options)
Follow-up to #1497 / #1532.

Only a single `connectionString` (or `GIT_PROXY_POSTGRES_CONNECTION_STRING`) is supported, and the pool is created with all defaults (`new Pool({ connectionString })`).

## Scope

- Support split `PGHOST` / `PGPORT` / `PGUSER` / `PGPASSWORD` / `PGDATABASE`.
- Expose pool sizing / timeout configuration.
- Provide structured TLS/ssl options (parity with the mongo backend's `options` object), instead of requiring `sslmode` to be embedded in the connection string.

dcoric:
## Description

> Extension of #1532 (the PostgreSQL sink), stacked on it. The diff below currently includes the #1532 changes. Once #1532 merges, this branch will be rebased onto `main` so only the connection-configuration changes remain, and the draft flag will be removed.

Makes the PostgreSQL sink's connection fully configurable, bringing it closer to parity with the mongo backend.

- Discrete connection fields `host`, `port`, `user`, `password`, `database`, used when `connectionString` is not set. If those are not set either, the standard `PGHOST` / `PGPORT` / `PGUSER` / `PGPASSWORD` / `PGDATABASE` environment variables are read by the client.
- `ssl`: `true` for TLS with default certificate verification, or an object of TLS options (`rejectUnauthorized`, `ca`, `cert`, `key`, ...).
- `pool`: connection pool tuning (`max`, `idleTimeoutMillis`, `connectionTimeoutMillis`).

Precedence: `connectionString` (the config field, then `GIT_PROXY_POSTGRES_CONNECTION_STRING`) wins; otherwise the discrete fields; otherwise the `PG*` environment variables. `ssl` and `pool` apply in all cases. If none of these resolve to a connection, GitProxy refuses to start rather than silently defaulting to `localhost`.

Unit tests cover the precedence, discrete fields, env fallback, `ssl` (boolean and object) and `pool` options.

## Related Issue

Resolves #1560

### General

### Documentation

### Configuration

  - [ ] Schema reference docs regenerated (`npm run gen-schema-doc` requires the external `generate-schema-doc` tool; needs running in an environment that has it)

### Tests


src/db/postgres/helper.ts
```diff
@@ -0,0 +1,214 @@
...
+let _bootstrapPromise: Promise<void> | null = null;
+
+/**
+ * True when some Postgres connection is configured: an explicit connection
+ * string, the discrete `host` field, or the standard `PGHOST` environment
+ * variable (which implies the `PG*` family is in use). Used to refuse startup
+ * loudly rather than silently defaulting to `localhost`.
+ */
+const hasConnectionConfig = (db: DatabaseConfig): boolean =>
+  Boolean(db.connectionString || db.host || process.env.PGHOST);
```
fabiovincenzi: This only checks `PGHOST`, but the message says "or the PG* env variables". A setup with PGUSER/PGDATABASE and no PGHOST is valid but gets refused. Could the check also look at PGUSER/PGDATABASE/PGHOSTADDR so only the truly-empty case is blocked?

src/db/postgres/helper.ts
```diff
@@ -0,0 +1,214 @@
...
+
+/**
+ * Build a `pg` PoolConfig from the resolved database config. A connection
+ * string (already env-resolved by `getDatabase`) takes precedence; otherwise
+ * the discrete fields are used. When neither is set, `pg` reads the `PG*`
+ * environment variables itself.
+ */
+const buildPoolConfig = (db: DatabaseConfig): PoolConfig => {
+  const config: PoolConfig = {};
+  if (db.connectionString) {
```
fabiovincenzi, line 54: When both `connectionString` and the discrete fields are set, the discrete fields are silently ignored. It's documented as the intended precedence, but it's an easy operational trap (e.g. someone changes host to repoint the DB, nothing happens because an old connectionString is still winning with no signal). Worth a `console.warn` here noting the discrete fields are being ignored?

jescalada (approved):
LGTM after fixing up Fabio's comments 👍🏼

src/db/postgres/helper.ts
dcoric: Good catch - fixed. The check now accepts any of PGHOST, PGHOSTADDR, PGUSER or PGDATABASE, so a PGUSER/PGDATABASE-only setup starts and only the truly-empty case is refused. Covered with a test.

src/db/postgres/helper.ts
dcoric: Added - a console.warn now fires when a connectionString is set alongside discrete fields, so a stale connection string cannot silently shadow a host change. Tested both ways (warns when both are set, quiet when only the connection string is used).

PR #1587, opened 2026-06-11
fix(config): recover cached git sources on detached head

dcoric:
## What
When a git-backed config source (`configurationSources`) pins a `branch`, the refresh cycle could fail: a plain `git pull` on the cached clone errors once the checkout is on a detached HEAD or the remote branch has moved. The cached repo then never recovers without manual deletion.

This change makes the refresh path branch-aware:
- the git env (`GIT_TERMINAL_PROMPT=false`, optional `GIT_SSH_COMMAND`) is now shared by clone, fetch, and pull, where previously only the initial clone used it — so authenticated refreshes work too
- with a pinned `branch`: `git fetch --all --prune`, then checkout, then `git pull --ff-only origin <branch>`
- without a pinned branch: plain `git pull` as before

## Why a separate PR
Split out of #1532 per review feedback there — it's a standalone ConfigLoader fix, not part of the postgres sink.

## Testing
`test/ConfigLoader.test.ts` gains coverage that a cached repository left on a detached HEAD (and on an unborn default branch) is recovered on the next refresh.


src/config/ConfigLoader.ts
```diff
@@ -336,6 +346,17 @@ export class ConfigLoader extends EventEmitter {
       } catch (error: unknown) {
         handleErrorAndThrow(error, `Failed to checkout branch ${source.branch}`);
       }
+
+      console.log(`Pulling latest changes from ${source.branch}`);
+      try {
+        await execFileAsync('git', ['pull', '--ff-only', 'origin', source.branch], {
```
jescalada, line 352: Might want to double-check if `--ff-only` is the right move here if our intention is to fetch once and then store in the cache... Perhaps we need some way to recover from force pushes (maybe `git reset --hard origin/main`)?

In fact, I think we could do without the `git pull` entirely and just make the whole thing fix itself on a bad pull (that rewrites history):

```ts
await execFileAsync('git', ['fetch', '--all', '--prune'], { cwd: repoDir, env: gitEnv });
await execFileAsync('git', ['checkout', '-f', '-B', source.branch, `origin/${source.branch}`], { cwd: repoDir, env: gitEnv });
```

What's the logic behind using `git fetch` vs `git pull`? I noticed we do both throughout the `loadFromGit` function but it's not clear to me why we choose one over the other.

jescalada (changes requested):
Looks good so far, just wondering about the logic behind choosing `git pull` over `git fetch` in the code - and if there isn't a way to simplify the `loadFromGit` function 🤔 

Also, the `--ff-only` flag might continuously throw an error if the repo has been force-pushed: `fatal: Not possible to fast-forward, aborting.`. The only way to break out would be to delete the cache manually.

PR #1581, opened 2026-06-08
feat: versioned schema migration runner for the PostgreSQL sink

dcoric:
## Description

> Draft, stacked on #1532 (PostgreSQL sink). The diff below currently includes the #1532 changes. Once #1532 merges, this branch will be rebased onto `main` so only the migration-runner changes remain, and the draft flag will be removed.

Adds a versioned schema-migration runner for the PostgreSQL sink, replacing the one-shot `CREATE TABLE IF NOT EXISTS` bootstrap.

- `src/db/postgres/migrations.ts` defines an ordered, append-only list of migrations plus a `runMigrations` runner.
- On startup the runner ensures a `schema_migrations` table exists, takes a transaction-scoped advisory lock (so concurrently starting processes do not race), and applies any not-yet-recorded migrations in version order.
- Version 1 is the existing schema. Because it uses `CREATE TABLE IF NOT EXISTS`, databases already bootstrapped by the current code adopt the runner transparently (version 1 is simply recorded).
- Unit tests cover apply / skip-applied / idempotency / rollback; a gated integration test verifies the real apply and idempotency against PostgreSQL.

## Related Issue

Resolves #1558

### General

- [x] I have read the [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) format
- [x] I have a [FINOS CLA](https://finosfoundation.atlassian.net/wiki/spaces/FINOS/pages/75530375/Contribution+Compliance+Requirements) on file

### Documentation

- [x] Documentation has been added/updated for any new features

### Configuration

- [ ] If configuration schema (`config.schema.json`) was modified: (not modified in this PR)

### Tests

- [x] Tests have been added/updated for new functionality
- [x] Unit tests pass (`npm test`)
- [x] Linting and formatting pass (`npm run lint` and `npm run format:check`)
- [x] Type checks pass (`npm run check-types`)

Issue #1558, opened 2026-06-03
PostgreSQL sink: versioned schema-migration mechanism


dcoric:
Follow-up to #1497 / #1532.

The schema is currently created once via `CREATE TABLE IF NOT EXISTS` in `src/db/postgres/helper.ts`. There is no `schema_migrations` / version table, no `ALTER` path, and no concurrency guard. Once a deployment has created its tables, future schema changes will not apply to existing databases.

## Proposed

A lightweight, dependency-free runner: a `schema_migrations` table plus `pg_advisory_xact_lock` to serialise migrations across concurrently booting processes. The current `APP_SCHEMA_SQL` becomes migration 1.

This is a prerequisite for any schema evolution, including the `repo_users` normalisation.

dcoric:
Delivered by #1581, merged into the feat/postgres integration branch; lands on main with #1687.

PR #1590, opened 2026-06-12
feat: normalise PostgreSQL repo permissions into a repo_users join table

### LINKED ISSUE #1559: PostgreSQL sink: normalise repo permissions into a repo_users join table
Follow-up to #1497 / #1532. Tracked by the `TODO(#1497-followup)` in `src/db/postgres/repo.ts`.

Repo permissions (`canPush` / `canAuthorise`) are stored as a JSONB column for v1 parity with the mongo/fs backends. A normalised `repo_users(repo_id, user, role)` table would be more queryable and indexable.

Depends on #1558 — existing JSONB data must be migrated into the new table.

dcoric:
## Description

> Draft, stacked on #1532 (PostgreSQL sink) and #1581 (schema-migration runner). The diff below currently includes those changes. Once they merge, this branch will be rebased onto `main` so only the `repo_users` changes remain, and the draft flag will be removed.

Normalises PostgreSQL repo permissions out of the `repos.users` JSONB column into a dedicated `repo_users(repo_id, username, role)` join table, the follow-up to the v1 JSONB parity shipped in #1532.

- Migration v3 (`repo_users_table`) creates `repo_users` with a composite primary key `(repo_id, username, role)`, a `CHECK (role IN ('canPush','canAuthorise'))` constraint, a `repo_id` index, and `ON DELETE CASCADE` to `repos`. It backfills the table from the existing JSONB permissions in the same step.
- Migration v4 (`drop_repos_users_jsonb`) drops the now-unused `repos.users` column.
- The repo adapter (`src/db/postgres/repo.ts`) reads permissions via a single `LEFT JOIN` + `array_agg(...) FILTER (...)` (no N+1) and reconstructs the `{ canPush, canAuthorise }` shape callers expect; writes go through idempotent `INSERT ... ON CONFLICT DO NOTHING` / `DELETE`.
- Unit tests cover the read/aggregate, create, add/remove and delete-cascade paths. Integration tests exercise them against a real PostgreSQL, including an end-to-end test of the JSONB to `repo_users` backfill across single-user, multi-user and same-user-in-both-roles repos.

### Depends on

This relies on #1558 / #1581 (the versioned schema-migration runner): the backfill is delivered as migration v3 and needs the runner to apply it. It should merge after #1532 and #1581.

### Known follow-ups (non-blocking)

- `createRepo`'s optional create-time permission inserts are not wrapped in a single transaction. Impact is low because the public `createRepo` wrapper always passes empty permission arrays, so the loop is not exercised in the normal app flow; a follow-up can run it on one client inside `BEGIN`/`COMMIT`.
- Postgres adopts the file backend's stricter semantics over mongo's in two edges: adding a user to a non-existent repo raises a foreign-key violation (mongo silently no-ops), and permission arrays are de-duplicated (mongo's `$push` can store duplicates). Both are functionally invisible to the `.includes()`-based consumers.

## Related Issue

Resolves #1559

### General

### Documentation

### Configuration

- [ ] If configuration schema (`config.schema.json`) was modified: (not modified in this PR)

### Tests


src/db/postgres/schemaMigrations.ts
```diff
@@ -107,6 +107,41 @@ export const MIGRATIONS: Migration[] = [
...
+    role     TEXT NOT NULL CHECK (role IN ('canPush', 'canAuthorise')),
+    PRIMARY KEY (repo_id, username, role)
+  );
+  CREATE INDEX IF NOT EXISTS repo_users_repo_id_idx ON repo_users (repo_id);
+
+  -- Backfill the normalised table from the existing JSONB permissions. The
+  -- legacy repos.users column is dropped in a later migration once the adapter
+  -- reads and writes repo_users instead.
+  INSERT INTO repo_users (repo_id, username, role)
+  SELECT r._id, elem.username, 'canPush'
```
jescalada: Wondering if it's okay to select the usernames directly rather than the lowercased version here and a few lines below? Notice how in `addUserToRole` and `removeUserFromRole` we're lowercasing them:

```ts
const addUserToRole = async (
  _id: string,
  user: string,
  role: 'canPush' | 'canAuthorise',
): Promise<void> => {
  await query(
    `INSERT INTO repo_users (repo_id, username, role)
     VALUES ($1, $2, $3)
     ON CONFLICT DO NOTHING`,
    [_id, user.toLowerCase(), role],
  );
  await query(`UPDATE repos SET last_modified = $2 WHERE _id = $1`, [
    _id,
    new Date().toISOString(),
  ]);
};
```

We might also want a few unit tests to cover this specifically so that previously added user entries are still retrievable

src/db/postgres/schemaMigrations.ts
```diff
@@ -107,6 +107,41 @@ export const MIGRATIONS: Migration[] = [
...
+    FROM repos r,
+         jsonb_array_elements_text(coalesce(r.users->'canPush', '[]'::jsonb)) AS elem(username)
+  ON CONFLICT DO NOTHING;
+
+  INSERT INTO repo_users (repo_id, username, role)
+  SELECT r._id, elem.username, 'canAuthorise'
```
jescalada: Same as above

src/db/postgres/repo.ts
```diff
@@ -134,81 +189,28 @@ export const updateRepo = async (repo: Partial<Repo>): Promise<void> => {
...
-          )
-        ),
-            last_modified = $5
-      WHERE _id = $1`,
-    [_id, `{${role}}`, role, lowered, new Date().toISOString()],
-  );
+  if (users !== undefined) {
+    await query(`DELETE FROM repo_users WHERE repo_id = $1`, [_id]);
+    for (const username of users.canPush ?? []) {
+      await addUserToRole(_id, username, 'canPush');
```
jescalada: Should we be handling errors in these? There are a few operations happening in `updateRepo` - we should do something to prevent invalid or half-filled entries (repos added successfully, but without the user roles assigned) in case there's some unexpected error (disconnection/network glitch, etc.)

Do we have anything to ensure rollbacks happen on errors?

jescalada (commented):
Just a few comments on this one 👍🏼

src/db/postgres/schemaMigrations.ts
dcoric: Good catch - fixed. The backfill now inserts lower(elem.username) for both roles, matching the runtime writers, and ON CONFLICT collapses any case-only duplicates. Covered three ways: a unit pin on the migration SQL, an updateRepo unit test with mixed-case input, and a mixed-case seed in the backfill integration test.

src/db/postgres/schemaMigrations.ts
dcoric: Same fix as above - both role inserts lowercase now.

src/db/postgres/repo.ts
dcoric: Agreed, that was a real gap. updateRepo now runs everything (the row update, the repo_users replacement and the last_modified bump) inside a single transaction via a new withTransaction helper (BEGIN/COMMIT with ROLLBACK on failure, client always released), so a mid-flight failure cannot leave a repo without its roles. The helper has its own unit tests for the commit and rollback paths.

Issue #1497, opened 2026-04-08
Add PostgreSQL as a supported sink backend


dcoric:
## Summary

Add PostgreSQL as a first-class `sink` backend alongside the existing `fs` and `mongo` backends.

The goal is additive support, not a backend replacement:
- keep `fs` and `mongo` working
- preserve the current API/UI contract
- support repos, users, pushes, and persistent UI sessions when `sink.type === "postgres"`

## Proposed Scope

Implement:
- `postgres` as a new `sink[]` type in config/schema/generated types
- Postgres connection support via `connectionString`
- optional env fallback via `GIT_PROXY_POSTGRES_CONNECTION_STRING`
- a `src/db/postgres` adapter implementing the existing `Sink` interface
- Postgres-backed session storage for `express-session`
- schema bootstrap on startup for required tables/indexes
- docs and tests for the new backend

Non-goals for v1:
- no migration/import from existing `fs` or `mongo` data
- no dual-write/dual-read transition mode
- no broader auth or UI refactors unrelated to Postgres

## Implementation Shape

Use a thin compatibility model to minimize churn:
- `users` table
- `repos` table
- `pushes` table
- session table for `express-session`

Keep external identifiers opaque strings so the HTTP/UI contract does not change.

Use a lightweight Postgres integration approach:
- `pg` for data access
- `connect-pg-simple` (or equivalent) for session persistence
- startup bootstrap for schema creation
- no ORM in the initial implementation

## Must-Fix Requirements

Any Postgres implementation should explicitly cover these points:

- Ensure Postgres session persistence is actually active during service startup.
  The session store must not silently fall back to `MemoryStore` before the pool is initialized.

- Preserve current rejection semantics.
  `reject()` must store the rejection payload in the same field/shape used by existing backends.

- Keep user identity handling internally consistent.
  If `_id` remains part of backend behavior, the Postgres schema and adapter must support it correctly.
  Otherwise, the adapter contract should be normalized intentionally.

- Preserve array shape for repo permissions.
  Removing the last `canPush` or `canAuthorise` user must leave `[]`, not `null`.

- Preserve current push listing behavior.
  Postgres `getPushes()` should return pushes in descending timestamp order, matching current backends.

- Add tests for parity with existing backends.
  At minimum:
  - config selection/env fallback
  - users CRUD/update behavior
  - repos CRUD and permission mutation behavior
  - pushes CRUD/approve/reject/cancel behavior
  - session-store initialization behavior

[...]

PR #1591, opened 2026-06-12
feat: AWS RDS/Aurora IAM authentication for the PostgreSQL sink

### LINKED ISSUE #1561: PostgreSQL sink: AWS RDS IAM authentication support
Follow-up to #1497 / #1532.

The mongo backend supports AWS auth via `AWS_CREDENTIAL_PROVIDER`. The postgres sink only accepts a static connection string. Add IAM token-based authentication for RDS/Aurora (generate and refresh the auth token used for the connection).

dcoric:
## Description

> Draft, stacked on #1532 (PostgreSQL sink) and #1584 (connection configuration). The diff below currently includes those changes. Once they merge, this branch will be rebased onto `main` so only the IAM-auth changes remain, and the draft flag will be removed.

Adds AWS RDS/Aurora IAM token authentication to the PostgreSQL sink, the follow-up to #1497 / #1532 (the mongo backend already supports AWS auth via `AWS_CREDENTIAL_PROVIDER`; the postgres sink previously only accepted a static password).

- A new `awsIamAuth` option (`{ enabled, region }`) on the postgres sink. When enabled, the pool is built from the discrete `host` / `port` / `user` fields and the static password is replaced by a token provider function. `pg` invokes it for every new connection, so each one receives a fresh (~15 min) IAM token and refresh is automatic - no password is stored.
- The token is generated from the AWS SDK default credential chain via `@aws-sdk/rds-signer`, loaded lazily and declared as an **optional** dependency so non-AWS installs stay lean and still type-check. Enabling IAM auth without the dependency installed throws a clear, actionable error.
- TLS is required by RDS for IAM auth, so `ssl` defaults to `true` when unset; a `connectionString` is ignored in this mode.
- `region` resolves from `awsIamAuth.region`, then `AWS_REGION` / `AWS_DEFAULT_REGION`, then the SDK's default region resolution.
- Unit tests cover token-as-password, TLS defaulting, explicit ssl preserved, connection-string ignored, region resolution (config / `AWS_REGION` / `AWS_DEFAULT_REGION` and their priority), the default port, the missing-dependency error, and token-generation failure propagation. There is no integration test - the IAM path needs a real RDS/Aurora instance and credentials.

```json
{
  "type": "postgres",
  "host": "mydb.abc123.eu-west-2.rds.amazonaws.com",
  "port": 5432,
  "user": "gitproxy_iam",
  "database": "gitproxy",
  "awsIamAuth": { "enabled": true, "region": "eu-west-2" },
  "enabled": true
}
```

### Known follow-ups (non-blocking)

- `tsc` is not fully independent of the optional dependency: after a slim install (`npm install --omit=optional`) the dynamic-import specifier resolves to a `TS2307`. CI uses `npm ci`, which installs optional dependencies, so type checks and builds pass; runtime is unaffected. A follow-up could suppress the import-line diagnostic.
- Pre-existing and out of scope: `logConfiguration` logs the full sink config, which in the non-IAM (static password) path serializes credentials. IAM mode is safe (the token lives only in the pool's password function and is never written back into the config), but redacting credentials in that log line is a worthwhile separate fix.

## Related Issue

Resolves #1561

### General

### Documentation

### Configuration

- [x] If configuration schema (`config.schema.json`) was modified: regenerated `src/config/generated/config.ts`

### Tests


src/db/postgres/helper.ts
```diff
@@ -0,0 +1,280 @@
...
+ * the discrete fields are used. When neither is set, `pg` reads the `PG*`
+ * environment variables itself. When `awsIamAuth` is enabled, the static
+ * password is replaced by a generated IAM token and the discrete fields drive
+ * the connection.
+ */
+const buildPoolConfig = (db: DatabaseConfig): PoolConfig => {
+  const config: PoolConfig = {};
+  const iamAuthEnabled = Boolean(db.awsIamAuth?.enabled);
+
+  if (iamAuthEnabled) {
```
fabiovincenzi, line 106: Same silent-precedence point as [ in #1584](https://github.com/finos/git-proxy/pull/1584#discussion_r3569049412): when `awsIamAuth.enabled` is set, `connectionString` is dropped without any signal. A `console.warn` noting it's being ignored in IAM mode would make it observable

src/db/postgres/helper.ts
```diff
@@ -14,25 +14,137 @@
...
+    if (db.password !== undefined) config.password = db.password;
+    if (db.database !== undefined) config.database = db.database;
+  }
+
+  // TLS applies regardless of how the connection itself was configured. RDS IAM
+  // auth mandates TLS, so default it on when IAM is enabled and `ssl` is unset.
+  if (db.ssl !== undefined) {
+    config.ssl = db.ssl as PoolConfig['ssl'];
+  } else if (iamAuthEnabled) {
+    config.ssl = true;
```
jescalada, line 157: I put this PR through Claude since I wasn't sure how AWS RDS/IAM is normally set up - this is what it said:

> `ssl: true` almost certainly does not work against RDS (...)
> pg 8.x passes ssl: true through with rejectUnauthorized on, and the RDS server certificate chains to Amazon's private RDS root, which is not in Node's Mozilla trust bundle. So the documented happy path — set `awsIamAuth.enabled` and let ssl default — will fail with a certificate verification error, and the operator's obvious escape is `rejectUnauthorized: false`, which throws away exactly the transport security that IAM auth requires. To work, it needs the RDS CA bundle in ca.

Up to you if this warning is relevant or not 😃

jescalada (commented):
I put this PR through Claude and it says that it RDS + IAM might not work with `ssl: true`. Should we add integration tests to prove the option works even (if we're not testing AWS IAM directly)?

src/db/postgres/helper.ts
dcoric: Added - IAM mode now logs a warning when a connectionString is configured, naming the fields it actually uses, so the precedence is observable. Covered with a test.

src/db/postgres/helper.ts
dcoric: Verified against the pg source and it is right: with ssl: true, pg merges nothing into the TLS options, so Node verifies against its default Mozilla trust store, which does not include Amazon's RDS root CA - a real RDS endpoint fails certificate verification on the documented fallback path. Handled without downgrading security: the default stays verify-on, but startup now warns (both when ssl is defaulted in IAM mode and when supplied ssl options lack a ca) that the RDS CA bundle must go in ssl.ca, and the architecture doc now leads with the bundle requirement including the download URL and an explicit do-not-use rejectUnauthorized:false note. On integration tests: the ssl option passing is unit-covered; exercising real RDS IAM needs AWS credentials and an RDS endpoint, which CI does not have, so I kept it at unit level - happy to add a docker-based TLS test in a follow-up if you think it is worth the lane time.

jescalada: Thanks! I think it's good for now 👍🏼 Let's add extra tests if we ever find anything weird happening in the deployment

jescalada:
@dcoric Needs conflicts fixed before merging 🙂

Issue #1400, opened 2026-02-10
checkUserPushPermission is checking the wrong user's permission to push


kriswest:
**Describe the bug**
`checkUserPushPermission` is intended to check whether the user performing the push is allowed to do so. However, it is currently checking the permission of the committer of the last commit in the push in its current location, rather than the user pushing it into git proxy. 

Commits in git are tagged with two users' details: the _author_ (who created the commit and originally pushed it to a repo) and the _committer_ (initially the same as the author, ~but changes whenever the commit is pushed to a different repository or branch~ but changes whenever a commit is created or modified (e.g., via `git commit`, `git commit --amend`, `git rebase`, `git merge`) - `git push` does NOT change the _committer_. When we process a push in git proxy, we do so using the original committer of the last commit rather than the user performing the push - which is not the intent of the check. 

`parsePush` is extracting the `committer` and `committerEmail` from the last commit in the push and setting them as the `user` and `userEmail` in the action object: 

https://github.com/finos/git-proxy/blob/6fb63d0597212e9e0056330d3a90df9997f2d505/src/proxy/processors/push-action/parsePush.ts#L93-L96

This is not the correct user information and needs to be corrected in `parsePush`, without using `action.commitData`.

**To Reproduce**
1. Setup a project in Git proxy where you are allowed to contribute but no other user is
2. Have a different user push a commit to a branch of the respository.
3. Clone that branch yourself.
4. While that different user's commit is the last one on that branch, push the branch to git proxy.
5. `checkUserPushPermission` will reject the push as it will use the other user's information as the committer,

**Expected behavior**
`parsePush` should use the details of the user trying to push through the proxy as the `user` and `userEmail` values of the action (rather than the user that original pushed the last commit), ensuring that `checkUserPushPermission` checks the correct user's details.

kriswest:
@finos/git-proxy-maintainers We'll need to decide what to do here - the current check is clearly wrong. 

If we assume the above answer is correct (it may not be), we either need to switch to processing the 'author' of all commits (overlapping with/replacing `checkAuthorEmails` but possible from the data we already have with a few tweaks) and ignore the fact we don't know who is doing the push 👎  OR work on interpreting the transport layer's auth details to identify the user doing the push.

Note the content above about potentially NOT receiving the username and instead needing to use an API call (specific to the upstream platform) to resolve a PAT into the user details (also 👎  but may be necessary).

[...]

Issue #1450, opened 2026-03-09
GitProxy as a GitHub/GitLab App


coopernetes:
As a developer consuming Git Proxy, it would be convenient for GitProxy to act on behalf of users via a [OAuth style integration](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/creating-an-oauth-app) to enable functions such as PR/MR creation, comments and reviews, etc. There is a current gap in the lifecycle of a contribution which `git` over HTTPS or SSH cannot accommodate directly. These operations often are accessible via APIs. However, those APIs are specific to the SCM provider and not generic.

A few open questions:

1. How extensible is GitProxy's current authentication mechanisms today? Is it possible to extend the UI to handle SCM specific authentication flows (OAuth, etc)?
2. Is it secure enough to store user session tokens per SCM? Should this be a standalone application instead which integrates with GitProxy but is not directly embedded?
3. Which SCM providers is this possible to accomplish? There is known OAuth support for [GitHub](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/creating-an-oauth-app) and [GitLab](https://docs.gitlab.com/api/oauth2/). Other providers such as [Codeberg](http://codeberg.org/) (built on top of [Forgejo](https://forgejo.org/docs/latest/user/oauth2-provider/)) seem to have support for this as well but it's not guaranteed for all public `git` providers.

Related issues:

- https://github.com/finos/git-proxy/issues/910
- https://github.com/finos/git-proxy/issues/1027
- https://github.com/finos/git-proxy/issues/1240

coopernetes:
Aspirational. Feel free to re-open if there's bandwidth to implement. It should probably be tagged with a specific milestone. For now, it's just filling up the backlog.

Issue #428, opened 2024-01-29
Queue-based management of push request 


JamieSlome:
**Is your feature request related to a problem? Please describe.**

When a developer pushes new commits via Git Proxy, they are required to subsequently *re-push* their code once it has received approval, i.e. the commit SHA has been marked as reviewed and approved, and subsequent pushes will result in a forwarding of the traffic directly to the upstream repository.

**Describe the solution you'd like**

Unless a commit or set of commits pushed by the developer fails the "automated" synchronous tests (i.e. server-side push protections), the request should be forwarded to the upstream repository once it has received a review and approval. In simple terms, instead of requiring the developer to *re-push*, the server manages this on behalf of the developer.

The implementation would benefit from a queue-based implementation.

**Describe alternatives you've considered**

**Additional context**

PR #1604, opened 2026-06-19
feat(proxy): resolve push identity from token via SCM provider API

coopernetes:
## Description

parsePush uses the last commit's committer as the push user. This adds a new chain processor that extracts the token from HTTP Basic auth, calls the SCM provider's user API (GitHub `GET /user` for now), and maps the SCM login to a git-proxy user via the gitAccount field.

- TokenIdentityProvider interface with hostname-based dispatch
- GitHubTokenIdentityProvider calling api.github.com/user
- resolveUserFromToken chain processor (non-blocking on failure)
- findUserByGitAccount DB lookup (file + mongo)
- GET/PUT /api/v1/user/:username/git-account endpoints
- In-memory token→user cache (5 min TTL, SHA-512 keyed) to avoid hitting the SCM API on every push. Only positive resolutions are cached. Cache is evicted per-user when gitAccount is updated via the API.

> **⚠️ This PR does not store PATs or tokens.** The token is never written to disk or any database. The in-memory cache stores only a one-way SHA-512 hash of `provider:token` as the lookup key, alongside the resolved git-proxy username. The hash is non-reversible — the original token cannot be recovered from it. The cache lives in process memory only and is cleared on restart.

This doesn't block a push if the gitAccount isn't mapped in order to allow introduction of the gitAccount via the UI. This acts as a "soft" check for now unless the maintainer team wishes to adopt this model and use it as a requirement for authorising the "pusher" identity link that is missing as per what is described in #1400

### How it works

1. `resolveUserFromToken` runs in the push chain after `parsePush`, before `checkUserPushPermission`
2. Extracts the token from the HTTP Basic auth header (the password field)
3. Checks in-memory cache (SHA-512 of `provider:token` as key) — returns immediately on hit
4. Dispatches to a `TokenIdentityProvider` based on the upstream hostname (github.com → `GitHubTokenIdentityProvider`)
5. Calls `GET /user` with the token to get the SCM login
6. Looks up the git-proxy user by `gitAccount` field — if found, sets `action.user` and `action.userEmail` from the DB user and stores in cache
7. If no `gitAccount` match, falls back to using the SCM login directly (non-blocking)

### Cache note

The cache of token hashes is in-process memory intentionally — caches should not persist across restarts as it is an API driven optimization (respect user's own rate limits, don't look up data that doesn't change). A database-backed cache would be the natural next step if horizontal scaling becomes a concern, but for a single-process proxy this is sufficient and avoids a schema migration.

### Limitations

- Does not work for a generic git repository provider that doesn't provide a user API. Forcing this behaviour within Git Proxy will constrain its applicability to only these providers which have an API for identity lookups to match them to a valid Git Proxy user.
- For specific providers (GitLab, Forgejo/Codeberg/Gitea), an additional scope is needed. Originally documented here: https://github.com/RBC/fogwall/blob/main/docs/CONFIGURATION.md#token-scope-requirements

> ### Token scope requirements
>
> The SCM login check calls `GET /user` (or equivalent) on the upstream SCM using the pusher's token. The token must carry at least the following scope:
>
> | Provider | API endpoint | Additional scope |
> | -------- | ------------ | ---------------- |
> | GitHub | `GET https://api.github.com/user` | No additional scopes required for either classic or fine-grained PATs. |
> | GitLab | `GET {uri}/api/v4/user` | `read_user` or `api` (not recommended, prefer `read_user`) |
> | Codeberg | `GET https://codeberg.org/api/v1/user` | `read:user` |
> | Gitea | `GET https://gitea.com/api/v1/user` | `read:user` |

- BitBucket is just... weird... It has two separate sets of permissions between `git` and Bitbucket APIs. A user email can be linked between both "realms" but you cannot use your email to push code to that platform. Supporting Bitbucket proper requires some credential rewriting which is error-prone and brittle. See [BitbucketProvider](https://github.com/RBC/fogwall/blob/a3e95ebca63d4104adf7f31768d1bacea3e27b85/fogwall-core/src/main/java/com/rbc/fogwall/provider/BitbucketProvider.java) and [BitbucketIdentityFilter](https://github.com/RBC/fogwall/blob/main/fogwall-core/src/main/java/com/rbc/fogwall/servlet/filter/BitbucketIdentityFilter.java) in RBC/fogwall for details on what is needed in the HTTP flow. It's shared here as prior art/learnings only.

## Related Issue

related to #1400

### General

### Documentation

- [ ] Required user docs for adding their gitAccount (GitHub username in this current iteration)
- [ ] Update any architectural docs with the identity resolution

### Configuration

no configuration changes introduced

### Tests

- [ ] API route tests for `GET/PUT /api/v1/user/:username/git-account` (coverage exists but UI integration testing is deferred)


src/proxy/processors/push-action/resolveUserFromToken.ts
```diff
@@ -0,0 +1,116 @@
...
+  try {
+    const authHeader = req.headers?.authorization;
+    if (!authHeader) {
+      step.log('No Authorization header — cannot resolve push identity from token');
+      action.addStep(step);
+      return action;
+    }
+
+    const [scheme, encoded] = authHeader.split(' ');
+    if (!scheme || !encoded || scheme.toLowerCase() !== 'basic') {
```
coopernetes, line 42: I'm not sure how much this code path even makes sense given that [git by default uses HTTP Basic authorization](https://git-scm.com/docs/http-protocol#_authentication). A Git Proxy custom credentialHelper _could_ be an interesting angle to explore but that requires a customization on the client side. Maybe worth tracking in the backlog?

src/db/mongo/users.ts
```diff
@@ -34,6 +34,12 @@ export const findUserByEmail = async function (email: string): Promise<User | nu
   return doc ? toClass(doc, User.prototype) : null;
 };
 
+export const findUserByGitAccount = async function (gitAccount: string): Promise<User | null> {
+  const collection = await connect(collectionName);
+  const doc = await collection.findOne({ gitAccount: { $eq: gitAccount.toLowerCase() } });
```
coopernetes, line 39: Any desire to support an list of accounts here? `gitAccount` is somewhat of a holdover from v1. It's also singular across the whole user context - there's no shape in the data model today that supports associative git account by upstream provider/hostname.

Ideally, we revisit this shape in support of this PR. Something like this:

```
# MongoDB doc
{
  # existing keys...
  "username": "git-proxy-user",
  "email": "user@corpo-example.com",
  "gitAccounts": {
    "github.com": ["foo", "bar"],
    "gitlab.com": [ "baz" ]
  }
}
```

kriswest:
We should probably make a decision on whether we're going to start storing PATs/passwords for git accounts in git proxy, or github/gitlab apps etc.. Its been brought up multiple times as a necessary change to fulfil the ultimate goals of several Git Proxy contributors (e.g. raising PRs). 

Perhaps a topic for the next meeting.

coopernetes:
@kriswest this PR does not propose storing the PATs. Only a _irrevisible SHA-512 hash of the token_ to avoid excessive calls to the user lookup APIs. Unless I missed something?

kriswest:
@coopernetes sorry I wasn't suggesting it did! I think its a great approach to solving the pusher validation issue using the current data we have on users - it was just prompting me raise the fact that other desired features are going to need to store PATs or be authorised applications in order to do some of the other things contributors have made clear they want to try and achieve with git-proxy and that we should get on a make a formal decision as to whether we're going to take that on soon or not. as it effects the design of various features (such as this one and proxy format/doing the second push for you/raising the PR). I'm aware you are looking at similar features in fogwall and would love to have a chat about approaches for git proxy soon.

coopernetes:
Understood, my mistake. I misinterpreted. 

Some good candidates for relevant issues worth discussing in a design session:

- #1450 (closed for now, needs to be scoped properly - TL;DR "finos/git-proxy as an OAuth integration")
- #428 (historical but still relevant)

src/proxy/processors/push-action/resolveUserFromToken.ts
```diff
@@ -0,0 +1,125 @@
...
+      scmTokenCache.store(provider.name, token, user.username);
+    } else {
+      step.log(
+        `No git-proxy user has gitAccount '${identity.login}' — ` +
+          `falling back to SCM identity. ` +
+          `Users can associate their account via PUT /api/v1/user/:username/git-account`,
+      );
+      action.user = identity.login;
+      if (identity.email) {
+        action.userEmail = identity.email;
```
jescalada: Is GitProxy identity guaranteed to be the same as the SCM identity? If not, should we document that proper push identity can only be obtained if the SCM user's email is set to match?

src/proxy/processors/push-action/tokenIdentity.ts
```diff
@@ -0,0 +1,114 @@
...
+export class GitHubTokenIdentityProvider implements TokenIdentityProvider {
+  readonly name = 'github';
+
+  matches(hostname: string): boolean {
+    return hostname === 'github.com';
+  }
+
+  async fetchScmIdentity(token: string): Promise<ScmUserInfo | null> {
+    try {
+      const response = await fetch('https://api.github.com/user', {
```
jescalada, line 84: Might want to add a timeout here in case GitHub API hangs

src/service/routes/users.ts
```diff
@@ -62,6 +63,57 @@ router.get('/:id', async (req: Request<{ id: string }>, res: Response) => {
...
+router.put('/:username/git-account', async (req: Request<{ username: string }>, res: Response) => {
+  if (!req.user) {
+    res.status(401).json({ error: 'Authentication required' });
+    return;
+  }
+
+  const { username, admin } = req.user as { username: string; admin: boolean };
+  const targetUsername = req.params.username.toLowerCase();
+
+  if (username !== targetUsername && !admin) {
```
jescalada: Should `username` be lowercased as well before checking? Not sure if `req.user.username` is already normalized elsewhere 🤔

src/service/routes/users.ts
```diff
@@ -62,6 +63,57 @@ router.get('/:id', async (req: Request<{ id: string }>, res: Response) => {
   res.send(toPublicUser(user));
 });
 
+// Get git account (SCM identity) for a user
+router.get('/:username/git-account', async (req: Request<{ username: string }>, res: Response) => {
+  const targetUsername = req.params.username.toLowerCase();
```
jescalada: I think we might be missing an auth check here:

```suggestion
  if (!req.user) {
    res.status(401).json({ error: 'Authentication required' });
    return;
  }
  const targetUsername = req.params.username.toLowerCase();
```

jescalada (approved):
LGTM - wondering if any other @finos/git-proxy-maintainers wants to check it out before merging?

Also: Is this a complete fix for #1400 or is there anything else we need to patch up for `checkUserPushPermission` to work as expected?

src/service/routes/users.ts
coopernetes: This new endpoint is consistent with the rest of the users Router endpoints. If auth is needed here across the suite of endpoints, let's track that in a separate issue+PR.

src/service/routes/users.ts
coopernetes: Usernames are normalized to lowercase at the DB layer on both write (`createUser`) and read (`findUser`), so req.user.username from the session is guaranteed lowercase. Can skip the additional normalization at the route level.

src/proxy/processors/push-action/resolveUserFromToken.ts
coopernetes: The only guarantee in this flow is that the `identity.login` string is reliable insofar as the PAT will always be linked to a real GitHub user (except in odd cases like using a separate GitHub OAuth App to generate an OAuth token then feeding that into a git client... not impossible but likely not a common setup for developers). 

This will decouple any email-to-user linkage from this new token resolver. The existing commit metadata in the chain already captures committer/author identity and links it to a GitProxy user account. This new step will just run as a later step to link back to the `gitAccount`. 

`gitAccount` was previously "overloaded" and set to an email address to link to internal GitProxy user account objects based on conventions established from the original maintainer (as far as I remember, I could be mistaken). As I understand it, that was an internal, organization specific convention. This PR reclaims that field to mean the pusher's SCM profile name / GitHub login making it a reliable identifier for who actually performed the push.

I'm gonna remove the `if (identity.email)` check. As discussed in #1400, the email address is only ever returned in that `GET /user` endpoint if a user explicitly goes against the private-by-default setting of hiding it. Almost no one on GitHub enables that setting because of the obvious privacy implication.

coopernetes:
Thanks for the review @jescalada , addressed those comments so just one final round of review. On your question regarding 1400, this is a partial fix. When a user has their `gitAccount` set, `checkUserPushPermission` will correctly check the actual pusher's permission. Without it, `parsePush` still falls back to the last committer. This is intentional and documented in the PR description; it's designed as a non-blocking soft check to allow gradual adoption before the decision on changing `gitAccount` to be an associative map of SCM-hostnames-to-identity (see https://github.com/finos/git-proxy/pull/1604#discussion_r3444701174) is sorted. Happy to add those changes here though so it's a complete solution to 1400.

src/service/routes/users.ts
jescalada: I was going by the SSH-related endpoints which all have the check, including `GET /:username/ssh-key-fingerprints` (although these are plugged into the UI).

I noticed the `git-account` endpoints are standalone at the moment. Should we add a UI form that interacts with these endpoints, and should the `GET /:username/git-account` be protected with the `!req.user` guard in that case?

src/db/mongo/users.ts
jescalada: Linked this to the original `gitAccount` removal issue: #1258

jescalada (approved):
LGTM! Just one previous comment of mine to check, and a conflict in `chain.ts` 👍🏼

src/proxy/chain.ts
```diff
@@ -25,6 +25,8 @@ import { handleErrorAndLog } from '../utils/errors';
 import { createProgressWriter } from './sideband';
 
 const branchPushChain: Processor['exec'][] = [
+  proc.push.parsePush,
```
coopernetes: @jescalada is this step no longer required?
```suggestion
```

src/service/routes/users.ts
coopernetes: @jescalada I ditched these new endpoints since there's already a UI driven POST which I can insert the token cache eviction into. The new endpoints are essentially a no-op from the UI and duplicates what already exists. I think this is good to merge now.

jescalada (approved):
LGTM! 🚀

PR #1411, opened 2026-02-21
feat: improve agentic AI development flow and guidelines

jescalada:
Closes #1410.

**Note:** This PR summary was generated via the `/summary` command and manually tweaked.

**What:** Add `CLAUDE.md` and a set of Cursor/Claude commands under `.claude/commands` to support AI-assisted development and consistent agent workflows.

**Why:** Agents need proper context for GitProxy’s architecture and conventions. Furthermore, the new commands help speed up typical tasks such as issue creation, PR summaries, docs updates, linting, and review follow-up to improve the dev experience and help humans focus on quality control and more pressing issues.

**Changes:**

- Add `CLAUDE.md` with project overview, build/run steps, high-level architecture (proxy, service, config, UI), development guidelines, and agent workflow (delegation).
- Add `.claude/commands/` with seven commands:
  - `address-review` (apply review feedback)
  - `commit` (incremental conventional commits)
  - `docs` (update README/Architecture/CLAUDE from branch changes)
  - `implement` (issue → branch → code → commit → summary)
  - `issue` (summarize issue for implementation)
  - `lint` (run lint/format)
  - `summary` (PR description summary).
- Update `.gitignore` to ignore `/plans` (where command outputs are stored).
- Update `README.md` Contributing section to point to `CLAUDE.md` and `.claude/commands` for AI-assisted development.

**How to verify:**

1. Open `CLAUDE.md` and confirm it describes proxy, service, config, UI, and push/pull/default chains.
2. Run each command (e.g. `/summary`, `/docs`) and confirm they behave as described in their files.
3. Confirm `plans/` is ignored by Git and that `README.md` links to `CLAUDE.md` and `.claude/commands`.

Credit goes to @sudiptob2 for original implementation in [Armada Spark](https://github.com/armadaproject/armada-spark)!

Issue #1410, opened 2026-02-21
Improve and standardize Agentic AI tooling


jescalada:
With all the new Agentic AI tools out there, it'd be great to add guidelines/standards to improve the quality of generated code and developer experience.

This should help Agentic AI usage be more consistent, allow for better context extraction (and thus improved code quality), and also speed up governance by delegating common maintenance tasks to AI.

Some of the tasks which would be great to delegate include:

- Addressing PR Reviews: Automatically fix trivial/low-impact PR comments, reply if necessary.
- Committing: Automate the commit process. Must figure out the diff and write an appropriate commit message in conventional commit format, and potentially sign-off commits
- Documentation: Check whether the current branch has modified anything relevant to the documentation, and updates it if necessary
- Implementation: Implement trivial/low-priority issues that have little chance of breaking the project or introducing security risks
- Issue creation: Create issues with proper context, reproduction steps, triaging, based on Slack chats, unstructured error messages, etc.
- Linting/formatting: Apply linting or formatting to make sure CI checks will pass
- Summarizing: Generate a PR description based on the changes in the current branch versus main

Note that for all of these tasks, the agent **MUST** ask for human review before doing anything on their behalf such as committing, replying to PR comments, etc.

### Additional context
One of our org's repos already has a working, battle-tested setup we could adapt to GitProxy: https://github.com/armadaproject/armada-spark

re-vlad:
> [@re-vlad](https://github.com/re-vlad) Absolutely! Feel free to reopen a PR with the desired changes - I haven't had time to look at mine!

Hi @jescalada! I'd like to contribute to this PR and help move it forward. Would you be open to adding me as a collaborator on your fork so I can push changes directly to your branch? If you prefer, I can also open a separate PR that builds on your work and credits you

jescalada:
@re-vlad Thanks! I added you as a contributor to my fork to expedite the dev process - feel free to pull the latest changes to integrate yours to the branch. I believe there might've been new changes since the last time, as Gemini support was already working after Denis' additions... 🤔

PR #1605, opened 2026-06-19
ci: add Grype CVE scanning workflow, clean up vulnerable deps

coopernetes:
## Description

Adds CVE scanning via [Grype](https://github.com/anchore/grype) and fixes all high/critical vulnerabilities in the main project's dependencies.

Note that the [OpenSSF scorecard](https://securityscorecards.dev/) is flagging a number of existing vulnerabilities. Also, for some reason (oversight?), Dependabot alerts are disabled.

Dependency review remains and will catch any _new_ dependencies being added. The CVE scan workflow is more complete and also adds a scheduled scan every Monday morning for ongoing hygiene. Not a required check for PRs so should be fine to introduce — audit the findings then action them accordingly.

Alternatively, this PR can be closed and Dependabot alerts can just be enabled since the version bumping is already handled by Dependabot anyways. This will require intervention from FINOS and agreement on if Dependabot is the tool of choice for generating alerts.

Findings available here: https://github.com/finos/git-proxy/security/code-scanning?query=is%3Aopen+pr%3A1605

The new workflow will also output the results (has to scan twice unfortunately) to a table format for quick review in the workflow logs.

### CVE scanning workflow

- Runs on push to main, PRs to main, and weekly (Monday 6am UTC)
- Scans the working directory with `--only-fixed` (only reports CVEs that have a fix available)
- Fails the job on high or above (`--fail-on high`)
- Uploads SARIF to GitHub's Security tab (Code Scanning alerts)
- Table output in CI logs for quick triage
- Grype DB cached between runs to avoid re-downloading on every run
- `./experimental/**` excluded — `li-cli` and `license-inventory` have their own dependency trees with unrelated vulnerabilities. The exclude can be removed once those packages are updated.

### Dependency bumps

All same-major, no breaking changes (🤞):

| Package | Before | After | CVE | Severity |
|---|---|---|---|---|
| `concurrently` | 9.2.1 | 9.2.3 | `shell-quote` GHSA-w7jw-789q-3m8p | Critical |
| `vitest` | 3.2.4 | 3.2.6 | Arbitrary file read via UI server | Critical |
| `@vitest/coverage-v8` | 3.2.4 | 3.2.6 | Depends on vitest | Critical |
| `vite` | 7.3.1 | 7.3.5 | GHSA-fx2h-pf6j-xcff | High |
| `cypress` | 15.9.0 | 15.17.0 | `tmp`, `qs`, `uuid` transitives | High/Moderate |
| `react-router-dom` | 6.30.3 | 6.30.4 | GHSA-2j2x-hqr9-3h42 | Moderate |

### npm overrides

For transitive deps whose direct parents are already at latest and haven't released fixes:

| Package | Before | After | CVE | Severity |
|---|---|---|---|---|
| `form-data` | 4.0.5 | 4.0.6 | GHSA-hmw2-7cc7-3qxx (CRLF injection) | High |
| `tmp` | 0.2.5 | 0.2.6 | Path traversal via unsanitized prefix | High |

`form-data` is pinned to `^4.0.5` in [cypress's own resolutions](https://github.com/cypress-io/cypress/blob/develop/package.json). `tmp` is a direct dependency of cypress at `~0.2.4`. Both overrides can be removed when cypress releases a version that bumps these.

### Result

Total vulnerabilities: **19 → 7** (0 critical, 0 high). The remaining 7 are moderate/low in `nyc` (requires major 17→18), `@babel/core` (low), and `brace-expansion` (moderate via `c8`).


.github/workflows/cve.yml
jescalada, line 1: Is this scanning our root `package-lock.json` only, or is our `website/package.json` and our Docker image included as well? The OSSF Scorecard vulnerabilities are largely coming from the website rather than the main package, meanwhile Docker-specific dependencies are excluded entirely from the scorecard.

Grype could be helpful for catching issues outside the main package - though I'm worried about false positives or having the CI constantly break due to the massive total number of dependencies.

jescalada (approved):
LGTM - just wondering if this also covers the Docker image and website dependencies. I prefer Grype over Dependabot since we can see vulnerabilities introduced in our PRs, whereas Dependabot would detect them after the fact.

Also, it would be nice to have it scan the Docker image, since Node has its own dependencies that aren't listed in our `package.json`.

jescalada:
Another question: What's the recommended process for fixing up vulnerabilities detected by Grype? Suppose a contributor sees their CI fail due to Grype - should they be running `npm run audit --fix` to fix it? Does this *guarantee* to fix the vulnerabilities, or should we have a local Grype script `npm run grype:fix` to smooth out the dev experience?

Perhaps we should add documentation for the audit/fix process so we don't discourage new contributions 🤔

coopernetes:
Thanks for the review @jescalada and good questions. I'll address the comments above here then respond on that thread about the other dependency scans we might want to add.

>  just wondering if this also covers the Docker image and website dependencies. I prefer Grype over Dependabot since we can see vulnerabilities introduced in our PRs, whereas Dependabot would detect them after the fact.

We can absolutely use Grype for both the image scanning (I believe this was the primary use for grype, scanning container images similar to [trivy](https://github.com/aquasecurity/trivy)). We just need to add another grype job on the image itself. Note that it _does_ require the image to be pushed to a registry if we wish to keep it as its own GHA workflow/job. Alternatively, we can add grype scan as a post-build step. 

Note, we currently only build the Docker image on the main branch and from tag/release. As its currently designed, we don't perform a docker build as part of a PR. Something to consider is adding `docker build` on a pull request (just without publishing) to catch issues related to that deployment image which wouldn't be caught at PR time. Then it would be trivial to add a grype scan step.

~On the website, since it uses npm, it is also scannable. I'm not sure if the `dir:.` target is recursive today.~ Confirmed these dependencies are included as well.

> What's the recommended process for fixing up vulnerabilities detected by Grype?

It's a bit of a case-by-case basis. Maybe a good candidate to add in custom commands/SKILLS (described in #1410) to make repeatable analysis of "dependency is vulnerable, check for a fixed-version, update a dependency manifest file". Remediating a Docker vulnerability is quite different from npm (and in the Docker case, it's multi-dimensional - git-proxy's own dependencies vulnerable? what about a transitive vulnerability from a base image?).

.github/workflows/cve.yml
coopernetes: After some local testing, it looks like grype is correctly reporting both the main app and the docusaurus app's dependencies recursively. I'll push up an enhancement to the output report so it's clear from the workflow logs:

```
$ grype dir:. --output template -t .grype-report.tmpl
SEVERITY | ADVISORY            | PACKAGE                                            | FIXED IN        | SOURCE
-------- | ------------------- | -------------------------------------------------- | --------------- | ------
Critical | GHSA-xq3m-2v4x-88gg | protobufjs@7.4.0                                   | 7.5.5           | /experimental/license-inventory/package-lock.json
```

On the note of false positives in CI - hopefully, this configuration will reduce the chance of false positives by only flagging actionable vulnerabilities:

https://github.com/finos/git-proxy/pull/1605/changes#diff-d929405c8d1f27de6b8eaa31eab07468a9bc588d8466f23666f329595e8a70edR4-R6

This is why I think it's wise to introduce this gradually and not as a required PR check. If the CVE scanning is flagging something as a false positive, it shouldn't block a merge but then should be added to the [.grype.yaml `ignores` config](https://oss.anchore.com/docs/reference/grype/configuration/) so it doesn't show up again (once a maintainer or trusted contributor has validated it of course).

coopernetes:
After resolving the main application dependencies, it's safe to close out the OpenSSF code scanning alert (security/code-scanning/64) since it's using osv under the hood and I don't believe scorecard is validating the versions properly compared to the full Grype scan against all the dependencies in main.

jescalada:
@coopernetes Thanks for the detailed response! I'm happy to help out with setting GHA to run Docker build on each PR, and also with the `SKILL.md` for handling vulnerabilities. The more we can automate, the better the quality of the resulting code and artifacts 🙂 

Note that there's an open PR already for adding basic skills for contributors: #1411 I could just extend the skills to cover remediation for vulnerable dependencies on both NPM and Docker via Grype. Would appreciate a quick look on that so we can merge!

jescalada (approved):

kriswest:
Conflicts resolved @jescalada, over to you to re-approve and merge (and to remind us to review behaviour and decide on whether/what we're keeping)

jescalada (approved):

PR #1564, opened 2026-06-06
chore(deps): bump vitest to v4 and fix tests

jescalada:
## Description

Bumps Vitest to v4 and fixes resulting failing tests from syntax changes.

Despite adding the `uiPort` mocks to fix some `EADDRINUSE` errors when running tests in parallel, I decided to keep the sequential tests (by disabling the `fileParallelism` option, formerly `poolOptions.forks.singleFork`).

This is because a few tests still fail when running `npm run test-shuffle`, and the MongoDB integration tests also fail on the CI when enabling parallelism. I made an issue to track this so we can benefit from parallel test execution speedup: #1565 

### General

- [x] I have read the [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) format
- [x] I have a [FINOS CLA](https://finosfoundation.atlassian.net/wiki/spaces/FINOS/pages/75530375/Contribution+Compliance+Requirements) on file

### Tests

- [x] Tests have been added/updated for new functionality
- [x] Unit tests pass (`npm test`)
- [x] Linting and formatting pass (`npm run lint` and `npm run format:check`)
- [x] Type checks pass (`npm run check-types`)

PR #1628, opened 2026-07-01
test(ssh): improve SSH proxy test coverage

### LINKED ISSUE #1586: Improve SSH Test Coverage
SSH support has just been added in #1332, adding >10k lines of code. The total test coverage on the PR was about ~70%, leaving lots of room for improvement:

Image

Achieving >90% coverage on SSH is difficult due to the implementation complexity and hard-to-reach edge cases (SSH connection flow is very if-else heavy with extensive error handling), but we should at least get more coverage on the following three files in `src/proxy/ssh` which account for most of the missing coverage:

Image

fabiovincenzi:
## Description

Add tests to improve SSH proxy coverage, focusing on the four modules with the most gaps:

- **server.ts**: push flow (no-op detection, security chain integration, pack size/chunk limits, timeout), path validation, auth edge cases, debug mode
- **GitProtocol.ts**: happy paths for all 4 exported functions
- **sshInternals.ts**: all exported functions with failure modes and error messages
- **AgentForwarding.ts** : CHANNEL_OPEN_CONFIRMATION handler, lock recovery

## Related Issue

Resolves #1586

### General

### Documentation

- [x] Documentation has been added/updated for any new features (N/A — test-only changes)

### Configuration

- [x] If configuration schema (`config.schema.json`) was modified (N/A — no config changes)

### Tests


test/ssh/GitProtocol.test.ts
```diff
@@ -269,23 +323,806 @@ describe('GitProtocol', () => {
...
+        };
+        return c;
+      });
+
+      // connectToRemoteGitServer passes requireAgentForwarding: true
+      // fetchGitHubCapabilities does not pass it (defaults to true)
+      // Both call validateSSHPrerequisites
+      vi.mocked(validateSSHPrerequisites).mockClear();
+
+      await fetchGitHubCapabilities(
```
jescalada, line 1131: If `fetchGitHubCapabilities` defaults to `true` and then we assert that `validateSSHPrerequisites` was called, isn't that incompatible with the test's description?

test/ssh/AgentForwarding.test.ts
```diff
@@ -334,105 +337,267 @@ describe('AgentForwarding', () => {
...
-          _count: 2,
-        },
-      };
+          _chanMgr: { _channels: {} as Record<number, any>, _count: 0 },
+        } as any;
+      }
+
+      function mockChannelModule(channelImpl?: (...args: unknown[]) => unknown) {
+        const mockChannel = { on: vi.fn(), write: vi.fn(), end: vi.fn() };
+        getChannelModuleSpy = vi.spyOn(sshInternals, 'getChannelModule').mockReturnValue({
```
jescalada: These `mockReturnValue` usages will likely fail after upgrading to Vitest 4 (underway in #1564).

If we merge this in first, and I can fix up any failing tests in that PR, otherwise you can fix those by passing in a function that returns an object, rather than the object itself.

test/ssh/server.test.ts
```diff
@@ -913,4 +913,462 @@ describe('SSHServer', () => {
...
+          privateKeyPath: `${testKeysDir}/test_key`,
+          publicKeyPath: `${testKeysDir}/test_key.pub`,
+        },
+        port: 2222,
+        enabled: true,
+        debug: true,
+      } as any);
+
+      const debugServer = new SSHServer();
+      expect(debugServer).toBeDefined();
```
jescalada, line 1390: We may want to check some other properties to make sure the server actually did start in debug mode 👍🏼

test/ssh/GitProtocol.test.ts
```diff
@@ -269,23 +323,806 @@ describe('GitProtocol', () => {
...
+      expect(result).toBeInstanceOf(Buffer);
+      expect(result.toString()).toContain('hello world');
+      expect(mockRemoteStream.end).toHaveBeenCalled();
+    });
+
+    it('should accumulate multiple data chunks before flush', async () => {
+      const ssh2 = await import('ssh2');
+      const Client = ssh2.Client as any;
+
+      const chunk1 = Buffer.from('0010chunk one\n');
```
jescalada: I think these pkt-line definitions might be wrong - though it's not clear why the test passes regardless:

`0010` means the payload has 16 bytes, but the actual bytes in the payload is 14 (4 + 9 + 1):

`"0010"` = 4 bytes
`"chunk one"` = 9 bytes
`"\n"` = 1 byte

Same for the other chunk which excludes the flush.

```suggestion
      const chunk1 = Buffer.from('000echunk one\n');
```

test/ssh/GitProtocol.test.ts
```diff
@@ -269,23 +323,806 @@ describe('GitProtocol', () => {
...
+      const chunk2 = Buffer.from('0010chunk two\n0000');
```
jescalada:
```suggestion
      const chunk2 = Buffer.from('000echunk two\n0000');
```

test/ssh/GitProtocol.test.ts
```diff
@@ -269,23 +323,806 @@ describe('GitProtocol', () => {
...
+        },
+      };
+    }
+
+    it('should collect data and return buffer when flush packet is received', async () => {
+      const ssh2 = await import('ssh2');
+      const Client = ssh2.Client as any;
+
+      // Build a valid pkt-line payload with flush packet
+      // "0012hello world\n0000"
```
jescalada: This comment may also be inaccurate since "hello world\n" has 12 bytes + 2 + 2 = 16 bytes, or `0010` in hexadecimal

```suggestion
      // "0010hello world\n0000"
```

jescalada (approved):
Excellent work on getting our coverage back to standard! 🎉 

Just a few minor comments/suggestions from me. I'd like to merge this ASAP for the coverage increase, despite conflicts with the Vitest 4 upgrade PR #1564. I'm happy to deal with failing tests later 🙂

test/ssh/GitProtocol.test.ts
fabiovincenzi: Good catch! Fixed.

test/ssh/GitProtocol.test.ts
fabiovincenzi: You're right, applied both suggestions. The old assertions only did .toContain(), so the length prefixes were never checked. The test now decodes the result with parsePacketLines and asserts the two lines plus that the flush packet is consumed exactly at the end of the buffer.

test/ssh/GitProtocol.test.ts
fabiovincenzi: Right! That title was a leftover. Renamed it to match.
The opposite (skip) path isn't easily testable: it's unreachable without exposing internals just for the test.

test/ssh/server.test.ts
fabiovincenzi: Good point, toBeDefined() passed regardless of debug mode. The only observable effect of debug mode is that the constructor sets a debug logging function in the options passed to ssh2.Server, so I now spy on that constructor and assert options.debug is a function when debug is enabled.

jescalada (approved):
LGTM! 👍🏼

PR #1640, opened 2026-07-08
feat: execute all steps and collect errors at end of chain

jescalada:
## Description

This PR implements one of [fogwall's](https://github.com/RBC/fogwall) quality-of-life features: collecting all errors in the chain and reporting to the user, so they don't have to make multiple pushes and fix one-by-one. This still fails mid-chain if non-validator steps (`parsePush`, `pullRemote`, `writePack`, etc.) error out.

`collectibleSteps` are defined in `chain.ts` (steps that can be collected and reported later, as opposed to immediately failing).

image

## Related Issue

Resolves #

### General

### Documentation

### Configuration

### Tests


src/config/generated/config.ts
```diff
@@ -36,6 +36,13 @@ export interface GitProxyConfig {
    * List of repositories that are authorised to be pushed to through the proxy.
    */
   authorisedList?: AuthorisedRepo[];
+  /**
+   * When enabled, the push validation chain continues past recoverable check failures (commit
+   * messages, author e-mails, secret scanning, etc.) and reports every rejection reason in a
+   * single push, instead of stopping at the first failed check. Security and infrastructure
+   * failures still stop the chain immediately.
+   */
+  collectAllChainErrors?: boolean;
```
coopernetes: Does this need to be configurable? Seems like a sensible new default. Since it only affects the response behaviour back to git clients, it shouldn't be considered a breaking change and needing a toggle.

src/proxy/chain.ts
```diff
@@ -55,8 +56,49 @@ const pullActionChain: Processor['exec'][] = [proc.push.checkRepoInAuthorisedLis
...
+ * that a single push reports every rejection reason at once.
+ *
+ * Steps not listed here stop the chain immediately when they fail
+ * such as repository authorisation, user push permission, or later steps
+ * depending on their output (pullRemote, writePack, getDiff)
+ */
+const collectibleSteps = new Set<Processor['exec']>([
+  proc.push.checkMessages,
+  proc.push.checkAuthorEmails,
+  proc.push.checkHiddenCommits,
```
coopernetes: Given that this step resolved an exploit, I would recommend making this one a "fail out early" step. The rest of the checks are sound and should run when a valid push with a set of walkable (traceable from base HEAD/main) refs is sent.

src/proxy/chain.ts
```diff
@@ -55,8 +56,49 @@ const pullActionChain: Processor['exec'][] = [proc.push.checkRepoInAuthorisedLis
...
+ * Steps whose failures are recoverable by the user (bad commit message, bad
+ * author e-mail, detected secret, etc). When collectAllChainErrors is enabled,
+ * a failure in one of these steps is recorded but the chain keeps running so
+ * that a single push reports every rejection reason at once.
+ *
+ * Steps not listed here stop the chain immediately when they fail
+ * such as repository authorisation, user push permission, or later steps
+ * depending on their output (pullRemote, writePack, getDiff)
+ */
+const collectibleSteps = new Set<Processor['exec']>([
```
coopernetes: Worth adding a class/object property like `private readonly boolean isCollectible` directly on the Processor type? Then each step can self-declare it's behaviour and this list doesn't need to remain statically defined (and needing updates when new checks are introduced).

coopernetes (approved):
Overall LGTM, just a few notes/suggestions.

src/config/generated/config.ts
kriswest: Agreed, I think you can lose the config option.

src/proxy/chain.ts
kriswest: Could also be worth thinking about how plugin behave - they could be collectible or not...

kriswest (approved):
LGTM, although I agree with @coopernetes that the config option can be removed and that processors could self declare their behaviour, which could have an impact on plugins. Up to you if you deal with the alter point in this PR or raise a follow-up issue. However, best to drop the config setting now if we don't need it

jescalada:
@kriswest @coopernetes Ready for another look! I'm not sure about the plugin behavior - should these be collectible as well? I'm happy to open another PR for this to get this merged faster 👍🏼

coopernetes (approved):
LGTM

PR #1641, opened 2026-07-09
feat: implement disconnect detection, cancel pushes on disconnect

jescalada:
## Description

Implements disconnect detection which marks pushes as canceled if the connection to the client is lost (such as when sending SIGINT after calling `git push`, or internet disconnection).

Originally developed in Fogwall, and explained [here](https://gist.github.com/coopernetes/d02d48efa759282ff8187da0d5dcae64#what-persistent-possession-enables). This PR demonstrates that implementing persistent clones isn't necessary for detecting client disconnects.

A new step called `clientDisconnected` is added if the connection was lost - so the UI clearly reports the canceled status and reason for canceling:

image

### Client-side

image

### Server-side

Note that `clearBareClone` runs in the `finally` block (right after `clientDisconnected` "step" is logged.

image

## Related Issue

Resolves #

### General

### Documentation

### Tests

### DIFF
--- a/src/proxy/chain.ts
+++ b/src/proxy/chain.ts
@@ -17,7 +17,7 @@
 import { Request, Response } from 'express';
 
 import { PluginLoader } from '../plugin';
-import { Action, RequestType, PushType } from './actions';
+import { Action, Step, RequestType, PushType } from './actions';
 import * as proc from './processors';
 import { Processor } from './processors/types';
 import { attemptAutoApproval, attemptAutoRejection } from './actions/autoActions';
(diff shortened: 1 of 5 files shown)


kriswest:
@jescalada could you add some more to the description of this issue? I'm not quite understanding it. Was it an experiment towards the implementation of a new proxy mode (ala fogwall) where you stay connected until approved then complete the push?

If the client does disconnect, we don't have to fail the push do we? It can still go through review and then be pushed again I assume...

PR #1642, opened 2026-07-10
fix(proxy): #1588 accept smart-HTTP paths without .git suffix

### LINKED ISSUE #1588: processUrlPath rejects smart-HTTP requests when the repo path omits .git
**Describe the bug**

`processUrlPath` only accepts request paths that contain `.git` before the smart-HTTP suffix (regex `(.+\.git)(\/.*)?/`). Many clients send paths without `.git`, e.g. `/owner/repo/info/refs?service=git-upload-pack`. Those requests fail parsing before the plugin chain runs. GitHub accepts both forms.

**To Reproduce**

1. Run `@finos/git-proxy@2.0.0` with a repository on the authorised list.
2. Send a smart-HTTP discovery request without `.git` in the path: `GET /owner/repo/info/refs?service=git-upload-pack`
3. Send the same request with `.git`: `GET /owner/repo.git/info/refs?service=git-upload-pack`
4. See error on step 2: smart-HTTP pkt-line `Invalid request received` (HTTP 200); `git clone` fails (`could not determine hash algorithm; is this a git repository?`). Step 3 succeeds.

**Expected behavior**

Both path forms parse and reach the normal action/plugin chain. Normalising `.git`-less paths to a `repoPath` ending in `.git` internally is fine.

**Screenshots**

N/A

- OS: Linux; macOS
- Browser: N/A (HTTP client / `git` CLI)
- Version: Node 20+; `@finos/git-proxy` 2.0.0

**Smartphone (please complete the following information):**

- Device: N/A
- OS: N/A
- Browser: N/A
- Version: N/A

**Additional context**

Suggested change inside `processUrlPath` (after the existing `.git` match fails):

```js
const SMART_HTTP = /^(.+?)(\/(?:info\/refs|git-upload-pack|git-receive-pack)(?:\?.*)?)$/;

const smartHttp = requestPath.match(SMART_HTTP);
if (smartHttp) {
  return { repoPath: smartHttp[1] + '.git', gitPath: smartHttp[2] };
}
```

Happy to open a PR if useful.

re-vlad:
## Description

Git smart-HTTP clients may send repository paths **without a `.git` suffix** (e.g. `GET /owner/repo/info/refs?service=git-upload-pack`). GitHub accepts this form; Git Proxy previously did not.

- `processUrlPath` only matched paths containing `.git` (`(.+\.git)(\/.*)?`). For `.git`-less smart-HTTP URLs it returned `null`, so - `proxyFilter` rejected the request with `Invalid request received` before the plugin chain ran.

This PR fixes that in two places (both are required for a full clone/fetch):

1. `processUrlPath` - after the existing `.git` regex, add a smart-HTTP suffix fallback for paths ending in `/info/refs`, `/git-upload-pack`, or `/git-receive-pack`. Normalise `repoPath` to end with `.git` internally so `parseAction` and `getRepoByUrl` continue to work unchanged.

2. `isPackPost` - make `.git` optional in the POST pack URL regex so `POST /owner/repo/git-upload-pack` is recognised and `extractRawBody` runs. Without this, discovery could succeed but clone/fetch would still fail on the second HTTP request.

**Complementary to, not a substitute for:** v1->v2 DB URL migration (#1535 / #1543), which normalises stored `repos.url`. This PR fixes **inbound wire paths** clients send at runtime.

## Related Issue

Resolves #1588

### General

### Documentation

_N/A — behaviour fix aligned with GitHub/git client conventions; no new user-facing configuration._

### Configuration

_N/A — `config.schema.json` not modified._

### Tests

**Local verification:** `npm test` (1141 passed), `npm run lint`, Prettier on changed files.

## Test plan

- [x] `processUrlPath('/octocat/hello-world/info/refs?service=git-upload-pack')` → `repoPath: /octocat/hello-world.git`
- [x] `processUrlPath('/github.com/octocat/hello-world/info/refs?service=git-upload-pack')` → proxied host path with `.git` normalised
- [x] `processUrlPath('/org/owner/repo/git-upload-pack')` → pack path parsed
- [x] Regression: paths with `.git` unchanged (`/octocat/hello-world.git/info/refs?...`, `/octocat/hello-world.git`, `/octocat/hello-world.git/`)
- [x] `isPackPost` true for `POST /a/b/git-upload-pack` and `POST /a/b.git/git-upload-pack`
- [x] `proxyFilter` allows `GET /github.com/finos/git-proxy/info/refs?service=git-upload-pack` without mocking `processUrlPath`
- [ ] Manual: `curl -H "User-Agent: git/2.43.0" "http://<proxy>/owner/repo/info/refs?service=git-upload-pack"` against authorised repo (both with and without `.git` in path)

### DIFF
--- a/src/proxy/routes/helper.ts
+++ b/src/proxy/routes/helper.ts
@@ -68,6 +68,9 @@ export const processGitUrl = (url: string): GitUrlBreakdown | null => {
  * into the embedded git end point and path for the git operation. */
 const PROXIED_URL_PATH_REGEX = /(.+\.git)(\/.*)?/;
 
+/** Fallback for smart-HTTP paths that omit the .git repo suffix (e.g. GitHub-style URLs). */
+const SMART_HTTP_PATH_REGEX = /^(.+?)(\/(?:info\/refs|git-upload-pack|git-receive-pack)(?:\?.*)?)$/;
+
 /** Type representing a breakdown of paths requested from the proxy server */
 export type UrlPathBreakdown = { repoPath: string; gitPath: string };
 
@@ -81,6 +84,11 @@ export type UrlPathBreakdown = { repoPath: string; gitPath: string };
  * - gitPath: /info/refs?service=git-upload-pack
  *
  * and processing /github.com/finos/git-proxy.git/info/refs?service=git-upload-pack
+ * would produce:
+ * - repoPath: /github.com/finos/git-proxy.git
+ * - gitPath: /info/refs?service=git-upload-pack
+ *
+ * Processing /github.com/finos/git-proxy/info/refs?service=git-upload-pack (no .git suffix)
  * would produce:
  * - repoPath: /github.com/finos/git-proxy.git
  * - gitPath: /info/refs?service=git-upload-pack
@@ -100,10 +108,18 @@ export const processUrlPath = (requestPath: string): UrlPathBreakdown | null => 
       repoPath: components[1],
       gitPath: components[2] ?? '/',
     };
-  } else {
-    console.error(`Failed to parse proxy url path: ${requestPath}`);
-    return null;
   }
+
+  const smartHttp = requestPath.match(SMART_HTTP_PATH_REGEX);
+  if (smartHttp) {
+    return {
+      repoPath: smartHttp[1] + '.git',
+      gitPath: smartHttp[2],
+    };
+  }
+
+  console.error(`Failed to parse proxy url path: ${requestPath}`);
+  return null;
 };
 
 /** Regex used to analyze repo URLs (with protocol and origin) to extract the repository name and
--- a/test/extractRawBody.test.ts
+++ b/test/extractRawBody.test.ts
@@ -91,6 +91,13 @@ describe('isPackPost()', () => {
     expect(isPackPost({ method: 'POST', url: '/a.git/git-upload-pack' } as Request)).toBe(true);
   });
 
+  it('returns true for git-upload-pack POST without a .git repo suffix', () => {
+    expect(isPackPost({ method: 'POST', url: '/a/b/git-upload-pack' } as Request)).toBe(true);
+    expect(
+      isPackPost({ method: 'POST', url: '/github.com/org/repo/git-receive-pack' } as Request),
+    ).toBe(true);
+  });
+
   it('returns false for other URLs', () => {
     expect(isPackPost({ method: 'POST', url: '/info/refs' } as Request)).toBe(false);
   });
--- a/test/testRouteFilter.test.ts
+++ b/test/testRouteFilter.test.ts
@@ -75,6 +75,25 @@ describe('url helpers and filter functions used in the proxy', () => {
     expect(processUrlPath(VERY_LONG_PATH)).toBeNull();
   });
 
+  it('processUrlPath should parse smart-HTTP paths without a .git repo suffix', () => {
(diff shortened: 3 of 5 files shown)


src/proxy/routes/helper.ts
```diff
@@ -100,10 +108,18 @@ export const processUrlPath = (requestPath: string): UrlPathBreakdown | null =>
       repoPath: components[1],
       gitPath: components[2] ?? '/',
     };
-  } else {
-    console.error(`Failed to parse proxy url path: ${requestPath}`);
-    return null;
   }
+
+  const smartHttp = requestPath.match(SMART_HTTP_PATH_REGEX);
```
fabiovincenzi, line 113: The fallback is only reached when PROXIED_URL_PATH_REGEX fails, but that regex is unanchored, and String.match doesn't have to consume the whole string, so it matches almost any path containing .git and silently drops the rest. That includes paths where .git sits inside the repo name:

/github.com/finos/.github/info/refs?service=git-upload-pack
consumed: /github.com/finos/.git
dropped:  hub/info/refs?service=git-upload-pack
→ repoPath "/github.com/finos/.git", gitPath "/"

validGitRequest("/") then rejects it, so the new fallback is dead code for these paths and the user still gets Invalid request received. .github and <user>.github.io show up in most orgs. They work fine today with an explicit .git in the remote, so this isn't a regression, but it's exactly the git-less form this PR is meant to enable.

Fix is to swap the order: the operation suffix is a protocol guarantee. Match the anchored /info/refs, /git-upload-pack or /git-receive-pack suffix against the end of the path, take the prefix as the repo path, and append .git only when it isn't already there, the same normalisation you're doing now, just reached first.

Ran it over the full matrix: fixes .github, user.github.io and the query-string case (…&foo=.git, which currently swallows the whole path into repoPath), and leaves every currently-working input identical. Worth anchoring PROXIED_URL_PATH_REGEX while you're in here too, byte-identical on everything that works today, and it kills a whole class of silent partial matches.

test/testRouteFilter.test.ts
```diff
@@ -75,6 +75,25 @@ describe('url helpers and filter functions used in the proxy', () => {
     expect(processUrlPath(VERY_LONG_PATH)).toBeNull();
   });
 
+  it('processUrlPath should parse smart-HTTP paths without a .git repo suffix', () => {
```
fabiovincenzi, line 78: Could you add the cases that pin the ordering down? Every input here has a repo name with no .git in it, which is why the problem isn't visible:
- /github.com/finos/.github/info/refs?service=git-upload-pack → /github.com/finos/.github.git
 - /github.com/user/user.github.io/git-receive-pack → /github.com/user/user.github.io.git
- /github.com/finos/git-proxy (no git operation) → still null

The last one guards the Action constructor path, which relies on processUrlPath returning null rather than inventing a repo when there's nothing to split on.

test/extractRawBody.test.ts
```diff
@@ -91,6 +91,13 @@ describe('isPackPost()', () => {
     expect(isPackPost({ method: 'POST', url: '/a.git/git-upload-pack' } as Request)).toBe(true);
   });
 
+  it('returns true for git-upload-pack POST without a .git repo suffix', () => {
```
fabiovincenzi, line 94: Could you add the negative that fixes the new boundary? isPackPost({ method: 'POST', url: '/git-upload-pack' }) must stay false, a single segment isn't a repo. It holds today, but nothing here would catch a future simplification of this regex that opens it up.

PR #1656, opened 2026-07-22
feat(db): add dateCreated and lastModified to Repo

sankalpsthakur:
## Summary
- Add optional `dateCreated` / `lastModified` (ISO-8601) on the Repo model
- Stamp both on create; bump `lastModified` on all user-role mutations (NeDB + Mongo)
- **No startup backfill** — existing records without these fields stay as-is until a follow-up migration PR (per @andypols review in #1656)

## Scope / issue linkage
This PR lands the **persistence slice** of #1486 (model fields + create/mutation timestamps). Remaining #1486 work is intentionally split out:
- **sortBy allowlist + UI sort options** — blocked on / coordinated with #1464 server-side pagination
- **Existing-record backfill** — follow-up PR introducing a versioned migration framework; first migration will derive Mongo `dateCreated` from `$toDate("$_id")` rather than defaulting to epoch (as suggested in review)

Unblocks Date Created / Date Modified sorting for newly created and recently mutated repos once #1464 lands.

Related: #1420 (broader admin sorting/pagination — #1464 covers the server half)

## Test plan
- [x] `npm test -- test/db` (163 passed; 45 integration tests skipped because no external MongoDB was configured)
- [x] Create a repo and confirm both dates are set
- [x] Mutate canPush/canAuthorise → `lastModified` advances, `dateCreated` unchanged
- [ ] Sort repo list by Date Created / Date Modified (once #1464 lands)

Fixes #1486

### AI/LLM disclosure

- AI coding tools (including Grok and/or Codex agent-assisted editing) were used to help draft or modify code and this PR description.
- I reviewed the complete change, understand the reasoning, and ran the reported local tests before submitting.
- This submission is original work of authorship under the project CLA / contributor terms; AI output was not pasted unreviewed.

PR #1661, opened 2026-07-28
feat: DB Migrations

andypols:
## Description

This PR provides adds DB migrations.  #1656 adds `dateCreated` and `lastModified` to repository records to support sorting or repository by date. This cannot be used without a corresponding migration of existing data.

Each migration is tracked by its own unique ID, rather than relying on a single global “current database version”. This means two branches can add separate migrations and be merged in either order. When both reach main, GitProxy checks the complete registry and runs whichever migration IDs have not yet been applied.

This gives us:

- a clear history of data changes
- one shared implementation across database backends
- a record of which changes have already run
- safe restart behaviour after a partial migration
- optional rollback support
- a consistent pattern for future schema and data changes

## Example migration

For example a migration for #1656 could be implemented like this.

```ts

import type { Repo } from '../types';
import type { Migration } from './index';

export const addRepoDates: Migration = {
  id: '20260722-add-repo-dates',
  up: async (sink) => { 
    const repos = await sink.getRepos(); 
    for (const repo of repos) {
      if (repo.dateCreated && repo.lastModified) { 
        continue; 
      }

      const dateCreated =
        repo.dateCreated ?? (await sink.deriveCreatedAt(repo._id)) ?? new Date().toISOString();

      await sink.updateRepo({
          ...repo, 
          dateCreated, 
          lastModified: repo.lastModified ?? dateCreated, 
      });
    } 
  }, 
};
```


src/db/types.ts
```diff
@@ -175,4 +176,8 @@ export interface Sink {
   addPublicKey: (username: string, publicKey: PublicKeyRecord) => Promise<void>;
   removePublicKey: (username: string, fingerprint: string) => Promise<void>;
   getPublicKeys: (username: string) => Promise<PublicKeyRecord[]>;
+  getAppliedMigrations: () => Promise<string[]>;
+  recordMigration: (id: string) => Promise<void>;
+  unrecordMigration: (id: string) => Promise<void>;
+  deriveCreatedAt: (id: string) => string | undefined;
```
jescalada, line 182: I'm wondering about the effects of adding these for Denis' Postgres implementation in #1532 - we'll need to implement these separately after merging this PR. Do you think these could be added to #1582? @dcoric

jescalada, line 182: Another note: `deriveCreatedAt` is the only synchronous method here besides `getSessionStore`. It's a migration-specific field that makes sense only for the Mongo migration (NeDB one returns `undefined`). We could have it return `undefined` on the Postgres and future sink implementations, but I feel it'd be cleaner to exclude from the `Sink` entirely...

On the bright side, the `dateCreated` and `lastModified` fields can be easily added to the Postgres PRs 🙂

jescalada (approved):
LGTM so far 🚀 We might want to add a few more tests for better coverage on the file and mongo handlers (`repo.ts` and `migrations.ts`) 👍🏼

src/db/types.ts
andypols: @jescalada I can see your thinking, but how would you work out something like `dateCreated` from an existing row in a generic way without the migration having to ask are you mongo?  Maybe that's ok but feels nicer to hide it in the abstraction?

andypols: I think I am going or leave it as it is and we can always raise a separate PR to remove it later

andypols: @dcoric I'm assuming you are ok with this for your postgres work.

jescalada (approved):

PR #1681, opened 2026-08-20
feat: backfill repo dateCreated/lastModified via migration

andypols:
This PR adds the migration for #1656 adding repo `dateCreated` and `lastModified`  using the migrations #1661

It reads repos through the sink, asks the backend to derive a creation time (Mongo derives it from the ObjectId; the file backend has no source and falls back to now), and writes back through the sink. Repos that already have a dateCreated are skipped, so the migration is safe to re-run.

The UI can now show the repos in order which is really useful when you want to what has been added recently.


jescalada (approved):
LGTM

06kellyjac:
Looks like it needs a rebase

PR #1583, opened 2026-06-08
fix: handle idle pool client errors in the PostgreSQL sink

dcoric:
## Description

> Draft, stacked on #1532 (PostgreSQL sink). The diff below currently includes the #1532 changes. Once #1532 merges, this branch will be rebased onto `main` so only the pool-error-handler change remains, and the draft flag will be removed.

Adds an idle-client error handler to the PostgreSQL connection pool.

`node-postgres` pools emit an `'error'` event when an idle client fails (for example when the database drops the connection). With no listener attached, Node treats this as an uncaught exception and crashes the process. This registers a listener in `ensurePool()` that logs the error and lets the pool recycle the client.

- Unit test verifies the listener is registered and that invoking it logs without throwing.

## Related Issue

Resolves #1562

### General

- [x] I have read the [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) format
- [x] I have a [FINOS CLA](https://finosfoundation.atlassian.net/wiki/spaces/FINOS/pages/75530375/Contribution+Compliance+Requirements) on file

### Documentation

- [ ] Documentation has been added/updated for any new features (internal robustness fix; no user-facing docs needed)

### Configuration

- [ ] If configuration schema (`config.schema.json`) was modified: (not modified in this PR)

### Tests

- [x] Tests have been added/updated for new functionality
- [x] Unit tests pass (`npm test`)
- [x] Linting and formatting pass (`npm run lint` and `npm run format:check`)
- [x] Type checks pass (`npm run check-types`)

PR #1687, opened 2026-08-24
feat: PostgreSQL sink backend

### LINKED ISSUE #1497: Add PostgreSQL as a supported sink backend

dcoric:
Adds PostgreSQL as a supported sink backend, complete with schema migrations, data migration, and production connection/auth options.

This is an integration branch. The PostgreSQL work was split into small, reviewable PRs that build on each other, and merging them into `main` one at a time would leave a half-configured backend in place for days at a time. They target this branch instead, so each keeps its own review and CI, and `main` sees the finished backend in one piece.

### Collected PRs

| PR | Scope | Status |
|----|-------|--------|
| #1532 | PostgreSQL sink backend (users, repos, pushes, session store) | Merged |
| #1581 | Versioned DDL schema migrations with advisory locking | Merged |
| #1582 | Data migration from mongo/fs into PostgreSQL | Merged |
| #1583 | Idle pool client error handling | Merged |
| #1584 | Configurable connection (discrete fields, TLS, pool tuning) | Merged |
| #1590 | Normalise repo permissions into a `repo_users` join table | Merged |
| #1591 | AWS RDS/Aurora IAM authentication | Merged |
| #1692 | Pushes hot-path indexes and slimmer list projections | Merged |
| #1689 | `sink-parity` agent skill and AGENTS.md steering for backend parity | Merged |

All nine pieces have merged and each was reviewed and CI-verified individually against this branch (full pipeline including the PostgreSQL integration lane). The branch carries no drift from `main`. This PR is the complete feature, ready for final review. Related issues: #1688, #1690, #1691.

### What lands

- `postgres` sink selectable via the existing `sink` config, alongside `fs` and `mongo`
- Full `Sink` interface parity with the mongo and NeDB backends, including `getRepoPushRollupsByCanonicalUrl`, `getPushesForUserProfile`, `updateRepo` and the migration hooks
- Versioned, advisory-locked DDL migrations applied on startup
- `migrate-to-postgres` command for existing mongo/fs deployments
- Connection via connection string, discrete fields or `PG*` env vars, with TLS and pool tuning
- Optional RDS/Aurora IAM token auth
- Design decisions recorded in the architecture doc, plus a `sink-parity` skill so future adaptor changes keep all backends aligned
- Unit plus integration coverage, with a dedicated PostgreSQL CI lane

### Note on migrations

`main` has since grown a cross-backend migration framework in `src/db/migrations`, which records logical migrations by string id through `Sink` hooks. That is a different concern from #1581, which versions the postgres DDL itself and serialises concurrent runs with an advisory lock. Both are kept: the postgres adapter implements the framework hooks in `src/db/postgres/migrations.ts`, and the DDL runner lives alongside it in `src/db/postgres/schemaMigrations.ts`.

Existing deployments are unaffected: the default sink remains the filesystem backend and `postgres` is opt-in.


.github/workflows/ci.yml
```diff
@@ -5,9 +5,9 @@ name: CI
 
 on:
   push:
-    branches: [main]
+    branches: [main, feat/postgres]
   pull_request:
-    branches: [main]
+    branches: [main, feat/postgres]
```
jescalada: Don't forget to revert these, we're getting duplicate jobs on the CI 😃 

Will do a full review and AI scan later 👍🏼

dcoric: Good catch, reverted - the triggers are back to main only. They existed so the stacked PRs (based on feat/postgres) got the full pipeline; with everything merged, this PR's own runs are covered by the original filter and the addition was just producing the duplicate runs you saw. The only ci.yml change left in the diff is the intended one: the PostgreSQL service container and integration test lane that ship with the feature.

src/db/postgres/helper.ts
```diff
@@ -0,0 +1,304 @@
...
+        '[postgres] connectionString is set; ignoring the discrete host/port/user/password/database fields',
+      );
+    }
+    config.connectionString = db.connectionString;
+  } else {
+    if (db.host !== undefined) config.host = db.host;
+    if (db.port !== undefined) config.port = db.port;
+    if (db.user !== undefined) config.user = db.user;
+    if (db.password !== undefined) config.password = db.password;
+    if (db.database !== undefined) config.database = db.database;
```
jescalada: Wondering if the `buildPoolConfig` function could be refactored? I noticed a lot of repeated code here, and it could be more readable 🤔

src/db/postgres/helper.ts
```diff
@@ -0,0 +1,304 @@
...
+  try {
+    await client.query('BEGIN');
+    const result = await fn(client);
+    await client.query('COMMIT');
+    return result;
+  } catch (err) {
+    try {
+      await client.query('ROLLBACK');
+    } catch {
+      // the original error is the one worth surfacing
```
jescalada, line 267: Is it okay to leave this empty?

src/db/postgres/repo.ts
```diff
@@ -0,0 +1,250 @@
...
+export const createRepo = async (repo: Repo): Promise<Repo> => {
+  const users = repo.users ?? { canPush: [], canAuthorise: [] };
+  const now = new Date().toISOString();
+  if (!repo.dateCreated) repo.dateCreated = now;
+  if (!repo.lastModified) repo.lastModified = now;
+  const result = await query<{ _id: string }>(
+    `INSERT INTO repos (project, name, url, date_created, last_modified)
+     VALUES ($1, $2, $3, $4, $5)
+     RETURNING _id`,
+    [repo.project ?? '', repo.name, repo.url, repo.dateCreated, repo.lastModified],
```
jescalada: I thought `Repo` was guaranteed to have the `project` attribute - any reason why we we're defaulting to `''` here?

From `/src/db/types`:
```ts
export class Repo {
  project: string;
  name: string;
  url: string;
  users: { canPush: string[]; canAuthorise: string[] };
...
```

src/db/postgres/users.ts
```diff
@@ -0,0 +1,255 @@
...
+    `SELECT ${SELECT_COLUMNS} FROM users WHERE public_keys @> $1::jsonb`,
+    [JSON.stringify([{ key: sshKey }])],
+  );
+  return result.rowCount === 0 ? null : rowToUser(result.rows[0]);
+};
+
+export const addPublicKey = async (username: string, publicKey: PublicKeyRecord): Promise<void> => {
+  const existingUser = await findUserBySSHKey(publicKey.key);
+  if (existingUser && existingUser.username.toLowerCase() !== username.toLowerCase()) {
+    throw new DuplicateSSHKeyError(existingUser.username);
```
jescalada: I think this might be exploitable with two concurrent `addPublicKey` requests. If two users add the same public key simultaneously, they second call might be able to bypass the `findUserBySSHKey` check if the first call didn't complete the DB update.

Very unlikely to happen with real usage, but something that AI scans will probably complain about 😄

src/db/postgres/pushes.ts
jescalada, line 1: AI scan comment on this one:

> `authorise` / `reject` / `cancel` in `pushes.ts`: All three do `getPush` → mutate in memory → `writeAudit`, with no transaction and no row lock, and `writeAudit` overwrites data wholesale via `ON CONFLICT DO UPDATE`. Two reviewers acting on the same push at the same moment produce a straight lost update, and worse, if the proxy writes step results for that push id between the read and the write, those results vanish from the audit record. The fix is cheap since `withTransaction` already exists: wrap the three of them and read with `SELECT data FROM pushes WHERE id = $1 FOR UPDATE`. Mongo has the same shape, so it's not a regression, but Postgres is the backend where you can actually fix it.

I'm not sure if it correctly scanned other files as well: I don't think we're using `withTransaction` outside of `updateRepo` in `/repo.ts`...

Long story short, we should check if it's appropriate to wrap all our `writeAudit`-related calls in `withTransaction` 👍🏼

src/db/postgres/repo.ts
```diff
@@ -0,0 +1,250 @@
...
+    user.toLowerCase(),
+    role,
+  ]);
+  await query(`UPDATE repos SET last_modified = $2 WHERE _id = $1`, [
+    _id,
+    new Date().toISOString(),
+  ]);
+};
+
+export const createRepo = async (repo: Repo): Promise<Repo> => {
```
jescalada, line 154: AI comment on this one:

> `createRepo` isn't atomic. The `INSERT INTO repos` returns an `id`, then the function loops `addUserToRole` once per username, each of which is two more un-transacted round trips. A crash or connection drop partway leaves a repo row with partial or empty `canPush` / `canAuthorise`. Since those arrays are what gate pushing and approving, "created but missing its grants" is a state I'd rather not be able to reach. Same `withTransaction` treatment, and you can collapse the loop into a single insert over `unnest($2::text[])`.

src/db/postgres/repo.ts
```diff
@@ -0,0 +1,250 @@
...
+
+/**
+ * Apply a partial update to a repo row. Only the supplied fields are written,
+ * matching mongo's `$set` / `$unset` behaviour: a field explicitly set to
+ * `undefined` is reset to the column default.
+ *
+ * Permissions live in the `repo_users` join table rather than a column, so a
+ * supplied `users` object replaces that repo's rows wholesale.
+ */
+export const updateRepo = async (repo: Partial<Repo>): Promise<void> => {
```
jescalada, line 190: Another interesting AI comment on this one:

> A design question rather than a bug: `repo_users.username` is plain `TEXT` with no foreign key, and `deleteUser` doesn't touch the table. Delete `alice`, create a new `alice` six months later, and she quietly inherits every grant the old one had. `updateUser` renaming a `username` has the same problem in reverse, orphaning grants. Mongo carries the identical hole, so it's not new, but normalising into a join table is precisely the moment you get to fix it, and leaving it means the normalisation buys structure without buying integrity.

TL;DR: Deleting a user by username and adding it again (with the same username) makes the new one inherit the old user's permissions. Might be worth making an issue for this since it's apparently reproducible in Mongo too 🤷🏼

src/db/postgres/schemaMigrations.ts
jescalada, line 1: Last AI comment:

> On migrations, one operational caveat worth documenting. `connect()` runs `runMigrations` lazily on the first query of any process, so the runtime DB role needs DDL rights permanently, which is awkward in the regulated shops that are the target audience here. And migration 5 drops `repos.users`, so during a rolling deploy any still-running older process breaks the moment a new one boots. I'd add an `autoMigrate: false` escape hatch plus a note in the architecture doc about deploy ordering. Related: `connect-pg-simple` runs with `createTableIfMissing: true`, which is a second DDL path living outside the versioned migration list entirely.

TL;DR: Make auto-migration optional since admin rights might no be enabled by default (especially in banks and the like). An organization would have to enable admin rights for the runtime DB, trigger the migration and then go back to regular rights.

website/docs/architecture/architecture.md
```diff
@@ -484,10 +484,121 @@ Sample values:
...
+
+- All pending migrations run inside a single transaction, so a statement that cannot run transactionally (for example `CREATE INDEX CONCURRENTLY`) is not yet supported by the runner.
+- Repo permissions (`canPush` / `canAuthorise`) are normalised into a `repo_users(repo_id, username, role)` join table (`ON DELETE CASCADE` from `repos`); the adapter reconstructs the permission arrays on read.
+- If `postgres` is selected as the active sink and no connection can be resolved, GitProxy refuses to start rather than silently falling back to an in-memory session store.
+
+##### PostgreSQL design decisions
+
+The adapter follows a few deliberate choices, made for parity with the existing backends rather than for idiomatic SQL:
+
+- **Pushes stay documents.** A push is an audit record: written once, updated through a handful of state flips, and read back whole. The `pushes` table therefore keeps the entire action as a JSONB `data` column, with typed columns (`timestamp`, the status booleans) only for the fields that queries filter and sort on. This mirrors how the mongo and NeDB backends treat pushes and keeps the row shape stable as the `Action` type evolves.
```
jescalada, line 608: I'm wondering about the implications of using JSONB for pushes, or the hybrid thing we have for users and repos. What are the downsides of this? Isn't it considerably harder to understand and maintain JSONB queries? Or is the idea to just be able to port over any Mongo behaviour easily (since we'd be operating with documents rather than rows)? What about efficiency of reading and writing to DB vs normalizing?

I'm worried about the maintainability of JSONB over normalized SQL tables. Although there's also the question of efficiency - if we normalized the DB, we would have to build the Action objects from different tables via queries 🤷🏼

jescalada (changes requested):
@dcoric Thanks for this! My main question is the JSONB design choice: I'd like to know why you chose this beyond symmetry with Mongo/NeDB. What are the pros and cons of JSONB over normalization, and most importantly, which gives faster reads & writes? 

@finos/git-proxy-maintainers I've reviewed all of the smaller component PRs previously, so this is a final pass + AI scan for the whole PR. I'd appreciate an extra look! 😃

src/db/postgres/pushes.ts
dcoric: Agreed — fixed in 61cb0382. `authorise`/`reject`/`cancel` now read the row with `SELECT ... FOR UPDATE` inside `withTransaction` and write through the same client, so two concurrent decisions (or a step-result write racing a reviewer) serialise instead of the later write discarding the earlier one. Also added an integration test running `authorise` and `reject` concurrently against the same push.

On the wider question: `writeAudit` itself stays un-wrapped — a single upsert is already atomic, and the proxy is the only writer while a push is in flight. It's the read-modify-write callers that needed the lock, and those three were the only ones in the postgres adapter.

src/db/postgres/repo.ts
dcoric: Fixed in 61cb0382, exactly as the scan suggested: the repo insert and its permission grants run in one `withTransaction`, and the per-user loop is collapsed into a single `INSERT ... SELECT ... FROM unnest($3::text[])`. The same set-based insert is now shared with `updateRepo`'s wholesale permission replacement, which had the same loop.

src/db/postgres/users.ts
dcoric: Good catch — fixed in 61cb0382. `addPublicKey` now runs in a transaction that first takes `pg_advisory_xact_lock(hashtextextended(key, 0))`: concurrent adds of the same key serialise on the lock and the loser then sees the winner's row in the duplicate check. (A unique constraint can't span elements of a JSONB array, hence the advisory lock rather than a schema change.) The target user row is also read `FOR UPDATE` for the fingerprint check. Added a deterministic integration test racing two users adding the same key — exactly one wins, the other gets `DuplicateSSHKeyError`.

Mongo carries the identical race; left out of scope here since it needs a different mechanism there.

src/db/postgres/schemaMigrations.ts
dcoric: Implemented in 435c2329: new `autoMigrate` option on the postgres sink (default `true`, so behaviour is unchanged unless opted into). With `autoMigrate: false` startup performs no DDL at all — it only verifies the schema is current and refuses to start with the pending versions named. Migrations are applied out-of-band with DDL-capable credentials via the new `npm run migrate:postgres:schema` script (same advisory-locked runner, so it's safe against concurrent runs).

Also closed the second DDL path you flagged: the `connect-pg-simple` session table is now owned by migration 7 (`IF NOT EXISTS` adopts databases where the store already created it) and `createTableIfMissing` is switched off, so every piece of DDL flows through the versioned list. The architecture doc gained sections on `autoMigrate` and deploy ordering, including the migration-5 rolling-deploy caveat.

src/db/postgres/repo.ts
dcoric: Agreed on filing an issue — opened #1705 covering all three backends. Fixing only postgres here would make the backends diverge behaviourally (mongo/fs carry the identical hole via the usernames embedded in `repos.users`), and keeping the sinks at parity is the whole point of this PR, so the cleanup belongs in a cross-backend change.

src/db/postgres/helper.ts
dcoric: Refactored in 435c2329: the duplicated discrete-field copying is now `applyDiscreteFields` (with an `includePassword` switch for IAM mode, where a token provider replaces the password), and the SSL and pool-tuning blocks moved into `applySsl` / `applyPoolTuning`. `buildPoolConfig` itself is left as a readable three-way branch over the connection modes. Behaviour is identical — the existing unit tests over warnings and precedence pass unchanged.

src/db/postgres/helper.ts
dcoric: Intentional — that's the catch around `ROLLBACK` inside `withTransaction`. If rollback itself fails, the connection is already broken: the error worth surfacing is the original one from the transaction body, which is rethrown on the next line, and `finally` releases the client so the pool discards it. The comment inside the block says as much. Happy to add a debug log there if you'd prefer the rollback failure visible too.

src/db/postgres/repo.ts
dcoric: The `Repo` class declares `project: string`, but what actually reaches the sink at runtime are plain objects from the HTTP handlers and tests, not validated class instances — and the mongo/fs backends silently accept a missing `project` today. The column is `TEXT NOT NULL DEFAULT ''`, so `?? ''` makes postgres accept the same inputs the other backends do rather than throwing a not-null violation only on this sink. Tightening it to a thrown error would be a cross-backend behaviour change, which I'd rather not smuggle into this PR.

website/docs/architecture/architecture.md
dcoric: Expanded this section in 435c2329 with a dedicated "JSONB over full normalisation is deliberate" bullet carrying the reasoning (and fixed the stale claim that repo permissions are still JSONB — they're normalised now). Fuller answer on the main review thread.

dcoric:
@jescalada Thanks for the thorough review! On the main question — JSONB vs normalization — this was discussed near the end of the Aug 26 community call, where we settled on storing documents as JSON in Postgres and avoiding full normalization, to keep maintenance simple and keep the sinks easy to hold at feature parity. The reasoning, now also written into the architecture doc's design-decisions section:

**Access patterns.** Pushes are written and read back *whole*, by id, and listed by timestamp. Nothing queries inside the document relationally — the handful of filtered fields (`error`, `blocked`, `authorised`, `canceled`, `rejected`, `timestamp`) are already promoted to real columns, and the two JSONB lookups the profile/activity pages make are covered by expression indexes (migration 6). Fully normalizing an `Action` would decompose a deeply nested document (steps, commit data, attestation) across many tables on every write and reassemble it with multi-way joins on every read, purely to serve queries the app never runs.

**Reads and writes.** A read is a single-row fetch by primary key either way, but JSONB skips the multi-join reassembly; a write is one upsert instead of a transactional multi-table write. For this workload — low write volume, document-shaped data — JSONB is at least as fast in both directions. Maintainability cuts the same way: the adapter stays a thin mapping, and a new `Action` field costs nothing here instead of a schema migration + insert/select/mapping change per table.

**Where normalization pays, we did it.** Repo permissions *are* queried relationally ("who can push where"), so they were normalized into `repo_users` (#1559 — this was one of #1497's open questions, resolved exactly this way). Users/repos are deliberately hybrid: real columns for everything queried or constrained (`username`, `email`, `oidc_id`, `admin`), JSONB only for the `publicKeys` blob. So the rule is: JSONB only where the data is genuinely a document.

**Parity and portability.** All three backends operating on the same document shapes is what makes the sink-parity contract tractable, and it's why `migrate:postgres` can move data between sinks without lossy transformation.

---

All the inline findings are now addressed on the branch:

- **61cb0382** — `authorise`/`reject`/`cancel` run in transactions with `FOR UPDATE`; `createRepo` is atomic with a set-based `unnest()` grant insert; `addPublicKey` closes its TOCTOU race with a transaction-scoped advisory lock. New unit + deterministic integration tests for the races.
- **435c2329** — `autoMigrate: false` escape hatch for DDL-restricted deployments (+ `npm run migrate:postgres:schema` for out-of-band runs), the session table folded into the versioned migration list (no more `createTableIfMissing` side channel), `buildPoolConfig` deduplicated, and the architecture doc updated with the JSONB rationale, `autoMigrate`, and deploy-ordering notes.
- The username-reuse/grant-inheritance hole is filed as #1705 — it's identical in mongo/fs, so it belongs in a cross-backend fix rather than this PR.

jescalada (approved):
LGTM, and thanks for the explanation on the JSONB decision 👍🏼 

@finos/git-proxy-maintainers Would appreciate an external review on this before merging and shipping 2.3!

andypols (approved):
LGTM.

src/config/generated/config.ts
```diff
@@ -554,6 +566,78 @@ export interface Database {
...
+  /**
+   * Enable IAM token authentication for the PostgreSQL connection.
+   */
+  enabled: boolean;
+  /**
+   * AWS region of the RDS/Aurora instance. Falls back to the `AWS_REGION` /
+   * `AWS_DEFAULT_REGION` environment variables, then the AWS SDK's default region resolution.
+   */
+  region?: string;
   [property: string]: any;
```
kriswest: You can get rid of these by setting `additionalProperties: false` in the schema. Usually a good idea as it forces you to add necessary properties to the schema rather than allowing arbitrary properties to be valid (and potentially required by not typed as such)

src/config/types.ts
```diff
@@ -24,6 +24,7 @@ export type ServerConfig = {
   GIT_PROXY_HTTPS_UI_PORT: string | undefined;
   GIT_PROXY_COOKIE_SECRET: string | undefined;
   GIT_PROXY_MONGO_CONNECTION_STRING: string;
+  GIT_PROXY_POSTGRES_CONNECTION_STRING: string | undefined;
```
kriswest, line 27: A default for this is set - should this be just typed as `string` like the mongo string?

src/db/postgres/helper.ts
```diff
@@ -0,0 +1,343 @@
...
+  }
+
+  return async () => {
+    const { Signer } = await loadRdsSigner();
+    const signer = new Signer({ hostname: host, port, username: user, region });
+    return signer.getAuthToken();
+  };
+};
+
+const hasDiscreteFields = (db: DatabaseConfig): boolean =>
```
kriswest, line 98:
```suggestion
/** 
 *  Used to test for *any* discrete connection config fields being set, which are ignored if a conneciton string is set. */
const hasDiscreteFields = (db: DatabaseConfig): boolean =>
```

src/db/postgres/migrate.ts
```diff
@@ -0,0 +1,127 @@
...
+  for (const user of users) {
+    const existing =
+      (await destination.findUser(user.username)) ||
+      (user.email ? await destination.findUserByEmail(user.email) : null);
+    if (existing) {
+      summary.users.skipped++;
+      continue;
+    }
+    // Legacy documents can lack optional fields the writers dereference:
+    // users synced from AD may have no email (the mail attribute is
```
kriswest, line 95: Is the email attribute optional? We use it to attribute pushes when falling back to author details... Doesn't matter for reviewers, but thought it was required for all...

src/db/postgres/migrate.ts
```diff
@@ -0,0 +1,127 @@
...
+    // optional) or gitAccount. Default them like the mongo upsert path does.
+    await destination.createUser({
+      ...user,
+      email: user.email ?? '',
+      gitAccount: user.gitAccount ?? '',
```
kriswest, line 100: Will conflict with one of the CVE resolutions in progress that will replace this field with scmIdentities

src/db/postgres/users.ts
```diff
@@ -0,0 +1,274 @@
...
+  const values: unknown[] = [];
+  const set = (column: string, value: unknown) => {
+    columns.push(column);
+    values.push(value);
+  };
+
+  if (username !== undefined) set('username', username);
+  if (email !== undefined) set('email', email);
+  if (user.password !== undefined) set('password', user.password);
+  if (user.gitAccount !== undefined) set('git_account', user.gitAccount);
```
kriswest, line 150: Expect conflict with incoming CVE resolution PR already noted

src/config/index.ts
```diff
@@ -329,6 +329,10 @@ export const getDatabase = () => {
       if (db.type === 'mongo' && !db.connectionString) {
         db.connectionString = serverConfig.GIT_PROXY_MONGO_CONNECTION_STRING;
       }
+      // same fallback for postgres
+      if (db.type === 'postgres' && !db.connectionString) {
+        db.connectionString = serverConfig.GIT_PROXY_POSTGRES_CONNECTION_STRING;
+      }
```
kriswest: I think both this fallback and the mongo fallback are wrong... As defaults are set in the default config they will always apply unless overridden with a falsey value...

All of the other configs with env var handling prefer the env var over the config file. That behavior should be standardized

website/docs/architecture/architecture.md
```diff
@@ -484,10 +484,137 @@ Sample values:
...
+    {
+      "type": "postgres",
+      "connectionString": "postgresql://user:pass@host:5432/gitproxy",
+      "enabled": true
+    }
+  ]
+}
+```
+
+If `connectionString` is omitted on the config entry, GitProxy falls back to the `GIT_PROXY_POSTGRES_CONNECTION_STRING` environment variable. This mirrors the behaviour of the mongo backend's `GIT_PROXY_MONGO_CONNECTION_STRING`.
```
kriswest, line 511: There's a default in proxy.config.json that it'll fall back to. Hence, does the environment variable ever actually override the config? I think you'd have to override the default and explicitly set a falsey value for the env var to apply.

I think this is the opposite for all other supported env var overrides, except for the mongo connection string, which I think is also incorrectly handled.

config.schema.json
```diff
@@ -570,6 +570,81 @@
             "enabled": { "type": "boolean" }
           },
           "required": ["type", "enabled"]
+        },
+        {
+          "type": "object",
+          "name": "PostgreSQL Config",
+          "description": "Connection properties for PostgreSQL. The `connectionString` may also be supplied via the `GIT_PROXY_POSTGRES_CONNECTION_STRING` environment variable. If neither a `connectionString` nor the discrete `host`/`port`/`user`/`password`/`database` fields are set, the standard `PGHOST`/`PGPORT`/`PGUSER`/`PGPASSWORD`/`PGDATABASE` environment variables are used.",
```
kriswest, line 577: May need updating is the env var handling is corrected

config.schema.json
```diff
@@ -570,6 +570,81 @@
...
+          "properties": {
+            "type": { "type": "string", "const": "postgres" },
+            "enabled": { "type": "boolean" },
+            "connectionString": {
+              "type": "string",
+              "description": "PostgreSQL client connection string, see [https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING). If omitted, `GIT_PROXY_POSTGRES_CONNECTION_STRING` is used as a fallback, then the discrete fields below, then the `PG*` environment variables. Takes precedence over the discrete fields when set."
```
kriswest, line 583: May need updating is the env var handling is corrected

config.schema.json
```diff
@@ -570,6 +570,81 @@
             "enabled": { "type": "boolean" }
           },
           "required": ["type", "enabled"]
+        },
+        {
```
kriswest, line 574: add '"additionalProperties": false` to each finite object to correct types so that they don't allow arbitrary additional properties (necessary for type checking to pick up types - otherwise they just look like a variant on property name and are allowed).

kriswest (commented):
Looks pretty good - a few inline comments to look at.

One thing I think is off is the ENV var handling - I think its also wrong for mongo DB connection string which probably casued that. Every other ENV var overrides config I believe - which makes sense as we set defaults in the default config, so those are really the fallback: Env var -> user config -> default config

Expect conflicts with the CVE resolution thats also refactoring gitAccount into scmIdentities.

kriswest:
@dcoric looks like I broke the CI with the conflict resolution. Will need the lock file regenerating with `npm install` to try again

src/config/generated/config.ts
dcoric: Done for the postgres objects: the sink entry, `pool` and `awsIamAuth` now set `additionalProperties: false`, which removes the index signatures from the generated `Pool` and `AwsIamAuth` types. This one remains because quicktype merges all three sink variants into a single `Database` interface and the mongo and fs entries are still open; I'll close those alongside the mongo env var follow-up. `ssl` and mongo `options` stay open on purpose, since they pass arbitrary driver options through.

src/config/types.ts
dcoric: I'd keep `string | undefined`. The postgres default lives in the default config rather than `env.ts`, and the env var is now purely an override that's absent unless set, like the other `GIT_PROXY_*` server variables. Mongo is `string` only because `env.ts` hardcodes a default; that's part of the mongo precedence issue, which I'll fix in a separate PR against main so this one doesn't change existing mongo deployments.

src/db/postgres/helper.ts
dcoric: Applied, with the typo fixed.

src/db/postgres/migrate.ts
dcoric: `User` types it as required, but users provisioned via ActiveDirectory take it from the optional `mail` attribute, and mongo and fs already accept such users. Postgres keeps that parity (nullable email with a partial unique index), and the migration defaults a missing value to `''` like the mongo upsert path. Such users can't be matched by email for push attribution, which is how every backend behaves today. Happy to tighten it across all backends in a follow-up.

src/db/postgres/migrate.ts
dcoric: Noted, thanks. If the scmIdentities change reaches main first, I'll sync it into feat/postgres and port it to the postgres adapter and the data migration.

src/db/postgres/users.ts
dcoric: Same here, I'll handle it together with the migration when it lands.

src/config/index.ts
dcoric: Agreed, thanks for spotting it. Postgres now follows env var, then user config, then default config, with tests for both directions. This also fixes the schema-migration docs, which suggest the env var for elevated credentials; the old fallback-only handling ignored it whenever the config had a connection string. I've kept mongo out of this PR because it changes behaviour for existing deployments, and will send a separate PR against main (including removing the hardcoded default in `env.ts`) with a release note.

website/docs/architecture/architecture.md
dcoric: Right, the default in `proxy.config.json` meant the env var could never apply. Doc updated here and in the Connection options section.

config.schema.json
dcoric: Updated to describe the env var as an override when set.

config.schema.json
dcoric: Same, updated to match the new precedence.

config.schema.json
dcoric: Done for the finite postgres objects (the entry, `pool`, `awsIamAuth`). `ssl` stays open because it accepts any Node TLS option that pg passes through. See the generated-types thread about the merged `Database` interface.

dcoric:
No worries, done: I regenerated it with npm 10.9.8, which is what CI uses on node 22. A plain `npm install` on npm 11 drops two nested entries that npm 10 requires (`@aws-sdk/rds-signer/node_modules/@aws-sdk/credential-providers` and `vite-tsconfig-paths/node_modules/typescript`), and that's what broke `npm ci`. main's lock has the same gap (`typescript@5.9.3`).

Once installs work, expect one test still failing, and it fails on main too: the push-id scoping test from the 1074b9bc fork merge uses a fake parent commit id (`'456'`), which the commit-id validation from 2ae671ac rejects as "Invalid commit ID format". I've left it alone here since it belongs with those fixes.

PR #1689, opened 2026-08-24
feat: add sink-parity agent skill

### LINKED ISSUE #1688: Add an agent skill that keeps the sink backends at feature parity
GitProxy now has three sink backends implementing the `Sink` interface in `src/db/types.ts`: `fs` (NeDB), `mongo`, and `postgres` (in progress via #1687). Every change to one backend is a parity obligation on the others, and recent history shows how easily that slips:

- New `Sink` members (`getRepoPushRollupsByCanonicalUrl`, `updateRepo`, the migration hooks) had to be retrofitted onto the postgres adapter after landing for mongo/fs.
- New `Repo` fields (`dateCreated` / `lastModified`, #1656) were stored and bumped by mongo/fs but silently dropped by the postgres adapter's update allowlist, which would have made the #1681 startup migration abort boot on postgres.

AI agents do a growing share of this work, so the guardrail should live where agents look: a skill in `.agents/skills/` (exposed to Claude via the `.claude/skills` symlink and to opencode via a command stub) that encodes the parity contract and checklist:

- `src/db/types.ts` is the contract: any new member or field must be implemented in `src/db/file`, `src/db/mongo`, and `src/db/postgres` in the same change.
- Parity is semantic, not just structural: same defaults on create, same timestamp bumps, same projections and filters, same error behaviour, matching case sensitivity.
- Column allowlists must not silently drop new fields; postgres schema changes are append-only migrations.
- Per-backend unit tests plus the verification commands that catch drift (`npm run check-types:server`, backend test suites, the postgres integration lane).

Acceptance: `sink-parity` skill file plus opencode stub on the `feat/postgres` integration branch, applicable to any agent that reads `AGENTS.md`.

dcoric:
Adds a `sink-parity` skill for AI agents working on the sink backends.

The skill encodes the parity contract between `src/db/file`, `src/db/mongo` and `src/db/postgres`: the `Sink` interface in `src/db/types.ts` as the single source of truth, step-by-step checklists for adding interface members, entity fields and postgres schema changes, the semantic rules the compiler cannot enforce (defaults, timestamp bumps, projections, case handling, error behaviour), and the verification commands that catch drift.

It is written for any agent that reads `AGENTS.md`: Claude picks it up through the existing `.claude/skills` symlink and opencode through the added command stub, following the same layout as the existing skills.

Targeted at the `feat/postgres` integration branch since the parity obligation becomes three-way once the PostgreSQL sink lands.

Resolves #1688


.agents/skills/sink-parity/SKILL.md
```diff
@@ -0,0 +1,63 @@
...
+
+## The contract
+
+- `src/db/types.ts` is the single source of truth. A new `Sink` member or entity field is not done until all three backends implement it in the same change; do not leave a backend behind for a follow-up.
+- `npm run check-types:server` enforces the interface structurally, but it cannot see semantic drift. The rest of this checklist exists for what the compiler cannot catch.
+
+## Adding or changing a Sink member
+
+1. Add the member to the `Sink` interface with a doc comment stating its semantics (ordering, case sensitivity, empty-result shape).
+2. Implement it in `src/db/file`, `src/db/mongo`, and `src/db/postgres`. Use the mongo implementation as the reference for behaviour unless the doc comment says otherwise.
```
jescalada: Is the assumption here that usually a Mongo implementation will be added first (and the others must be filled in)? I feel this could be misleading if the agent takes it too literally...

.agents/skills/sink-parity/SKILL.md
```diff
@@ -0,0 +1,63 @@
...
+## Verify before pushing
+
+```
+npm run check-types:server
+cross-env NODE_ENV=test npx vitest --run test/db
+npm run lint
+npm run format:check
+```
+
+Postgres integration tests (`npm run test:integration:postgres`) need a reachable PostgreSQL database; CI runs them in the dedicated lane.
```
jescalada, line 69: I think it'd be good to add a small testing section too:
- Explain how units vs integration tests are set up, and tell it to add both types for each adaptor
- Tell it to make sure the operations performed in each test are identical across sinks (unless N/A)
- Tell it to make sure the outputs of each test are identical across sinks

.agents/skills/sink-parity/SKILL.md
jescalada: I'm wondering about what the "source of truth" should even be... The Mongo implementation is by default the "official" one, but it doesn't mean it'll always be the one to use as a template, especially if new Postgres functionality is added.

jescalada (commented):
LGTM, just a few suggestions for agent output quality 👍🏼

.agents/skills/sink-parity/SKILL.md
dcoric: Agreed, that wording was misleading. Reworded to be contract-first: the doc comment on the interface member defines the behaviour, whichever backend implements it first sets the semantics and records them there, and the others match the comment rather than any particular backend.

.agents/skills/sink-parity/SKILL.md
dcoric: Added a Tests section covering exactly this: unit versus integration setup per backend, both kinds required for a touched backend (fs has no external service, so unit-only there), and aligned scenarios with the same operations and the same asserted outputs across sinks so parity is proven by the suite.

.agents/skills/sink-parity/SKILL.md
dcoric: Clarified in the contract bullet: src/db/types.ts plus its doc comments are the source of truth, and no backend implementation, mongo included, outranks the contract. Mongo being first historically made it the de facto template, but the skill no longer encodes that assumption.

Issue #1691, opened 2026-08-24
Reduce stored push document size and rollup scan cost across sink backends


dcoric:
Discussion issue, applies to every sink backend (fs, mongo, and the PostgreSQL sink proposed in #1687).

Every push row stores the complete diff twice. `getDiff` writes it to both `steps[].logs` (via `step.log(diff)`) and `steps[].content` (via `step.setContent(diff)`), and `writeAudit` persists the whole action document as-is in all backends. Large pushes therefore produce multi-megabyte rows, and everything that touches push documents pays for it:

- The repo activity rollup scans all push rows on every repos page. Indexing can make the scan cheap on PostgreSQL, but it remains O(number of pushes) on every backend, and on mongo/fs it reads the full documents.
- List endpoints and exports ship or filter over fat documents.
- Storage grows with diff size rather than with metadata size.

Ideas worth weighing, none of which are backend-specific:

1. Store the diff once, not twice: drop the duplicate from `steps[].logs` or make `setContent` and the log reference one value.
2. Move diff content out of the push document into its own store or table, loaded only by the push-detail view.
3. A retention or truncation policy for step logs and diff content on old, settled pushes.
4. Maintain the repo activity rollup incrementally (a counters document or table updated on push transitions) instead of recomputing it by scanning all pushes.

Filing as a design discussion before any implementation; the immediate, index-level mitigations for the PostgreSQL sink are tracked separately.

PR #1692, opened 2026-08-24
perf: index the PostgreSQL pushes hot paths and slim list projections

### LINKED ISSUE #1690: Index the pushes hot paths and align list projections in the PostgreSQL sink
Follow-up to the PostgreSQL sink (#1532, integration in #1687), prompted by the review remark on #1532 that the pushes table is the scale risk.

Push rows are large: the stored action carries the full diff inside `steps` (both `steps[].logs` and `steps[].content`), so any query that reads the `data` JSONB detoasts megabyte rows. Three hot paths currently do that on the PostgreSQL sink:

1. `getRepoPushRollupsByCanonicalUrl` runs on every repos page and selects `data->>'url'` from every push row: a full scan that detoasts every row.
2. `getPushes` (dashboard list) returns the full `data` document, diffs included. The mongo backend projects a field allowlist that excludes `steps`, so this is also a backend parity gap.
3. `getPushesForUserProfile` filters on JSONB expressions (`data->>'userEmail'`, reviewer username) with no index support.

Proposed, all append-only:

- A covering expression index so the rollup becomes an index-only scan that never touches the heap: `ON pushes ((data->>'url'), timestamp) INCLUDE (error, rejected, canceled, authorised, blocked, allow_push) WHERE type = 'push'`.
- A partial index matching the default pending-list query: `ON pushes (timestamp DESC) WHERE type = 'push' AND blocked AND NOT error AND NOT authorised AND NOT allow_push`.
- Expression indexes for the user-profile predicates.
- List queries select `data - 'steps'`, matching the mongo projection and keeping diffs out of list responses.

Acceptance: schema migration plus query changes on the PostgreSQL sink branch family, with unit coverage. The broader question of shrinking the stored push document itself is tracked separately.

dcoric:
Indexes the PostgreSQL pushes hot paths and aligns list projections with the mongo backend. Extension of #1532 and #1581, stacked on the schema migrations branch; prompted by the review remark on #1532 that the pushes table is the scale concern.

Push rows carry the full diff inside `steps`, so any query reading the `data` JSONB detoasts very large rows. This PR:

- adds schema migration v6 `pushes_hot_path_indexes`: a covering expression index that turns the repo activity rollup into an index-only scan (it previously detoasted every push row on every repos page), a partial index matching the default pending-review dashboard query, and expression indexes for the user profile predicates
- changes list queries (`getPushes`, `getPushesForUserProfile`) to select `data - 'steps'`, matching the mongo backend's list projection, which also excludes `steps`; the push detail view (`getPush`) still returns the full document
- reserves migration versions 4 and 5 for the `repo_users` normalisation branch (#1590), documented in the registry; the runner applies by version order so the gap is harmless

The wider question of shrinking the stored push document itself (the diff is currently stored twice per push, in `steps[].logs` and `steps[].content`, across all backends) is deliberately out of scope here and tracked in #1691.

Like the other stacked PRs, the diff shown includes the parent branches until they merge into `feat/postgres`; this PR's own change is the last commit.

Resolves #1690


jescalada (approved):
LGTM after fixing conflicts 🚀

PR #1701, opened 2026-08-26
refactor: store diff in action object (#1693)

### LINKED ISSUE #1693: Refactor `getDiff` to store the diff into the action

Right now, `getDiff` stores the diff as content in a step via `step.setContent`, making the diff harder to fetch from plugins (as of the plugin revamp #1683) and subsequent actions like `scanDiff`:

```ts
...
    step.log(`Executing "git diff ${commitFrom} ${action.commitTo}" in ${path}`);
    const revisionRange = `${commitFrom}..${action.commitTo}`;
    const diff = await git.diff([revisionRange]);
    step.log(diff);
    step.setContent(diff);
  } catch (error: unknown) {
    const msg = getErrorMessage(error);
    step.setError(msg);
  } finally {
    action.addStep(step);
  }
  return action;
```

Refactor `getDiff` so that adding the diff as content isn't necessary. We might need to look into *why* it was done this way in the first place. If we need to keep it as step content, we can just store the diff regardless via `action.diff = diff`, and then refactor any usages to use that instead.

**Additional context**
Related plugin revamp: #1683

amnkarn:
Fixes #1693

### Description
This PR refactors `getDiff` and related consumers to store and access the git diff directly on the `action` object (`action.diff`). This makes it easier for plugins and subsequent actions (like `scanDiff`) to retrieve the diff without needing to extract it from the `step` content.

### Changes Made
- **Backend:** Updated `getDiff` to assign the diff string directly to `action.diff` and updated the `Action` interface.
- **Processors:** Updated `scanDiff.ts` to retrieve the diff from `action.diff`.
- **UI:** Updated `PushActionView` in `types.ts` to accept `Step | string` for the diff property.
- **UI Services:** Modified `getPush` in `git-push.ts` to handle the new string format.

### Backward Compatibility
To ensure historical database records and existing UI components do not break, fallback logic has been added. If `action.diff` is not present (or is not a string), the system gracefully falls back to extracting the content from `steps.find((s) => s.stepName === 'diff')?.content` in both `scanDiff.ts` and `git-push.ts`.


src/ui/types.ts
```diff
@@ -54,7 +54,7 @@ export interface BackendResponse {
 }
 
 export interface PushActionView extends Omit<Action, ActionMethods> {
-  diff: Step;
+  diff: Step | string;
```
dcoric: **[Blocking]** This introduces a TypeScript compile error:

```
src/ui/types.ts(56,18): error TS2430: Interface 'PushActionView' incorrectly extends interface 'Omit<Action, ActionMethods>'.
  Types of property 'diff' are incompatible.
    Type 'string | Step' is not assignable to type 'string | undefined'.
      Type 'Step' is not assignable to type 'string'.
```

`Action.diff` added in this PR is `string | undefined`, so this declaration no longer satisfies the base interface. Current PR CI doesn't run `check-types`, so it slips through, but it breaks `npm run check-types` and every contributor's IDE.

Suggested fix: `diff: string | Step | undefined;`. That is also more honest, since records without a `diff` step (e.g. tag pushes) are already `undefined` at runtime today.

src/proxy/processors/push-action/getDiff.ts
```diff
@@ -49,6 +49,7 @@ const exec = async (_req: Request, action: Action): Promise<Action> => {
     step.log(`Executing "git diff ${commitFrom} ${action.commitTo}" in ${path}`);
     const revisionRange = `${commitFrom}..${action.commitTo}`;
     const diff = await git.diff([revisionRange]);
+    action.diff = diff;
     step.log(diff);
     step.setContent(diff);
```
dcoric, line 54: The diff is now persisted in three places per push: `action.diff`, step `content`, and step `logs`. All three get serialized into the audit DB on every `writeAudit` (NeDB and Mongo). NeDB's `getPushes` has no projection, so the pushes-list payload also grows by another full-diff copy per push.

Now that all consumers read `action.diff` (the legacy fallback is only needed for old records), could we drop `step.setContent(diff)` and the full-diff `step.log(diff)` here and keep just a summary (e.g. diff size)? If you'd rather keep this PR minimal, let's at least note the extra storage copy in the description and track the cleanup as a follow-up.

src/proxy/processors/push-action/scanDiff.ts
```diff
@@ -179,11 +179,10 @@ const exec = async (_req: Request, action: Action): Promise<Action> => {
 
   const { steps, commitFrom, commitTo } = action;
   step.log(`Scanning diff: ${commitFrom}:${commitTo}`);
+  const diff = action.diff ?? steps.find((s) => s.stepName === 'diff')?.content;
```
dcoric, line 182: Two test gaps worth closing:

1. Precedence is untested. If `action.diff` and the `diff` step content ever diverge, this silently prefers `action.diff`. A test pinning that precedence would lock the semantics in (the legacy fallback itself is already covered by the existing scanDiff tests).
2. `getPush` in `test/ui/git-push.test.ts` is only mocked with the legacy step shape, so the new `typeof data.diff == 'string'` branch has no coverage. A test with a top-level string `diff` (including the empty-string case, which should still win over the fallback) would help.

src/ui/services/git-push.ts
```diff
@@ -30,7 +30,10 @@ const getPush = async (id: string): Promise<ServiceResult<PushActionView>> => {
     const data: Action = response.data;
     const actionView: PushActionView = {
       ...data,
-      diff: data.steps.find((x: Step) => x.stepName === 'diff')!,
+      diff:
+        typeof data.diff == 'string'
```
dcoric: Nit: use `===` over `==` to match codebase style.

One more expectation to manage: the plugin benefit from #1693 won't materialize from this change alone. Plugins are prepended to the chain, so they run before `getDiff` and can't read `action.diff` in the same pass. Real plugin access needs post-diff plugin positioning (part of the #1683 revamp scope). Worth softening that point in the PR description.

dcoric (changes requested):
Thanks for picking this up @amnkarn, I verified the consumer coverage end-to-end (scanDiff, getPush, PushDetails, Cypress mocks) and the legacy fallbacks keep old DB records working, so the compat approach is right. Requesting changes on one compile-level issue plus a few design/test follow-ups:

- `PushActionView` type breaks `tsc` (inline, blocking)
- diff now persisted 3x per push (inline)
- two test gaps (inline)
- plugin access expectation (inline)

Also: the -40 lines in `package-lock.json` (removal of `"extraneous": true` entries) look like an unrelated npm-prune side effect from a local install. Could you revert them or move them to a separate PR so the diff stays clean?

amnkarn:
Thanks for the thorough review @dcoric!

I have addressed all the feedback:

    1. **`PushActionView` interface fix**: Updated `PushActionView`
  to `Omit<Action, ActionMethods | 'diff'>` with `diff?: string |
  Step;`. `npm run check-types` passes clean now.
    2. **Strict equality**: Updated `typeof data.diff === 'string'`
  in `git-push.ts`.
    3. **Precedence test**: Added unit test in
  `test/processors/scanDiff.test.ts` ensuring `action.diff` takes
  precedence over step content if both exist.
    4. **`getPush` test coverage**: Added unit tests in 
  `test/ui/git-push.test.ts` covering top-level string `diff` and
  empty string `diff`.
    5. **Cleaned `package-lock.json`**: Reverted the unrelated
  `package-lock.json` changes.
    6. **Storage/Audit log note & Plugin scope**: Noted in the PR
  description that `step.setContent(diff)` will be evaluated for
  complete removal in a follow-up once full legacy DB record
  migration/deprecation is done, and clarified that plugin benefit
  materialization will occur under #1683 once post-diff plugin
  positioning is in place.

Issue #1702, opened 2026-08-26
26th August 2026 - GitProxy Roadmap meeting Minutes


kriswest:
## Date

20260826 - 4pm BST / 11am EDT

## Meeting info

- [Meeting link](https://zoom-lfx.platform.linuxfoundation.org/meeting/95173610154?password=42652580-53eb-40b7-9535-ed4d49ee534e)

## Meeting notices

- FINOS **Project leads** are responsible for observing the FINOS guidelines for [running project meetings](https://community.finos.org/docs/governance/meeting-procedures/). Project maintainers can find additional resources in the [FINOS Maintainers Cheatsheet](https://community.finos.org/docs/finos-maintainers-cheatsheet).

- **All participants** in FINOS project meetings are subject to the [LF Antitrust Policy](https://www.linuxfoundation.org/antitrust-policy/), the [FINOS Community Code of Conduct](https://community.finos.org/docs/governance/code-of-conduct) and all other [FINOS policies](https://community.finos.org/docs/governance/#policies).

- FINOS meetings involve participation by industry competitors, and it is the intention of FINOS and the Linux Foundation to conduct all of its activities in accordance with applicable antitrust and competition laws. It is therefore extremely important that attendees adhere to meeting agendas, and be aware of, and not participate in, any activities that are prohibited under applicable US state, federal or foreign antitrust and competition laws. Please contact legal@finos.org with any questions.

- FINOS project meetings may be recorded for use solely by the FINOS team for administration purposes. In very limited instances, and with explicit approval, recordings may be made more widely available.

## Agenda

- [ ] Convene & roll call (5mins)
- [ ] Display [FINOS Antitrust Policy summary slide](https://community.finos.org/Compliance-Slides/Antitrust-Compliance-Slide.pdf) and review Meeting Notices (see above)
- [ ] Approve past meeting minutes
- Git Proxy Roadmap discussion
  - [ ] 5 mins per team on goals for the project
  - [ ] Recently opened issues
  - [ ] Start planning
- [ ] AOB, Q&A & Adjourn (5mins)

## Git Proxy Project – Consolidated Roadmap Summary

[...]

PR #1703, opened 2026-08-28
fix: broken UI build path

### LINKED ISSUE #1704: UI files are not present in 2.0 and 2.1 NPM packages
As mentioned by @re-vlad in #1702, we weren't shipping the built UI files along with the NPM release:

Image
 
This issue is present on both 2.0 and 2.1 releases, which should be patched or deprecated.

As a consequence, our quickstart guide no longer works as expected:

Running `npx -- @finos/git-proxy` does not return any particular error. However, `localhost:8080` (or other URLs) fail to get served correctly since the files are missing:

Image

I believe this might have broken when I refactored the NPM publish flow to include only build files (since we were shipping the entire source before). Another culprit might be `tsconfig.publish.json` explicitly excluding `src/ui/**`.

We should get the build process fixed, and set up a proper testing script (local and CI executable) so that we catch any bad packages during publish or each PR. 

**Additional context**
#1702

jescalada:
## Description

Originally brought up by @re-vlad on the roadmap call #1702

Fixes issue with UI being excluded from the built NPM package. Docker build was unaffected. This affects 2.0 and 2.1, so we'll have to get a patch out for 2.1.1 (or deprecate these two and ship the fix in 2.2):

image

Also adds a smoke test CI script to check that the UI is actually present and executable in the built package:

### After `npm run build` on `main`

image

### After `npm run build` with these changes

image

## Related Issue

Resolves #1704

### General

### Documentation

### Configuration

### Tests


package.json
```diff
@@ -61,6 +61,7 @@
     "build": "npm run generate-config-types && npm run build-ui && npm run build-ts",
     "build-ts": "tsc --project tsconfig.publish.json && node scripts/fix-shebang.js",
     "build-ui": "vite build",
+    "build-validate": "./scripts/build-validate.sh",
```
fabiovincenzi: Do you mean "./scripts/package-smoke-test.sh"?

scripts/package-smoke-test.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+CLEAN_BUILD=true
+if [[ "$1" == "--no-clean-build" ]]; then
```
fabiovincenzi: With `set -u`, running this without arguments like you specify in CONTRIBUTING.md dies before doing anything:

```
./scripts/package-smoke-test.sh: line 5: $1: unbound variable
```

scripts/package-smoke-test.sh
```diff
@@ -0,0 +1,77 @@
...
+  || { echo "FAIL: no JS bundle under dist/build/assets."; exit 1; }
+
+# Install like a normal user
+cd "$WORK"
+npm init -y >/dev/null
+npm install --no-audit --no-fund --loglevel=error "$TARBALL"
+
+# Boot from a scratch cwd so it writes its .data/.tmp
+mkdir -p "$WORK/run" && cd "$WORK/run"
+"$WORK/node_modules/.bin/git-proxy" > "$WORK/server.log" 2>&1 &
```
fabiovincenzi, line 55: We never check that the server we started is actually the one answering. If something else is already on 8080, our `git-proxy` dies on EADDRINUSE and every assertion runs against the alredy running one.

package.json
jescalada: Good catch 🙂

scripts/package-smoke-test.sh
jescalada: Interesting, didn't notice this one until actually running it locally again. One of my later commits must've broken the local run... Fixed now 👍🏼

jescalada:
Hmm, still having issues here - will take a better look at it tomorrow!

jescalada:
@fabiovincenzi Ready for a final look! I factored the smoke test into its own file so we don't get any port clashing issues.

fabiovincenzi (approved):
LGTM!

PR #1708, opened 2026-08-31
refactor(config): drive deprecation warnings from config.schema.json

### LINKED ISSUE #1680: Pull deprecated status from config.schema.json, make warnings generic
In #1626, we added deprecation warnings for legacy config fields such as `sslKeyPemPath`, `sslCertPemPath` and `proxyUrl`. However, the checks are baked into `src/config/deprecatedFields.ts`. We can make this cleaner and more extensible by picking up deprecated fields from `config.schema.json` and keeping it as the source-of-truth.

Fix `getDeprecatedConfigWarnings()` to pull the deprecated status from `schema.config.json` and use that to generate the warnings. Perhaps we could add `x-deprecated-replacement` to the schema so users how to replace the deprecated fields.

**Additional context**
Mentioned in #1629 and #1665, original PR #1626, related issue #1545

amitvijapur:
## Description

`getDeprecatedConfigWarnings` hardcoded one check per deprecated field, while `config.schema.json` already marked the same three fields `"deprecated": true`. The two had to be kept in step by hand, so adding a deprecation to the schema produced no warning until someone remembered to edit `deprecatedFields.ts` as well.

Warnings are now generated from the schema. Marking a property `"deprecated": true` is sufficient; `x-deprecated-replacement` names the field that supersedes it and selects the "use X instead" wording, and its absence gives the "deprecated and ignored" wording used by `proxyUrl`.

Warning text is unchanged for all three existing fields.

To be clear about what this is: it is a refactor, not a bug fix. Behaviour for the current three fields is identical, which is why the pre-existing tests pass untouched. The change is that a fourth deprecation now needs no code. I checked that concretely by temporarily marking a fourth property deprecated in the schema: the new code emits a warning, the old code returns `[]`.

The loop reads top-level `properties` only. There are no nested deprecations today, and a deprecation nested under `tls` or `configurationSources` would not be picked up.

## Related Issue

Resolves #1680

### General

- [ ] I have a [FINOS CLA](https://finosfoundation.atlassian.net/wiki/spaces/FINOS/pages/75530375/Contribution+Compliance+Requirements) on file

### Documentation

- [x] Documentation has been added/updated for any new features — N/A, no user-facing behaviour change

### Configuration

  - [x] TypeScript types regenerated (`npm run generate-config-types`) — output is byte-identical, quicktype ignores the `x-` keyword
  - [x] Schema reference docs regenerated (`npm run gen-schema-doc`) — see note below

### Tests

## Notes for the reviewer

**On `gen-schema-doc`.** I ran it and confirmed my schema change produces a byte-identical `reference.mdx`, since `json-schema-for-humans` ignores the custom keyword. I have deliberately not committed the regenerated file, because running it also pulls in an unrelated `sidebandProgress` section: that property is present in `config.schema.json` on `main` but missing from the committed `website/docs/configuration/reference.mdx`, so the doc is already stale independently of this PR. Happy to include that regeneration here if you would rather, but it looked like it belongs in its own commit.

**On the two new tests.** Neither the no-replacement branch (`proxyUrl`) nor the suppression branch (deprecated field set *and* its replacement set) had coverage before this PR. Both are covered now.

**On `_.get`.** `src/config/validators.ts` has an equivalent dotted-path lookup, but it is a local closure inside `validateConfigRegex` rather than an exported helper, so reusing it would have meant refactoring that function too. I used `lodash.get` instead, since lodash is already a direct dependency used elsewhere in `src/`. Happy to extract and share the existing helper instead if you prefer one implementation.

**Disclosure.** This PR was written with AI assistance. I understand and can justify the change end to end; the points I am least certain about are flagged above (the top-level-only scope, and the `_.get` choice).


src/config/deprecatedFields.ts
```diff
@@ -14,27 +14,42 @@
...
 
+interface SchemaProperty {
+  deprecated?: boolean;
+  'x-deprecated-replacement'?: string;
+}
+
+const schemaProperties = (schema as { properties: Record<string, SchemaProperty> }).properties;
+
+function isSet(value: unknown): boolean {
+  return typeof value === 'string' && value.trim() !== '';
```
jescalada: Might want to expand this so that it works with booleans, numbers, etc. instead of just strings:

```suggestion
  return value !== undefined && value !== null && (typeof value !== 'string' || value.trim() !== '')
```

jescalada (dismissed):
Thanks for the contribution @amitvijapur! Looks good after applying my suggestion.

Do you mind filling in [EasyCLA](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/13399143/256513079/1708/#/?version=2) in order to get this merged?

amitvijapur:
Applied the suggestion in 14c72702, thanks. Added a test covering the non-string case so the widened behaviour is pinned. EasyCLA is signed and now passing.

amitvijapur:
@jescalada when you have a moment — your suggestion is applied in `14c72702`, EasyCLA is signed and passing, and all checks are green. Only the review state from 1 September is still outstanding.

(GitHub does not let me send a re-review request from a fork, hence the comment.)

amitvijapur:
Gentle follow-up on this one. Everything on my side is done:

- @jescalada's suggestion applied in `14c72702`
- EasyCLA passing ("You are authorized to contribute")
- All checks green, branch mergeable, no conflicts

The only thing outstanding is the review state from 1 September, which predates the change that addressed it. I can't send a re-review request from a fork, so I'm flagging it here instead.

@jescalada if you're busy, would any other maintainer be able to take a look? Happy to rebase or adjust anything if it's gone stale.

kriswest (approved):
LGTM

amitvijapur:
Thanks for the review @kriswest.

Merged current main in `fd2d7f75` to clear the `BEHIND` state, which also re-triggers CI. The branch was 24 commits behind after the root-level cleanup in #1712, which moved the vitest configs; our tests still pass under the new `test/vitest.config.ts` (5 passed).

On the red checks from the previous run: the `build-windows` failure was `test/proxy/performance.test.ts > Error Handling Performance > should handle errors quickly without memory leaks`. This PR only touches `config.schema.json`, `src/config/deprecatedFields.ts` and its test, so it does not reach proxy performance or memory. #1726 also fails `build-windows` but on a different test (`test/ssh/security.test.ts`), which suggests the Windows runner rather than either change. Flagging in case it is worth a look separately.

Disclosure: written with AI assistance, reviewed by me before posting.

amitvijapur:
One thing needs a maintainer: the workflow runs on `fd2d7f75` are all sitting at `action_required`, so CI has not actually executed on the merge commit.

```
CI                    action_required
E2E Tests             action_required
CodeQL                action_required
CVE Scanning          action_required
Code Cleanliness      action_required
Dependency Review     action_required
Package Smoke Test    action_required
Unused Dependencies   action_required
```

Only the seven checks that do not need approval have reported (EasyCLA, labelling, Netlify preview), which is why the PR shows `BLOCKED` despite the approval. Could someone approve the workflow run? @kriswest @jescalada

Nothing else outstanding as far as I can tell: approved, no conflicts, no longer behind, and our tests pass locally under the relocated `test/vitest.config.ts` (5 passed).

PR #1713, opened 2026-09-03
fix(proxy): restore pusher email from the token cache

### LINKED ISSUE #1400: checkUserPushPermission is checking the wrong user's permission to push

sankalpsthakur:
### Summary
`checkUserPushPermission` looks up the pusher by `action.userEmail`. After #1604, `resolveUserFromToken` maps a GitHub PAT to a GitProxy user on a cold cache, but the token cache only stored the username. On a cache hit the username was overwritten and the last-commit committer email from `parsePush` was left in place, so the permission check ran against the wrong user (the remaining path on #1400 that @jescalada reproduced). Tag pushes never ran `resolveUserFromToken` at all.

- Cache `{ username, email }` and restore both on hit
- Run `resolveUserFromToken` first on the tag-push chain, same as branch pushes

Does **not** close #1400: GHES / non-github.com providers and unmatched `gitAccount` are unchanged.

### Validation
- `npx vitest run test/processors/resolveUserFromToken.test.ts test/chain.test.ts test/testCheckUserPushPermission.test.ts`
- 58 passed (30 + 25 + 3)

Related to #1400

### AI/LLM disclosure
AI coding tools assisted with the change and this description.


jescalada (commented):
@coopernetes Wondering if this looks appropriate and how it affects your pusher identity work 🤔

coopernetes:
Thanks @sankalpsthakur for the contribution. The cache-hit path leaving the committer email in place is a real bug, and adding the resolver to the tag chain is correct. Appreciate raising this.

After reviewing this change and considering the underlying flaw in how users are resolved from tokens today (only affecting the auth decision on some paths, and an incomplete fix for the problems described in #1400), we should pursue a more complete and direct fix where commit metadata (committer, tagger, author) is _not_ considered at all for git-proxy authorization decisions as part of a git push. This is due to the simple fact that it is client controlled and it is too simple to forge in order to elevate permissions onto a repo a particular user isn't intended to have push access to.

There is a more complete fix on the way which will refactor this completely. The git-proxy user can (and should) contain the additional, non-permission-affecting information such as email addresses but _only_ for attribution checks and never for authorization. It touches the same three files this PR changes, so this would be superseded on merge either way.

@jescalada I'd suggest we hold this in favour of the larger change. Once it lands, if there's anything still missing, I'd be happy to accept a contribution then. Until then, we're going to revisit these assumptions rather than continue living with them and patching around it.

